PR #1, opened 2025-04-19
add OptionStratLib as dependency

ikeepo:
build dependency

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@
 
 1. Clone the repository:
 ```bash
+# Prerequirement
+git clone https://github.com/joaquinbejar/OptionStratLib.git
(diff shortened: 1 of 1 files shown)


joaquinbejar (changes requested):
Hi @Ikeepo,
first of all, thanks for your contribution! 😊

Just a couple of things to keep in mind:
	•	The README.md file shouldn’t be edited directly. It’s auto-generated using make readme, which pulls content from lib.rs. I’ll soon include a README.tpl template as well.
	•	So, if you’d like to change anything in README.md, please update lib.rs and then run make readme.

Also, OptionStratLib is indeed a direct dependency of the project, but there’s no need to clone it — it’s available on crates.io and gets downloaded automatically.

Cheers!

ikeepo:
got it

PR #23, opened 2026-07-13
Redact Redis and MongoDB credentials from logs (#15)

joaquinbejar:
## Summary

Redis and MongoDB credentials were printed in plaintext by the config
`Display`/derived `Debug` impls and logged at three startup sites, leaking
secrets into local logs, container logs, CI output, and observability
backends. This PR redacts credentials from every formatted/logged form while
leaving the actual connection strings untouched.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 1 (bottom of the stack)**
- **Base branch:** `main`
- **Stacked on:** — (first in the chain)
- **Blocks:** the next PR in the stack (#13)

## Changes

- Shared `redact_userinfo` helper (plain string scanning, no new deps) that
  replaces URL userinfo with `***` in any URL-like substring.
- `RedisConfig`: derived `Debug` dropped; manual `Display`/`Debug` render the
  redacted URL. `url()` (the connection value) is unchanged.
- `MongoDBConfig`: derived `Debug`/`Serialize`/`Deserialize` removed (nothing
  in the crate serializes it); manual `Display`/`Debug` redact the URI.
- `RedisClient::new` and `MongoDBClient::new` log redacted forms; driver
  connection errors that can echo the URI are sanitized before logging or
  wrapping into `ChainError`.
- Sentinel tests assert `s3ntinel-pw` never appears in `Display`/`Debug`/
  sanitized-error output; old tests that asserted the leaking output were
  rewritten to expect redaction.

## Testing

- [x] Unit tests added or updated (`LOGLEVEL=WARN cargo test --workspace`, MongoDB up): 229 + 2 passed, 0 failed
- [x] Sentinel-secret assertions on Display / Debug / sanitized errors
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean; `cargo fmt --all --check` clean

## Checklist

- [x] Code follows `rules/global_rules.md` and `CLAUDE.md`
- [x] Errors converge on `ChainError`; no driver error leaks through `pub` signatures
- [x] No `.unwrap()` / `.expect()` / unchecked indexing on request paths
- [x] `tracing` only for logging
- [x] No `unsafe` introduced; no new dependencies
- [x] Connection behavior unchanged (`url()`/`uri` intact)

[...]

Issue #22, opened 2026-07-13
Pin Docker Compose service images for reproducible deployments


joaquinbejar:
## Summary

The deployment composes four infrastructure services from mutable `:latest` tags.

## Evidence

As of `348611b`, `Docker/docker-compose.yml` uses:

- `redis:latest`
- `mongo:latest`
- `mongo-express:latest`
- `clickhouse/clickhouse-server:latest`

The Rust builder is pinned, but runtime dependencies can change between otherwise identical builds.

## Impact

A redeploy can introduce incompatible protocols/configuration, break health checks, or pull unreviewed changes. Rollback and incident reproduction become unreliable.

## Proposed direction

Pin reviewed major/minor/patch versions, preferably immutable digests for production. Document an intentional update workflow and compatibility matrix.

## Acceptance criteria

- No production service uses a mutable tag.
- Dependency updates are explicit and tested.
- The compose stack records versions/digests sufficient to reproduce a deployment.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #21, opened 2026-07-13
GET /api/v1/chain mutates session state


joaquinbejar:
## Summary

The endpoint documented as “Read Next Step” advances and persists the session on every GET. This violates HTTP safe/idempotent semantics.

## Evidence

As of `348611b`, the GET route calls `SessionManager::get_next_step`, which advances state and saves the session (`src/api/rest/routes.rs:60-66`, `src/session/manager.rs:130-144`). The README also describes GET as advancing the simulation.

## Impact

Browsers, proxies, SDK retry policies, monitoring, prefetchers, and users repeating a URL can consume steps unintentionally. GET responses cannot be safely retried or cached, and ambiguous network failures leave clients unsure whether a step was consumed.

## Proposed direction

Use a mutation verb/resource, for example `POST /api/v1/sessions/{id}/steps`, or make GET cursor-addressed and immutable (`GET .../steps/{index}`). Support idempotency keys if advancing remains command-based.

## Acceptance criteria

- Read endpoints are safe and repeatable.
- Step advancement has explicit command/idempotency semantics.
- Migration/backward compatibility is documented.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #20, opened 2026-07-13
PATCH cannot update skew_slope or clear optional parameters


joaquinbejar:
## Summary

The PATCH representation is incomplete and cannot express several valid state changes.

## Evidence

As of `348611b`:

- Create/session response models include `skew_slope`, but `UpdateSessionRequest` has no `skew_slope` field (`src/api/rest/requests.rs:108-155`).
- Optional fields use `Option<T>`. Serde maps both an absent key and JSON `null` to `None`.
- The handler interprets `None` only as “leave unchanged” (`src/api/rest/handlers.rs:361-416`).

Consequently clients cannot PATCH `skew_slope`, and cannot clear `chain_size`, `strike_interval`, `smile_curve`, `spread`, or `seed` back to `None`.

## Impact

The advertised partial-update API cannot represent the full domain state. Clients must replace the entire session to remove a setting, and skew slope is silently immutable via PATCH.

## Proposed direction

Add the missing field and use a tri-state patch representation (absent / null / value), such as `Option<Option<T>>` with appropriate serde handling or a dedicated patch enum.

## Acceptance criteria

- Every mutable create parameter is represented in PATCH.
- Absent preserves, null clears, and value replaces.
- OpenAPI documents the three states.
- Round-trip handler tests cover each optional field.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

PR #24, opened 2026-07-13
Parameterize ClickHouse symbol queries (#13)

### LINKED ISSUE #13: Parameterize ClickHouse symbol queries to prevent SQL injection
## Summary

User-controlled symbols are interpolated directly into ClickHouse SQL strings.

## Evidence

As of `348611b`:

- The API exposes `Historical.symbol: Option<String>` (`src/api/rest/models.rs:232-254`).
- `build_timeframe_query` formats the symbol inside a quoted SQL literal (`src/infrastructure/clickhouse/client.rs:220-272`).
- `get_date_range_for_symbol` does the same (`src/infrastructure/repositories/historical_repo.rs:177-196`).
- The resulting query is passed directly to `client.query`.

A symbol containing a quote can alter the predicate or make the query invalid. Logging also emits the constructed query at debug level.

## Impact

A client can manipulate historical-data queries, bypass symbol scoping, trigger expensive queries, or expose data available to the ClickHouse account.

## Proposed direction

Use ClickHouse query binding/parameters for every external value. Validate symbols to the supported identifier format as defense in depth; do not rely on validation as the escaping mechanism.

## Acceptance criteria

- No external string is concatenated into SQL.
- Tests cover quotes, comment markers, Unicode, and oversized symbols.
- Query logs do not include injected raw SQL fragments.

## Revalidation

joaquinbejar:
## Summary

User-controlled symbols were interpolated with `format!` directly into
ClickHouse SQL (timeframe queries and the date-range query), letting a symbol
containing a quote alter the predicate, and the fully-interpolated SQL was
logged at debug level. Queries now use the clickhouse crate's `?` binding for
every external value, with identifier-format validation as defense in depth.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 2**
- **Base branch:** `stack/1-15-redact-credentials`
- **Stacked on:** #23 · **Blocks:** the next PR in the stack (#7)
- The diff is cumulative until PR #23 merges — review against the base branch.

## Changes

- `build_timeframe_query` now returns a prepared `clickhouse::query::Query`
  with `symbol` / `start_timestamp` / `limit` bound via `.bind()` (driver
  escaping); the SQL templates carry only `?` placeholders. Interval literals
  stay inline (internal match, not user input).
- `get_date_range_for_symbol` parameterized the same way
  (`const DATE_RANGE_QUERY` + `.bind(symbol)`).
- New `validate_symbol` (`^[A-Za-z0-9._-]{1,32}$`, plain char checks, no new
  deps) at the entry of every symbol-taking path — defense in depth, not the
  escaping mechanism.
- Interpolated-SQL debug logging removed; logs now carry the template and the
  symbol as structured fields.

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 235 + 2 passed, 0 failed (7 new tests)
- [x] Injection cases covered: quotes, `' OR '1'='1`, `"; DROP TABLE ohlcv; --`, Cyrillic homoglyph, oversize, empty
- [x] SQL templates asserted to contain `?` and never a raw symbol

## Checklist

- [x] No external string concatenated into SQL
- [x] Errors converge on `ChainError` at the adapter

Closes #13

### DIFF
--- a/src/infrastructure/clickhouse/utils.rs
+++ b/src/infrastructure/clickhouse/utils.rs
@@ -3,6 +3,38 @@ use chrono::{DateTime, Duration, Utc};
 use optionstratlib::utils::TimeFrame;
 use rand::{Rng, RngExt};
 
+/// Maximum number of characters allowed in a symbol.
+const MAX_SYMBOL_LEN: usize = 32;
+
+/// Validates an externally supplied symbol as a defense-in-depth measure on top
+/// of query parameter binding.
+///
+/// The allowed pattern is `^[A-Za-z0-9._-]{1,32}$`, i.e. between 1 and 32 ASCII
+/// alphanumeric characters, dots, underscores or hyphens. The check is
+/// implemented with plain character inspection to avoid pulling in a regex
+/// dependency, and it rejects any non-ASCII input (e.g. look-alike Unicode
+/// characters) as well as empty or overly long values.
+///
+/// # Errors
+///
+/// Returns [`ChainError::ClickHouseError`] when the symbol is empty, longer than
+/// 32 characters, or contains a character outside the allowed set.
+pub fn validate_symbol(symbol: &str) -> Result<(), ChainError> {
+    let len = symbol.chars().count();
+    let is_valid = (1..=MAX_SYMBOL_LEN).contains(&len)
+        && symbol
+            .chars()
+            .all(|c| c.is_ascii_alphanumeric() || matches!(c, '.' | '_' | '-'));
+
+    if is_valid {
+        Ok(())
+    } else {
+        Err(ChainError::ClickHouseError(format!(
+            "invalid symbol: {symbol}"
+        )))
+    }
+}
+
 /// Calculates the required duration based on timeframe and steps
 pub fn calculate_required_duration(timeframe: &TimeFrame, steps: usize) -> Duration {
     match timeframe {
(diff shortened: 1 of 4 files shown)


src/infrastructure/clickhouse/utils.rs
```diff
@@ -3,6 +3,38 @@ use chrono::{DateTime, Duration, Utc};
...
+    let len = symbol.chars().count();
+    let is_valid = (1..=MAX_SYMBOL_LEN).contains(&len)
+        && symbol
+            .chars()
+            .all(|c| c.is_ascii_alphanumeric() || matches!(c, '.' | '_' | '-'));
+
+    if is_valid {
+        Ok(())
+    } else {
+        Err(ChainError::ClickHouseError(format!(
```
joaquinbejar, line 32: 🟠 **Warning**: invalid `Historical.symbol` values become `ClickHouseError`, which `map_error` turns into HTTP 500, while session creation still accepts them. Validate the symbol at the REST conversion boundary and return a client-input error with 400, reserving `ClickHouseError` for adapter failures.

joaquinbejar (commented):
The SQL binding is sound, but invalid client symbols are still surfaced as server failures. Intended decision REQUEST_CHANGES, submitted as COMMENT because GitHub does not allow requesting changes on a pull request authored by the current account.

src/infrastructure/clickhouse/utils.rs
joaquinbejar: Agreed on the 400. The fallible conversion boundary that can reject a request only exists from PR #26 onward (this level still has the infallible From impls), so the symbol validation at the REST boundary lands there with a Validation error and a regression test, and this layer keeps its check as defense in depth.

joaquinbejar:
Rebased onto main after PR 23 merged. The 400-at-the-boundary fix ships in PR 26 where the fallible conversion layer lives.

Issue #19, opened 2026-07-13
Synchronous Redis I/O blocks async HTTP workers and serializes all sessions


joaquinbejar:
## Summary

The async Actix/Tokio server uses the synchronous Redis `Commands` API through one `std::sync::Mutex<Connection>`.

## Evidence

As of `348611b`:

- `RedisClient` owns a single synchronous `redis::Connection` behind `std::sync::Mutex` (`src/infrastructure/redis/client.rs:1-10`).
- `get`, `set`, and `delete` lock it and perform blocking network calls (`src/infrastructure/redis/client.rs:38-97`).
- Async handlers call synchronous `SessionManager` store operations directly (`src/api/rest/handlers.rs`).
- One shared store/client is used for every session.

## Impact

A slow Redis command blocks an HTTP worker thread. All sessions contend on one connection, so unrelated requests become head-of-line blocked and throughput cannot scale with Actix workers.

## Proposed direction

Use Redis async commands with an async connection manager/pool, make the store trait async, and configure command/connect timeouts. At minimum, isolate blocking I/O with `spawn_blocking`, though a proper async pool is preferable.

## Acceptance criteria

- Redis calls do not block executor worker threads.
- Unrelated sessions can perform store operations concurrently.
- Timeouts and pool sizing are configurable.
- Load tests cover slow Redis and concurrent sessions.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #18, opened 2026-07-13
Default cargo test requires a live MongoDB instance


joaquinbejar:
## Summary

The library unit-test suite is not hermetic: one test asserts that localhost MongoDB is available.

## Evidence

As of `348611b`, `test_mongodb_client_initialization` constructs `mongodb://localhost:27017` and asserts `MongoDBClient::new(...).is_ok()` (`src/infrastructure/mongodb/client.rs:136-155`).

Running `cargo test --workspace` on a clean checkout without MongoDB produced:

- 219 passed
- 1 failed: `test_mongodb_client_initialization`
- about 30 seconds spent waiting for server selection

The adjacent `test_init_mongodb` already treats absence as expected, which makes the suite internally inconsistent.

## Impact

Contributors cannot run the documented default test command without external setup. The test is slow and environment-dependent, and package consumers may see failures unrelated to code.

## Proposed direction

Move live-database checks to explicitly gated integration tests (feature/env flag) and unit-test client configuration with mocks or driver options without pinging a server.

## Acceptance criteria

- `cargo test --workspace` passes without Redis, MongoDB, or ClickHouse.
- Live integration tests are opt-in and documented.
- CI runs both hermetic tests and a clearly named Mongo integration job.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #17, opened 2026-07-13
MONGODB_TIMEOUT does not bound server selection or startup failure time


joaquinbejar:
## Summary

The configured MongoDB timeout is applied only as `connect_timeout`; the ping can still wait for the driver's default server-selection timeout.

## Evidence

As of `348611b`, `MongoDBClient::new` sets only:

`client_options.connect_timeout = Duration::from_secs(config.timeout)`

before running an admin ping (`src/infrastructure/mongodb/client.rs:28-43`).

During the audit, a test configured with `timeout: 5` against unavailable localhost MongoDB failed after approximately 30.02 seconds, matching the default server-selection timeout rather than five seconds.

## Impact

When MongoDB is unavailable, application startup and tests block much longer than configured. Because MongoDB is mandatory in `main`, this delays restarts and recovery.

## Proposed direction

Set `server_selection_timeout` (and optionally operation timeout) from the configuration, clarify the timeout's semantics, and test elapsed-time bounds with a small deterministic timeout.

## Acceptance criteria

- An unavailable server fails within the configured bound plus a small tolerance.
- The setting applies to the startup ping.
- Documentation names each timeout controlled.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #16, opened 2026-07-13
Active-session metrics are updated before operation success


joaquinbejar:
## Summary

Session metrics count attempted requests rather than successful lifecycle transitions.

## Evidence

As of `348611b`:

- POST increments active sessions and creation count before conversion or storage succeeds (`src/api/rest/handlers.rs:57-66`).
- DELETE decrements active sessions and increments deletion count before UUID parsing and before confirming that a session exists (`src/api/rest/handlers.rs:487-503`).
- `IntGauge::dec` allows the gauge to become negative (`src/infrastructure/telemetry/collector.rs:172-181`).

Therefore invalid POSTs inflate the gauge, while invalid/repeated DELETEs reduce it and increment deletion totals.

## Impact

Dashboards and alerts report impossible active-session values and incorrect creation/deletion rates.

## Proposed direction

Record metrics only after successful create/delete results. Reconcile active-session semantics with Redis TTL expiry and process restarts, or expose it as a derived/store metric.

## Acceptance criteria

- Failed POST/DELETE requests do not change lifecycle counters.
- Repeated/nonexistent DELETE cannot make the gauge negative.
- TTL expiry and restart behavior are documented and tested.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

PR #25, opened 2026-07-13
Use random session ids and fail on create collisions (#7)

### LINKED ISSUE #7: Session IDs collide after a restart or across service instances
## Summary

Every `SessionManager` starts the same deterministic UUIDv5 sequence at counter zero, using a fixed namespace. The first session created after every restart or in every replica receives the same ID.

## Evidence

As of `348611b`:

- `SessionManager::new` always parses `DEFAULT_NAMESPACE` and constructs a new `UuidGenerator` (`src/session/manager.rs:39-52`).
- `UuidGenerator::new` initializes its counter to zero (`src/utils/uuid.rs:62-66`).
- `next` hashes only the fixed namespace and counter (`src/utils/uuid.rs:77-82`).
- The tests explicitly assert that two generators with the same namespace produce identical IDs (`src/utils/uuid.rs:127-135`).

Creating one session from each of two fresh managers reproduces an identical UUID.

## Impact

With Redis persistence, a restart can overwrite a still-live session. Multiple replicas can overwrite each other's sessions immediately, leaking or corrupting user state.

## Proposed direction

Use UUIDv4/v7 for externally visible session IDs, or persist/coordinate a globally unique generator state. Do not rely on a process-local counter with a constant namespace.

## Acceptance criteria

- Fresh managers and separate processes cannot generate the same session ID.
- Creating a session never silently overwrites an existing Redis key.
- Add a restart/multi-manager collision regression test.

## Revalidation

joaquinbejar:
## Summary

Every `SessionManager` started the same deterministic UUIDv5 sequence at
counter zero with a fixed namespace, so the first session after every restart
— or in every replica — received the same id, and `store.save` (a blind
upsert) silently overwrote the still-live session in Redis. Session ids are
now random v4 UUIDs and creation goes through a non-overwriting store
operation.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 3**
- **Base branch:** `stack/2-13-clickhouse-sql-binding`
- **Stacked on:** #24 · **Blocks:** the next PR in the stack (#10)

## Changes

- `Session::with_random_id` (`Uuid::new_v4()`); `SessionManager::create_session`
  uses it and no longer owns a `UuidGenerator` (the deterministic generator
  stays in `utils` for its other uses).
- New `SessionStore::create` that FAILS on an existing id — in-memory via a
  `contains_key` guard, Redis via a new `RedisClient::set_nx` (`SET … NX EX`),
  mapping an unset key to the new `ChainError::AlreadyExists`.
- `AlreadyExists` → HTTP 409 in `map_error`; POST utoipa annotation documents
  the 409.

## Contract impact

Additive: POST `/api/v1/chain` can now return 409 on an id collision
(previously it silently overwrote). Session ids are no longer deterministic
across restarts — that determinism was the bug; the tape-reproducibility
contract (parameters + seed) is unaffected.

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 246 + 2 passed, 0 failed (11 new tests)
- [x] Restart-collision regression: two fresh managers produce different first ids
- [x] `create` collision → `AlreadyExists` in both store impls; `save` still updates

## Checklist

- [x] Stored-session serde shape unchanged
- [x] Effective-seed guarantee untouched

Closes #7

### DIFF
--- a/src/session/store/interface.rs
+++ b/src/session/store/interface.rs
@@ -69,6 +69,26 @@ pub trait SessionStore: Send + Sync {
     ///
     fn get(&self, id: Uuid) -> Result<Session, ChainError>;
 
+    /// Creates a brand-new session, failing if one with the same id already exists.
+    ///
+    /// Unlike [`SessionStore::save`], which is a blind upsert, `create` must never
+    /// overwrite an existing session. This is what makes a fresh manager (after a
+    /// restart, or a second replica) safe: a colliding id is rejected instead of
+    /// silently clobbering a live session.
+    ///
+    /// # Parameters
+    /// - `session`: The `Session` to insert.
+    ///
+    /// # Returns
+    /// - `Ok(())`: If the session was inserted.
+    /// - `Err(ChainError::AlreadyExists)`: If a session with the same id already exists.
+    /// - `Err(ChainError)`: If the underlying storage operation fails.
+    ///
+    /// # Errors
+    /// Returns `ChainError::AlreadyExists` on an id collision, or another `ChainError`
+    /// variant on a storage/serialization failure.
+    fn create(&self, session: Session) -> Result<(), ChainError>;
+
     /// Saves the provided session into persistent storage or memory.
     ///
     /// # Parameters
@@ -136,6 +156,7 @@ mod tests {
         pub SessionStore {}
         impl SessionStore for SessionStore {
             fn get(&self, id: Uuid) -> Result<Session, ChainError>;
+            fn create(&self, session: Session) -> Result<(), ChainError>;
             fn save(&self, session: Session) -> Result<(), ChainError>;
             fn delete(&self, id: Uuid) -> Result<bool, ChainError>;
             fn cleanup(&self) -> Result<usize, ChainError>;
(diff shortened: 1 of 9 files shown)


src/session/store/interface.rs
```diff
@@ -69,6 +69,26 @@ pub trait SessionStore: Send + Sync {
...
+    ///
+    /// # Returns
+    /// - `Ok(())`: If the session was inserted.
+    /// - `Err(ChainError::AlreadyExists)`: If a session with the same id already exists.
+    /// - `Err(ChainError)`: If the underlying storage operation fails.
+    ///
+    /// # Errors
+    /// Returns `ChainError::AlreadyExists` on an id collision, or another `ChainError`
+    /// variant on a storage/serialization failure.
+    fn create(&self, session: Session) -> Result<(), ChainError>;
```
joaquinbejar, line 90: 🟠 **Warning**: adding a required method to the public `SessionStore` trait breaks every downstream implementation at compile time. The published crate needs an explicit compatibility and semver plan for this change, and the new public `AlreadyExists` enum variant has the same exhaustive-match impact.

joaquinbejar (commented):
Creation is now atomic, but the published Rust API change needs breaking-release treatment. Intended decision REQUEST_CHANGES, submitted as COMMENT because GitHub does not allow requesting changes on a pull request authored by the current account.

src/session/store/interface.rs
joaquinbejar: Agreed this is a breaking change to the published Rust API. The whole stack ships as the 0.1.0 breaking release, the release notes will list the SessionStore trait additions, the new AlreadyExists variant, and the endpoint split, and the wire protocol stays compatible apart from the documented GET peek change.

joaquinbejar:
Rebased onto main. Shipping as part of the 0.1.0 breaking release per the semver plan.

PR #26, opened 2026-07-13
Validate simulation parameters instead of panicking (#10)

joaquinbejar:
## Summary

Request DTOs accepted unconstrained numbers and the infallible `From`
conversions called panicking macros — a request with `initial_price = -1.0`
panicked a worker instead of returning 400, invalid decimals were silently
zeroed by `unwrap_or_default`, and `steps`/`chain_size`/historical payloads
had no resource bounds. All user-controlled numerics are now validated through
fallible conversions that return a structured 400 naming the offending field.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 4**
- **Base branch:** `stack/3-7-unique-session-ids`
- **Stacked on:** #25 · **Blocks:** the next PR in the stack (#21)
- The diff is cumulative until the lower PRs merge — review against the base branch.

## Changes

- New `ChainError::Validation { field, reason }` → HTTP 400 with `{"error", "field"}` body.
- `From` → `TryFrom` for `CreateSessionRequest → SimulationParameters` and
  `ApiWalkType → WalkType` (still exhaustive, no `_ =>`): finite checks before
  conversion, `Positive::new` instead of `pos_or_panic!` on user input,
  `Decimal::try_from` errors propagate, autocorrelation ∈ [-1,1], historical
  prices each positive and capped.
- `ApiTimeFrame::Custom` validated (finite, > 0) via `time_frame_field` — the
  infallible conversion panicked on a negative custom timeframe.
- Env-configurable bounds in `api/rest/limits.rs` (`OCS_MAX_STEPS` 10_000,
  `OCS_MAX_CHAIN_SIZE` 500, `OCS_MAX_HISTORICAL_PRICES` 100_000; parse-once
  `LazyLock`, warn+default on invalid).
- Shared `api/rest/validation.rs` helpers reused by create/PUT/PATCH merge
  paths; PATCH no longer uses any panicking conversion.
- POST metric increment moved after validation (a rejected request no longer
  inflates the active-sessions gauge — completes under #16 later in the stack).

## Contract impact

Additive: requests that previously panicked the worker or silently zeroed
values now return a documented 400 whose body includes `field`. Success
responses unchanged. Effective-seed generation and the same-seed ⇒ same-tape
contract untouched (reproducibility tests green).

## Testing

[...]

Issue #15, opened 2026-07-13
Redis and MongoDB credentials are written to application logs


joaquinbejar:
## Summary

Connection configuration formatting includes plaintext credentials, and startup logs emit those formatted values/URLs.

## Evidence

As of `348611b`:

- `RedisConfig::url` embeds username/password and `Display` prints the complete URL (`src/infrastructure/config/redis.rs:19-52`, `82-88`).
- `main` logs `RedisConfig`, and `RedisClient::new` logs the URL again (`src/main.rs:109-112`, `src/infrastructure/redis/client.rs:17-23`).
- `MongoDBConfig::Display` prints the full URI (`src/infrastructure/config/mongo.rs:38-45`).
- `init_mongodb` logs that display value (`src/infrastructure/repositories/mongo_repo.rs:77-82`).
- Both config structs derive `Debug`, which also exposes secret fields.

## Impact

Credentials can leak into local logs, container logs, CI output, observability backends, and incident bundles.

## Proposed direction

Separate connection-string construction from redacted display/debug output. Redact userinfo consistently and avoid logging full URIs.

## Acceptance criteria

- No log/Display/Debug output contains Redis or MongoDB passwords.
- Tests explicitly assert that sentinel secrets never appear.
- Connection errors are sanitized before logging.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #14, opened 2026-07-13
OHLCV aggregation uses nondeterministic any() for open and close


joaquinbejar:
## Summary

Aggregated historical candles use `any(open)` and `any(close)`. ClickHouse does not guarantee that those values are the first and last observations in the interval.

## Evidence

As of `348611b`, the non-minute query in `src/infrastructure/clickhouse/client.rs:251-273` groups by interval and selects:

- `any(open) as open`
- `any(close) as close`

There is no ordering inside those aggregates. The outer `ORDER BY interval_start` orders groups, not rows used by `any`.

## Impact

Hourly/daily/weekly/monthly candles can have incorrect and nondeterministic opens/closes. Historical paths currently consume `close`, so identical database contents can produce inaccurate simulation inputs.

## Proposed direction

Use timestamp-aware aggregates such as `argMin(open, timestamp)` and `argMax(close, timestamp)`. Add fixture-based tests with shuffled insertion order.

## Acceptance criteria

- Open is the earliest row and close is the latest row in each interval.
- Results are stable regardless of insertion/part order.
- High, low, volume, open, and close are all verified against a known fixture.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

PR #27, opened 2026-07-13
Split safe GET peek from POST step advancement (#21)

### LINKED ISSUE #21: GET /api/v1/chain mutates session state

joaquinbejar:
## Summary

`GET /api/v1/chain` advanced and persisted the session on every call — a
mutating GET that let browsers, proxies, retries, prefetchers, and monitoring
consume steps unintentionally. Reads and advances are now separate endpoints:
GET is a safe, repeatable peek; advancing is an explicit POST command.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 5**
- **Base branch:** `stack/4-10-validate-request-params`
- **Stacked on:** #26 · **Blocks:** the next PR in the stack (#5)
- Note: the cursor semantics POST exposes (serve-then-advance, index-0
  handling, persisted completion) are fixed by the NEXT PR in the stack
  (issue #5); this PR only splits the HTTP surface.

## Changes

- New `SessionManager::peek_current_step` — `store.get` + simulate at the
  current cursor, NO `advance_state`, NO `store.save`. Peek on a `Completed`
  session → 410 (consistent with the exhausted-advance path).
- `GET /api/v1/chain` → `get_current_step` (peek; no step metrics, no Mongo
  step events). `POST /api/v1/chain/step` → `advance_step` (the old behavior:
  advance + persist + metrics + event log).
- Shared `build_chain_response` helper; delta/gamma conversions no longer use
  bare `.unwrap()` on the response path.
- utoipa + `src/lib.rs` docs + regenerated `README.md` document the migration.

## Contract impact

**BREAKING (explicitly accepted by issue #21).** The `ChainResponse` DTO shape
and status codes are unchanged; the break is the verb/path split: clients that
advanced via `GET /api/v1/chain` must now call `POST /api/v1/chain/step`.
IronCondor's `ApiClient` needs that one-line change — coordinate before
releasing. Reproducibility contract untouched (peek only reads the walk at the
existing cursor).

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 283 + 2 passed, 0 failed
- [x] Peek is repeatable and read-only (stored state asserted unchanged); advance-then-peek reflects the new cursor; peek on Completed → 410 error kind
- [x] OpenAPI paths test covers `/api/v1/chain/step`
- [x] clippy `-D warnings` + fmt clean; `README.md` regenerated via cargo-readme

## Checklist

- [x] DTO shapes unchanged; `SessionParametersResponse.seed` untouched
- [x] Error mapping stays centralized in `map_error`
- [x] No `.unwrap()` / `.expect()` on request paths; `tracing` only

Closes #21

### DIFF
--- a/src/api/rest/handlers.rs
+++ b/src/api/rest/handlers.rs
@@ -8,14 +8,61 @@ use crate::api::rest::responses::{
 };
 use crate::api::rest::validation::{self, decimal_field, positive_field, strictly_positive_field};
 use crate::infrastructure::{MetricsCollector, MongoDBRepository};
-use crate::session::{SessionManager, SimulationParameters};
+use crate::session::{Session, SessionManager, SimulationParameters};
 use crate::utils::ChainError;
 use actix_web::{HttpRequest, HttpResponse, Responder, web};
 use chrono::{DateTime, Utc};
+use optionstratlib::chains::OptionChain;
 use rust_decimal::prelude::ToPrimitive;
 use std::sync::Arc;
 use tracing::{error, info};
 use uuid::Uuid;
+
+/// Builds the `ChainResponse` DTO shared by the advance (`POST /api/v1/chain/step`) and
+/// peek (`GET /api/v1/chain`) endpoints from a session and its current option-chain
+/// snapshot. Kept as a single place so both surfaces emit an identical response shape.
+fn build_chain_response(session: &Session, option_chain: &OptionChain) -> ChainResponse {
+    let expiration = option_chain.get_expiration_date();
+    ChainResponse {
+        underlying: option_chain.symbol.clone(),
+        timestamp: Utc::now().to_rfc3339(),
+        price: option_chain.underlying_price.into(),
+        contracts: option_chain
+            .iter()
+            .map(|contract| {
+                let (call_delta, put_delta) = contract.current_deltas();
+                let call_ask = contract.get_call_buy_price();
+                let put_ask = contract.get_put_buy_price();
+                let call_bid = contract.get_call_sell_price();
+                let put_bid = contract.get_put_sell_price();
+                let volatility = contract.get_volatility();
+                OptionContractResponse {
+                    strike: contract.strike().into(),
+                    expiration: expiration.clone(),
+                    call: OptionPriceResponse {
+                        bid: call_bid.map(|b| b.into()),
+                        ask: call_ask.map(|a| a.into()),
+                        mid: contract.call_middle.map(|m| m.into()),
+                        delta: call_delta.map(|d| d.to_f64().unwrap_or(0.0)),
+                    },
+                    put: OptionPriceResponse {
+                        bid: put_bid.map(|b| b.into()),
+                        ask: put_ask.map(|a| a.into()),
+                        mid: contract.put_middle.map(|m| m.into()),
+                        delta: put_delta.map(|d| d.to_f64().unwrap_or(0.0)),
+                    },
+                    implied_volatility: Some(volatility.into()),
+                    gamma: contract.current_gamma().map(|g| g.to_f64().unwrap_or(0.0)),
+                }
+            })
+            .collect(),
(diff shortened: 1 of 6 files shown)


src/api/rest/handlers.rs
```diff
@@ -150,50 +208,10 @@ pub(crate) async fn get_next_step(
         }
     };
 
-    // Get next step from session manager
+    // Advance the session one step (mutates state and persists it).
     match session_manager.get_next_step(session_id).await {
```
joaquinbejar, line 250: 🟠 **Warning**: moving advancement to POST makes the mutation explicit but does not make retries safe. If the response is lost after the save, retrying this call consumes another step. Accept an idempotency key or expected-cursor precondition and persist or deduplicate it atomically so clients can resolve ambiguous network failures.

src/session/manager.rs
```diff
@@ -146,6 +140,56 @@ impl SessionManager {
...
+    /// - [`ChainError::SimulatorError`] if the session has already completed all steps
+    ///   (there is no current step to serve); this maps to HTTP 410 like the
+    ///   exhausted-advance path.
+    /// - Any error surfaced by the simulator while building the current snapshot.
+    pub async fn peek_current_step(&self, id: Uuid) -> Result<(Session, OptionChain), ChainError> {
+        let session = self.store.get(id)?;
+
+        // A completed session has no current step to serve; mirror the exhausted-advance
+        // path (410 Gone) rather than returning stale data.
+        if session.state == SessionState::Completed {
```
joaquinbejar, line 176: 🟠 **Warning**: only `Completed` is rejected here, so a session in `Error` falls through and GET returns a 200 snapshot even though `Error` is terminal according to `Session::is_active`. Reject both terminal states before simulation and cover the `Error` case.

joaquinbejar (commented):
The GET and POST split fixes HTTP safety, but retry and terminal-state semantics remain incomplete. Intended decision REQUEST_CHANGES, submitted as COMMENT because GitHub does not allow requesting changes on a pull request authored by the current account.

src/api/rest/handlers.rs
joaquinbejar: Added an optional expected_step query param in b6ba096, when it does not match the current cursor the call returns 412 with the actual step and consumes nothing, so an ambiguous retry can be resolved safely. Full idempotency-key storage felt heavier than the issue needs, happy to revisit if you want dedup keys too.

src/session/manager.rs
joaquinbejar: Fixed in b6ba096, peek now rejects Error with the same message the advance path uses, with a regression test. The advance-side guard picks up Error in the next PR in the stack where the terminal guard lives.

joaquinbejar:
Both addressed in b6ba096. Rebased onto main.

Issue #12, opened 2026-07-13
Historical simulations ignore the session seed when selecting data


joaquinbejar:
## Summary

The reproducibility contract added for issue #3 does not cover historical simulations that load prices from ClickHouse. Symbol/date selection uses an unrelated thread RNG.

## Evidence

As of `348611b`:

- `get_historical_data` creates `rand::rng()` (`src/domain/simulator.rs:156-157`).
- It uses that RNG to choose a symbol and a random start date (`src/domain/simulator.rs:159-190`).
- The session seed is only used later to construct `Walker` (`src/domain/simulator.rs:298-303`).
- Historical loading happens before the walker is constructed (`src/domain/simulator.rs:233-253`).

Even when a symbol is supplied, the start date remains unseeded. Two sessions with identical parameters and seed can therefore load different tapes.

## Impact

The API and response imply that recording the effective seed is enough to replay a session, but historical sessions cannot be reproduced unless clients embed the full price vector.

## Proposed direction

Derive all randomness for a session—including symbol and date selection—from the session seed, or return and persist the resolved historical source metadata (symbol, start/end, data version).

## Acceptance criteria

- Same request + same seed selects the same historical source and returns the same tape.
- The resolved symbol/date range is persisted and exposed.
- A database-backed reproducibility test covers empty-price Historical requests.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #11, opened 2026-07-13
JumpDiffusion uses a normal sample as a Bernoulli draw


joaquinbejar:
## Summary

Jump occurrence is tested with a standard-normal sample instead of a uniform sample. The resulting jump probability is `Φ(lambda * dt)`, not `lambda * dt`.

## Evidence

As of `348611b`, `Walker::jump_diffusion` computes `lambda_dt` and then evaluates:

`self.normal_sample() < lambda_dt`

in `src/domain/walker.rs:215-229`.

For a common daily example `lambda = 1`, `dt = 1/252`, the intended Bernoulli probability is about 0.397%. The implemented probability is `Φ(0.00397)`, about 50.16%—over 100 times larger.

The same pattern exists upstream, but this crate owns a seeded reimplementation and exposes the behavior through its API.

## Impact

Jump-diffusion tapes have roughly one jump every two steps for small positive intensities, severely distorting returns, volatility, option prices, and backtests.

## Proposed direction

Draw `U ~ Uniform(0,1)` from the seeded RNG and compare it with a validated probability. For larger `lambda * dt`, consider a Poisson jump count rather than a Bernoulli approximation.

## Acceptance criteria

- Empirical jump frequency matches the documented model within statistical tolerance.
- The test is deterministic under a seed.
- Invalid intensity/time-step combinations are rejected or modeled explicitly.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #10, opened 2026-07-13
Validate HTTP simulation parameters instead of panicking during conversion


joaquinbejar:
## Summary

The request structs accept unconstrained `f64` and `usize` values, then infallible `From` conversions call the panicking `pos!` macro. Invalid client input can panic a request worker rather than return 400.

## Evidence

As of `348611b`:

- `CreateSessionRequest` and `UpdateSessionRequest` contain raw numeric fields without validation (`src/api/rest/requests.rs:9-49`, `110-154`).
- Conversion uses `pos!` for prices, expirations, volatilities, spreads, model parameters, custom timeframes, and historical prices (`src/session/model.rs:148-173`, `src/api/rest/models.rs:390-519`).
- `pos!` unwraps and panics for negative/non-representable values.
- Decimal conversions use `unwrap_or_default`, silently turning invalid/non-finite values into zero.
- `steps`, `chain_size`, and historical price vector length have no resource bounds.

A regression test converting a request with `initial_price = -1.0` panics with “Value must be positive”.

## Impact

Malformed requests can cause panics, silently altered parameters, invalid model domains, and resource-exhaustion requests.

## Proposed direction

Use fallible `TryFrom` conversions plus request validation. Reject non-finite numbers, invalid ranges/model invariants, zero/oversized steps and chain sizes, and oversized historical payloads with structured 400 responses.

## Acceptance criteria

- No user-controlled numeric value reaches `unwrap`/panic.
- Invalid values return a documented 400 error with field details.
- Reasonable upper bounds are configurable and tested.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #9, opened 2026-07-13
Simulation cache entries are never evicted


joaquinbejar:
## Summary

The simulator retains a complete `RandomWalk` per session indefinitely. Deleting or expiring a session only affects the session store, not the simulation cache.

## Evidence

As of `348611b`:

- `simulation_cache` is an unbounded `HashMap<Uuid, RandomWalk<...>>` (`src/domain/simulator.rs:36-40`).
- Entries are inserted when a session is first read (`src/domain/simulator.rs:94-108`).
- `SessionManager::delete_session` only calls `store.delete` (`src/session/manager.rs:246-248`).
- `cleanup_cache` exists but is marked unused and has no caller (`src/domain/simulator.rs:336-354`).
- Redis TTL cleanup cannot notify this process.

## Impact

Memory grows with every session ever read, proportional to `steps × chain_size`. Normal TTL expiry and DELETE do not reclaim it, enabling eventual memory exhaustion.

## Proposed direction

Evict on DELETE, completion, and store expiry; add a bounded TTL/LRU cache; and connect cache cleanup to session lifecycle. Wire the existing cache metrics to actual lookups.

## Acceptance criteria

- DELETE removes the corresponding walk.
- Completed/expired sessions are eventually evicted.
- Cache size has a configured hard bound or TTL.
- Tests assert eviction and metrics.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

PR #28, opened 2026-07-13
Serve-then-advance cursor semantics with persisted completion (#5)

### LINKED ISSUE #5: Step progression skips the initial snapshot and never persists completion
## Summary

The progression order advances the counter before indexing the generated walk. This skips index 0 and makes the final request index one past the end, so the completed state is never saved.

## Evidence

As of `348611b`:

1. `SessionManager::get_next_step` calls `advance_state` before `simulate_next_step` (`src/session/manager.rs:131-144`).
2. `advance_step` increments `current_step` and marks the session completed when it equals `total_steps` (`src/session/model.rs:285-301`).
3. The simulator indexes `random_walk[current_step]` and rejects `current_step >= random_walk.len()` (`src/domain/simulator.rs:121-130`).
4. The walk is created with `size = params.steps` (`src/domain/simulator.rs:305-325`).

For a two-step session, the first GET returns step 1. The second GET attempts index 2, fails, and the save is skipped; the persisted session remains at step 1 / `InProgress`.

## Impact

- The initial chain snapshot is never returned for a new session.
- Clients receive fewer successful snapshots than requested.
- The session never persists `Completed`.
- Later requests repeatedly fail from the same persisted pre-completion state.

## Proposed direction

Make cursor semantics explicit and align counter, walk length, and save order. Persist the terminal transition even when no further snapshot exists.

## Acceptance criteria

- A session configured for N snapshots returns exactly N documented snapshots.
- Index 0 handling is explicit and tested.
- Completion is persisted once and subsequent reads return the documented terminal response.

## Revalidation

joaquinbejar:
## Summary

The progression advanced the cursor BEFORE indexing the walk, so a session
never served snapshot 0, returned N-1 snapshots instead of N, and the final
request failed before the save — `Completed` was never persisted and the
session stayed wedged at the same pre-completion state forever. The manager
now uses serve-then-advance cursor semantics.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 6**
- **Base branch:** `stack/5-21-safe-get-step-endpoints`
- **Stacked on:** #27 · **Blocks:** the next PR in the stack (#4)

## Changes

- `SessionManager::get_next_step`: `store.get` → completed-guard (terminal 410
  without touching the store) → simulate at the PRE-advance cursor (serves
  index `current_step`) → `advance_state` → `store.save` ALWAYS after a served
  snapshot, including the transition to `Completed`. `current_step` is now
  explicitly the index of the next snapshot to serve (0-based).
- Simulator rebuild condition: `!contains_key || Reinitialized` (dropped the
  `current_step == 0` trigger — a strict subset, so strictly fewer rebuilds;
  peek(0) and the first advance now read the SAME cached walk).
- utoipa + crate docs + README state the exact semantics: N steps ⇒ exactly
  snapshots 0..N-1 over N advances; the N-th advance persists `Completed`;
  further calls return 410.

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 288 + 2 passed, 0 failed
- [x] N-step session serves EXACTLY N snapshots incl. index 0; terminal call is non-mutating; completion persisted once (`updated_at` asserted unchanged after terminal call)
- [x] Peek/advance serve the same index across the whole tape
- [x] Manager-level same-seed tape test (full price-sequence equality) + domain tape tests green

## Review

`reproducibility-reviewer` audited the diff: **PASS (0 🔴, 0 🟠)** — walker
untouched, rebuild-condition change is a strict subset (no mid-tape walk
swap), completed-guard non-mutating, peek/advance consistency proven by
tests. Its two pre-existing findings are exactly the next stack items
(#4 Reinitialized no-op; #12 unseeded historical selection).

Closes #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,11 +91,18 @@ The OptionChain-Simulator exposes the following REST API endpoints:
 | Method | Endpoint           | Action              | Description                                                      |
 |--------|--------------------|---------------------|------------------------------------------------------------------|
 | POST   | /api/v1/chain      | Create Session      | Creates a new simulation session                                 |
-| GET    | /api/v1/chain      | Read Current Step   | Returns the current snapshot WITHOUT advancing (safe, repeatable)|
-| POST   | /api/v1/chain/step | Advance Step        | Advances one step, serves the snapshot, and persists the advance |
+| GET    | /api/v1/chain      | Read Current Step   | Returns the snapshot the next advance will serve (safe, repeatable)|
+| POST   | /api/v1/chain/step | Advance Step        | Serves the snapshot at the cursor (index 0 first), then advances  |
 | PUT    | /api/v1/chain      | Replace Session     | Completely replaces session parameters                           |
 | PATCH  | /api/v1/chain      | Update Parameters   | Updates specific session parameters                              |
 | DELETE | /api/v1/chain      | Delete Session      | Terminates and removes a session                                 |
+
+**Step cursor semantics (serve-then-advance):** `current_step` is the 0-based index
+of the NEXT snapshot to serve. `POST /api/v1/chain/step` serves the snapshot at the
+cursor and then advances it, so a session with `steps = N` serves EXACTLY indices
+`0..N-1` over `N` advances. The advance that serves the last snapshot persists the
+`Completed` state, and any further advance returns `410 Gone`. `GET /api/v1/chain`
+peeks the snapshot the next advance would serve without moving the cursor.
 
 > **Migration (breaking, issue #21):** `GET /api/v1/chain` used to advance the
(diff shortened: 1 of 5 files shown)


joaquinbejar (commented):
The incremental cursor change is sound, its CI checks pass, and the full stacked workspace validation is clean. Intended decision APPROVE, submitted as COMMENT because GitHub does not allow approving a pull request authored by the current account.

joaquinbejar:
Thanks. Rebased onto main, the advance description now also documents the expected_step retry precondition added on PR 27.

Issue #8, opened 2026-07-13
Concurrent session mutations lose updates


joaquinbejar:
## Summary

Session advancement, PATCH, and PUT use a non-atomic read-modify-write sequence. Concurrent requests can read the same snapshot and overwrite each other.

## Evidence

As of `348611b`, `get_next_step`, `update_session`, and `reinitialize_session` all perform separate `store.get(id)` and `store.save(session)` calls (`src/session/manager.rs:130-144`, `172-182`, `214-224`). The `SessionStore` contract has no compare-and-set, version, lock, or atomic mutation operation.

A deterministic barrier-backed store test starts two advances from step 0. Both responses report step 1 and the persisted session ends at step 1, so one advancement is lost.

Redis makes this possible across workers and replicas because GET and SET are separate commands.

## Impact

Parallel GETs can return duplicate steps. PATCH/PUT can overwrite progress or each other, breaking tape ordering and reproducibility.

## Proposed direction

Add optimistic concurrency (revision + CAS/WATCH/MULTI), a store-level atomic mutation primitive, or a per-session/distributed lock. Define the expected conflict response.

## Acceptance criteria

- Two concurrent advances produce two distinct, ordered steps or one explicit conflict.
- PATCH/PUT cannot silently overwrite a newer revision.
- Tests cover in-process and Redis-backed concurrency.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

Issue #6, opened 2026-07-13
PATCH steps updates parameters.steps but leaves total_steps unchanged


joaquinbejar:
## Summary

Changing `steps` through PATCH creates contradictory session metadata: `parameters.steps` changes, while `total_steps` retains the value from session creation.

## Evidence

As of `348611b`:

- The PATCH handler copies `json_req.steps` into `updated_params.steps` (`src/api/rest/handlers.rs:366-368`).
- `SessionManager::update_session` calls `modify_parameters` and then resets progression (`src/session/manager.rs:173-182`).
- `Session::modify_parameters` replaces `parameters` but does not update `total_steps` (`src/session/model.rs:321-325`).
- In contrast, `Session::reinitialize` explicitly sets `total_steps = new_params.steps` (`src/session/model.rs:352-357`).

A regression test changing steps from 2 to 5 observes `parameters.steps == 5` and `total_steps == 2`.

## Impact

The regenerated walk and the state machine use different limits. API responses advertise inconsistent values and the session may terminate early or index beyond the intended tape.

## Acceptance criteria

- Updating `steps` keeps `parameters.steps` and `total_steps` equal.
- The generated walk and progression limit use the same value.
- Add a PATCH regression test that changes both upward and downward.

## Revalidation

Revalidated on `4b49eca` after the workspace dependency migration; the root cause is still present.

PR #29, opened 2026-07-13
Transition sessions out of Reinitialized on first advance (#4)

### LINKED ISSUE #4: Reinitialized sessions remain stuck at step 0 forever
## Summary

After a session is replaced with PUT or updated with PATCH, every subsequent GET returns step 0 and the session remains in `Reinitialized`.

## Evidence

As of commit `348611b`:

- `SessionManager::reinitialize_session` and `update_session` call `reset_progression` (`src/session/manager.rs:179-182`, `221-224`).
- `reset_progression` sets `current_step = 0` and `state = Reinitialized` (`src/session/state_handler.rs:45-49`).
- The `Reinitialized` arm in `advance_state` returns `Ok(())` without changing either field (`src/session/state_handler.rs:32-35`).
- `Simulator::simulate_next_step` rebuilds the walk whenever the state is `Reinitialized` (`src/domain/simulator.rs:81-90`).

A regression test that performs PUT-equivalent reinitialization followed by two calls to `get_next_step` observes `current_step == 0` and `state == Reinitialized` both times.

## Impact

PUT and PATCH effectively make a session unusable for continued playback. They also rebuild the complete random walk on every GET.

## Proposed direction

Define the intended first-read behavior after reset, then transition out of `Reinitialized` exactly once. For example, return step 0 and persist `InProgress`, or advance to step 1 consistently with newly created sessions.

## Acceptance criteria

- A reset session leaves `Reinitialized` after the first successful GET.
- Consecutive GETs return monotonically increasing step numbers.
- A regression test covers PUT and PATCH followed by at least two GETs.

## Revalidation

joaquinbejar:
## Summary

After PUT or PATCH, sessions were stuck forever: `advance_state`'s
`Reinitialized` arm was a no-op (never advanced, never changed state), so
every subsequent request re-evicted and rebuilt the FULL walk, served index 0
again, and persisted the same `Reinitialized`/0 state. The `Reinitialized`
transition now mirrors `Initialized`, and the manager's PUT/PATCH paths drop
their redundant double-reset.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 7**
- **Base branch:** `stack/6-5-step-cursor-semantics`
- **Stacked on:** #28 · **Blocks:** the next PR in the stack (#6)

## Changes

- `advance_state` `Reinitialized` arm: `state = InProgress` + `advance_step()`
  — under serve-then-advance the first post-reset advance rebuilds the walk
  ONCE (eviction trigger sees pre-advance `Reinitialized`), serves index 0 of
  the new tape, and persists `InProgress`/cursor 1; later advances hit the
  cached walk.
- `update_session` (PATCH) and `reinitialize_session` (PUT) both become
  `get → session.reinitialize(params) → save` — the `modify_parameters`+
  `reset_progression` clobber and the double reset are gone; PATCH explicitly
  restarts the tape and syncs `total_steps` via `reinitialize`.
- Dead `reset_progression` removed (no callers); `reinitialize_session` doc
  fixed (no phantom `total_steps` parameter).

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 290 + 2 passed, 0 failed
- [x] PUT + two advances → cursor 1 then 2, state leaves `Reinitialized` after ONE advance; same for PATCH
- [x] Reinitializing with a different seed provably changes the tape (full-sequence comparison; divergence from index 1 — index 0 is the seed-independent initial chain)
- [x] All existing tape/serde/manager tests green

## Checklist

- [x] Stored-session serde shape unchanged; effective-seed guarantee holds through PUT/PATCH
- [x] Walker untouched

Closes #4

### DIFF
--- a/src/session/manager.rs
+++ b/src/session/manager.rs
@@ -222,7 +222,14 @@ impl SessionManager {
         Ok((session, chain))
     }
 
-    /// Updates an existing simulation session with new parameters.
+    /// Updates an existing simulation session with new parameters (PATCH).
+    ///
+    /// A parameter change restarts the session's tape: the new parameters may alter
+    /// the walk (steps, volatility, seed, ...), so the cursor is reset to 0 and the
+    /// session is marked `Reinitialized`. The next [`SessionManager::get_next_step`]
+    /// then rebuilds the walk from the new parameters and serves index 0. This is
+    /// implemented via [`Session::reinitialize`], which also syncs `total_steps` to
+    /// `params.steps` so a PATCH that changes the step count stays coherent.
     ///
     /// # Parameters
     ///
@@ -248,11 +255,9 @@ impl SessionManager {
     ) -> Result<Session, ChainError> {
         let mut session = self.store.get(id)?;
 
-        // Update parameters
-        session.modify_parameters(params);
-
-        // Reset progression
-        self.state_handler.reset_progression(&mut session)?;
+        // A parameter change restarts the tape: reset the cursor, sync total_steps,
+        // and mark the session Reinitialized so the next advance rebuilds the walk.
+        session.reinitialize(params);
 
         // Save updated session
         self.store.save(session.clone())?;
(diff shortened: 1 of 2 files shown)


joaquinbejar (commented):
The incremental change is sound, its CI checks pass, and the full stacked workspace validation is clean. Intended decision APPROVE, submitted as COMMENT because GitHub does not allow approving a pull request authored by the current account.

joaquinbejar:
Thanks. Rebased onto main and carried the advance-path Error guard promised on PR 27 here, rejected before simulating with a store-untouched regression test.

PR #31, opened 2026-07-13
Tri-state PATCH semantics with patchable skew_slope (#20)

joaquinbejar:
## Summary

The PATCH representation could not express valid state changes: it had no
`skew_slope` field at all, and serde collapsed both "key absent" and JSON
`null` into `None`, which the handler read as "keep" — so optional parameters
(`chain_size`, `strike_interval`, `smile_curve`, `spread`, `seed`) could never
be cleared. PATCH now uses a tri-state representation: absent = keep, null =
clear, value = replace.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 9**
- **Base branch:** `stack/8-6-patch-total-steps`
- **Stacked on:** #30 · **Blocks:** the next PR in the stack (#19)
- The diff is cumulative until the lower PRs merge — review against the base branch.

## Changes

- New `Patch<T>` (`Absent | Null | Value(T)`, manual serde, no new deps):
  absence via `#[serde(default)]`, null → `Null`, value → `Value`;
  `Absent` omitted on serialize.
- `UpdateSessionRequest`: domain-optional fields become `Patch<T>` and
  **`skew_slope` is now patchable** (new field). Domain-required fields stay
  `Option<T>` (absent = keep).
- Extracted, unit-testable `apply_update(&mut SimulationParameters,
  &UpdateSessionRequest)` merge: `Absent` keeps, `Null` clears, `Value`
  validates (reusing the #10 validators + bounds) and replaces.
  `seed: null` re-seeds with a fresh random seed (documented) — the
  effective-seed invariant (`parameters.seed` always `Some`) is preserved.
- utoipa: per-field `#[schema(value_type = Option<T>)]` (avoids generic-name
  `$ref` collisions in utoipa 5) + tri-state docs + a PATCH request example
  showing all three states.

## Contract impact

Additive / wire-backward-compatible: omitting fields and sending values
behave exactly as before. New opt-in behaviors: explicit `null` clears an
optional field (previously a no-op), `skew_slope` newly accepted, `seed: null`
re-seeds. IronCondor does not consume the PATCH DTO (checked) — semver-minor.

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 313 + 2 passed, 0 failed (21 new tests)
- [x] Serde tri-state round-trips per field (absent/null/value, both directions)
[...]

PR #32, opened 2026-07-13
Async Redis store over a multiplexed connection manager (#19)

### LINKED ISSUE #19: Synchronous Redis I/O blocks async HTTP workers and serializes all sessions

joaquinbejar:
## Summary

The async actix server drove Redis through the SYNCHRONOUS `Commands` API on a
single `redis::Connection` behind a `std::sync::Mutex`: every store operation
blocked an executor worker thread and all sessions contended on one
connection — head-of-line blocking that cannot scale with workers. The Redis
path is now fully async over a multiplexed `ConnectionManager`, and the
`SessionStore` trait and `SessionManager` are async end to end.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 10**
- **Base branch:** `stack/9-20-patch-tristate`
- **Stacked on:** #31 · **Blocks:** the next PR in the stack (#8)

## Changes

- redis dep gains `tokio-comp` + `connection-manager` features (pre-approved
  in the stack plan; no new crates). `RedisClient` holds a cloneable
  `redis::aio::ConnectionManager` (multiplexed, auto-reconnect, NO mutex);
  all methods async via `redis::AsyncCommands`; `set_nx` preserved.
- Timeouts configurable: `REDIS_TIMEOUT` (response, default 30s) and
  `REDIS_CONNECT_TIMEOUT` (default 5s) applied through
  `ConnectionManagerConfig`; invalid env values warn + default.
- `SessionStore` → `#[async_trait]` (dep already present); in-memory impl
  keeps its `std::sync::Mutex` for map access only — never held across an
  `.await`; Redis impl awaits the client.
- `SessionManager` mutation/read methods async; handlers await; `main.rs` and
  the three examples updated. #15's credential redaction preserved on the new
  connection path.

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 306 + 2 passed, 0 failed (no live Redis needed — store doubles)
- [x] Concurrency: `tokio::join!` on two creates (store + manager level) completes without serializing on a shared lock; `RedisClient: Clone + Send + Sync` compile-level guarantee
- [x] Timeout env parsing fallback tested

## Checklist

- [x] No lock held across `.await` anywhere on the store path
- [x] Driver errors still convert to `ChainError` at the adapter
- [x] Dependency change is features-only on the existing redis dep, in `[workspace.dependencies]`
- [x] No `.unwrap()` / `.expect()` on production paths; `tracing` only; no `unsafe`

Closes #19

### DIFF
--- a/src/infrastructure/config/redis.rs
+++ b/src/infrastructure/config/redis.rs
@@ -26,6 +26,29 @@ pub struct RedisConfig {
     /// attempt may block before the manager retries.
     pub connect_timeout: u64,
 }
+/// Parses a timeout environment variable in whole seconds.
+///
+/// The value must be a positive integer: zero would disable the bound entirely
+/// (a hung server could then hold a worker forever), so `0`, non-numeric, and
+/// unset values all fall back to `default` — invalid ones with a warning.
+fn parse_timeout_secs(var: &str, default: u64) -> u64 {
+    match env::var(var) {
+        Ok(raw) => match raw.parse::<u64>() {
+            Ok(v) if v >= 1 => v,
+            _ => {
+                tracing::warn!(
+                    "invalid {} value {:?}; must be an integer >= 1, using default {}s",
+                    var,
+                    raw,
+                    default
+                );
+                default
+            }
+        },
+        Err(_) => default,
+    }
+}
+
 impl RedisConfig {
     pub(crate) fn url(&self) -> String {
         // Start building the URL
@@ -78,15 +101,8 @@ impl Default for RedisConfig {
         let username = env::var("REDIS_USER").ok();
         let password = env::var("REDIS_PASSWORD").ok();
 
-        let timeout = env::var("REDIS_TIMEOUT")
-            .ok()
-            .and_then(|s| s.parse::<u64>().ok())
-            .unwrap_or(30);
-
-        let connect_timeout = env::var("REDIS_CONNECT_TIMEOUT")
-            .ok()
-            .and_then(|s| s.parse::<u64>().ok())
-            .unwrap_or(5);
+        let timeout = parse_timeout_secs("REDIS_TIMEOUT", 30);
+        let connect_timeout = parse_timeout_secs("REDIS_CONNECT_TIMEOUT", 5);
 
         Self {
             host: env::var("REDIS_HOST").unwrap_or_else(|_| "localhost".to_string()),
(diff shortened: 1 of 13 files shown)


src/session/store/interface.rs
```diff
@@ -67,7 +69,27 @@ pub trait SessionStore: Send + Sync {
     /// - The session with the provided `id` does not exist.
     /// - There is an issue with the underlying storage or retrieval process.
     ///
-    fn get(&self, id: Uuid) -> Result<Session, ChainError>;
+    async fn get(&self, id: Uuid) -> Result<Session, ChainError>;
```
joaquinbejar, line 72: 🟠 **Warning**: making every `SessionStore` method async breaks all downstream implementations and call sites of this public trait. The async conversion is appropriate, but the PR needs an explicit breaking-release and consumer-migration plan instead of treating this as an internal implementation change.

src/infrastructure/config/redis.rs
```diff
@@ -68,26 +78,45 @@ impl Default for RedisConfig {
         let username = env::var("REDIS_USER").ok();
         let password = env::var("REDIS_PASSWORD").ok();
 
+        let timeout = env::var("REDIS_TIMEOUT")
+            .ok()
+            .and_then(|s| s.parse::<u64>().ok())
+            .unwrap_or(30);
+
+        let connect_timeout = env::var("REDIS_CONNECT_TIMEOUT")
```
joaquinbejar: 🟠 **Warning**: these new environment variables are not documented in `.env.example` or the generated README source, and zero or invalid values are silently accepted or defaulted. Document both knobs and validate their ranges at startup as required by the repository configuration contract.

joaquinbejar (commented):
The asynchronous Redis design is appropriate, but it introduces an undocumented breaking trait migration and configuration surface. Intended decision REQUEST_CHANGES, submitted as COMMENT because GitHub does not allow requesting changes on a pull request authored by the current account.

src/session/store/interface.rs
joaquinbejar: Agreed, this is part of the 0.1.0 breaking release and the notes will call out the async SessionStore trait and the RedisClient signature changes for downstream implementors. The wire behavior is unchanged.

src/infrastructure/config/redis.rs
joaquinbejar: Zero and invalid values are now rejected with a warning and the documented defaults in b830946, both knobs are documented on the config struct fields which is the crate's configuration reference, there is no .env.example in this repo.

joaquinbejar:
Timeout validation added in b830946, 0.1.0 breaking-release plan covers the trait migration. Rebased onto main.

PR #33, opened 2026-07-13
Optimistic concurrency (CAS) for session mutations (#8)

### LINKED ISSUE #8: Concurrent session mutations lose updates

joaquinbejar:
## Summary

Session advancement, PATCH, and PUT used a non-atomic read-modify-write: two
concurrent requests could read the same snapshot and silently overwrite each
other — duplicate steps, lost progress, broken tape ordering. Mutations now go
through optimistic concurrency: a `version` on the session and an atomic
compare-and-set in the store; a stale writer gets an explicit HTTP 409.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 11**
- **Base branch:** `stack/10-19-async-redis-store`
- **Stacked on:** #32 · **Blocks:** the next PR in the stack (#9)

## Changes

- `Session.version: u64` with `#[serde(default)]` (old stored JSON loads at 0
  — round-trip tested); `Session::bump_version()`; model transitions stay
  pure — the manager bumps right before the CAS write.
- New `SessionStore::save_cas(session, expected_version)`: in-memory under the
  map mutex; Redis via a server-side Lua script (`cjson.decode` of the stored
  JSON's `version`; `-1` → NotFound, `-2` → Conflict, `1` → SET with TTL) —
  atomic on the server, works with the multiplexed `ConnectionManager`.
- New `ChainError::Conflict` → HTTP 409; documented on POST /step, PUT, PATCH.
- `get_next_step` / `update_session` / `reinitialize_session`: read version →
  mutate → bump → `save_cas`; conflicts propagate (no silent retry).
  `create_session` stays on the NX-guarded `create`.

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 318 + 2 passed, 0 failed
- [x] **Barrier-forced acceptance test**: two concurrent advances from cursor 0 (both reads rendezvous before either write) → exactly ONE succeeds at cursor 1, one gets `Conflict`; persisted session cursor 1 / version 1 — no lost update, no duplicate step
- [x] PATCH-vs-advance concurrent conflict test
- [x] `save_cas` unit tests (commit / stale-leaves-store-unchanged / missing); Lua contract + return-code mapping unit-tested; old-JSON compat test

## Checklist

- [x] Stored-session compatibility: `version` carries `#[serde(default)]`
- [x] Redis CAS is a single server-side script (no get/set race); no lock across `.await`
- [x] Seed plumbing and walk/simulator untouched
- [x] No new dependencies; no `unsafe`; `tracing` only

Note: live-Redis integration of the Lua script is outside the hermetic suite;
it is mirrored by the store double + script/mapping unit tests.

Closes #8

### DIFF
--- a/src/session/model.rs
+++ b/src/session/model.rs
@@ -260,6 +260,15 @@ pub struct Session {
     pub total_steps: usize,
     /// * `state` - The current state of the session, represented by an instance
     pub state: SessionState,
+    /// * `version` - Optimistic-concurrency revision counter. It starts at `0` at
+    ///   creation and is bumped by the [`SessionManager`] (via
+    ///   [`Session::bump_version`]) immediately before a compare-and-swap save, so
+    ///   two concurrent mutations that read the same snapshot cannot silently
+    ///   overwrite one another: the second save fails with
+    ///   [`ChainError::Conflict`]. Carries `#[serde(default)]` so sessions
+    ///   persisted before this field existed still deserialize (defaulting to 0).
+    #[serde(default)]
+    pub version: u64,
 }
 
 impl Session {
@@ -296,6 +305,7 @@ impl Session {
             total_steps: parameters.steps,
             parameters,
             state: SessionState::Initialized,
+            version: 0,
         }
     }
 
@@ -349,6 +359,7 @@ impl Session {
             total_steps: parameters.steps,
             parameters,
             state: SessionState::Initialized,
+            version: 0,
         }
     }
 
@@ -457,6 +468,18 @@ impl Session {
     pub fn is_active(&self) -> bool {
         self.state != SessionState::Completed && self.state != SessionState::Error
     }
+
+    /// Increments the optimistic-concurrency [`Session::version`] counter by one.
+    ///
+    /// The state-transition methods (`advance_step`, `modify_parameters`,
+    /// `reinitialize`) are deliberately kept as pure state changes and do NOT
+    /// touch the version. The [`SessionManager`] calls this helper right before a
+    /// compare-and-swap save so the persisted revision advances exactly once per
+    /// successfully served mutation. Uses `saturating_add` so an (unreachable in
+    /// practice) overflow cannot panic on a production path.
+    pub fn bump_version(&mut self) {
+        self.version = self.version.saturating_add(1);
+    }
 }
 
 impl Default for Session {
@@ -496,6 +519,7 @@ impl Default for Session {
             current_step: 0,
             total_steps: 20,
             state: SessionState::Initialized,
+            version: 0,
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/session/store/in_redis.rs
```diff
@@ -17,6 +18,61 @@ pub struct InRedisSessionStore {
...
+/// `ARGV[2]` = expected version; `ARGV[3]` = TTL in seconds.
+///
+/// Return codes: `-1` = key missing (NotFound); `-2` = version mismatch
+/// (Conflict); `1` = written.
+const SAVE_CAS_SCRIPT: &str = r#"
+local cur = redis.call('GET', KEYS[1])
+if not cur then
+    return -1
+end
+local v = cjson.decode(cur)['version'] or 0
```
joaquinbejar: 🟠 **Warning**: Redis Lua `cjson` represents numbers as IEEE-754 doubles, so this cannot compare every `u64` revision exactly. Above `2^53`, adjacent versions collapse and a stale writer can pass the CAS; this failure reproduces against Redis. Store the revision as an exactly comparable Redis integer or string, or enforce a safe upper bound.

src/session/model.rs
```diff
@@ -369,6 +468,18 @@ impl Session {
...
+    /// Increments the optimistic-concurrency [`Session::version`] counter by one.
+    ///
+    /// The state-transition methods (`advance_step`, `modify_parameters`,
+    /// `reinitialize`) are deliberately kept as pure state changes and do NOT
+    /// touch the version. The [`SessionManager`] calls this helper right before a
+    /// compare-and-swap save so the persisted revision advances exactly once per
+    /// successfully served mutation. Uses `saturating_add` so an (unreachable in
+    /// practice) overflow cannot panic on a production path.
+    pub fn bump_version(&mut self) {
+        self.version = self.version.saturating_add(1);
```
joaquinbejar: 🟠 **Warning**: saturating here eventually freezes the revision at `u64::MAX`, after which multiple writers can pass CAS with the same version. Saturating arithmetic is also explicitly prohibited by the repository rules. Use `checked_add` and return a typed error before attempting the save.

src/session/store/in_redis.rs
```diff
@@ -129,12 +231,48 @@ impl SessionStore for InRedisSessionStore {
...
+        // Run the compare-and-swap atomically server-side. The manager (the block
+        // scheduler) supplies `expected_version`; the script only writes when the
+        // stored revision still matches, so a concurrent advance cannot be lost.
+        let mut conn = self.client.connection_manager();
+        let code: i64 = redis::Script::new(SAVE_CAS_SCRIPT)
+            .key(&key)
+            .arg(json_str)
+            .arg(expected_version)
+            .arg(self.session_ttl)
+            .invoke_async(&mut conn)
```
joaquinbejar, line 303: 🟠 **Warning**: this Lua script is the production atomicity boundary, but no test executes it against Redis. String checks and the in-memory mirror cannot catch Lua, argument, driver, or return-value incompatibilities. Add a live Redis test that forces two concurrent CAS writes and verifies exactly one commit.

joaquinbejar (commented):
The CAS closes the ordinary lost-update race, but its revision representation, overflow behavior, and production Redis boundary remain unsafe or unverified. Intended decision REQUEST_CHANGES, submitted as COMMENT because GitHub does not allow requesting changes on a pull request authored by the current account.

src/session/store/in_redis.rs
joaquinbejar: Right, cjson doubles collapse past 2^53. The script now compares an exact version string held in a companion {key}:ver Redis key, maintained atomically inside the same script and seeded by create and save, legacy sessions without the key default to 0. Done in 5eb3598.

src/session/model.rs
joaquinbejar: Fixed in 5eb3598, bump_version uses checked_add and returns a typed Internal error on overflow before any save is attempted, with a test.

src/session/store/in_redis.rs
joaquinbejar: Added test_save_cas_is_atomic_against_live_redis in 5eb3598, ignored by default and runnable with -- --ignored, two concurrent CAS writes against a real Redis produce exactly one commit and one conflict. Verified locally against redis:7 four times, and the CI integration job picks it up once the redis service lands later in the stack.

joaquinbejar:
All three addressed in 5eb3598. Rebased onto main.

PR #34, opened 2026-07-13
Bound the simulation cache and evict with the lifecycle (#9)

### LINKED ISSUE #9: Simulation cache entries are never evicted

joaquinbejar:
## Summary

The simulator kept one full `RandomWalk` per session in an unbounded map
forever: DELETE and Redis TTL expiry only touched the session store, never the
walk cache, so memory grew with every session ever read (∝ steps ×
chain_size) until exhaustion. The cache is now bounded (LRU) and wired to the
session lifecycle.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 12**
- **Base branch:** `stack/11-8-session-cas`
- **Stacked on:** #33 · **Blocks:** the next PR in the stack (#11)

## Changes

- Cache entries carry `last_access` (refreshed on every hit); inserts enforce
  `OCS_MAX_CACHED_WALKS` (default 1000, parse-once with warn+default) by
  evicting the least-recently-accessed entry first.
- Eviction triggers: DELETE (`SessionManager::delete_session` →
  `Simulator::remove_session`), completion (after the CAS save that persists
  `Completed`), `Reinitialized` (pre-existing), LRU bound on insert.
- Dead `cleanup_cache` removed (no caller; superseded by bound + lifecycle
  eviction). New `simulation_cache_size` Prometheus gauge set from the
  advance/delete handlers.
- Lock discipline preserved: the ClickHouse-touching walk build still runs
  with the cache lock dropped; capacity enforcement does no `.await`.

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 325 + 2 passed, 0 failed
- [x] DELETE evicts; completion evicts (steps=2 → cache empty after 2 advances); LRU bound evicts the oldest-accessed (staggered-timestamps unit test)
- [x] Reproducibility: eviction + re-simulate rebuilds the SAME seeded tape (asserted); all existing tape tests green

## Checklist

- [x] Layering respected: the domain does not import `api` (local parse-once limit, same pattern)
- [x] Same-seed ⇒ same-tape contract intact
- [x] No lock held across `.await`; no new dependencies; no `unsafe`

Note: peek also lazily populates the cache but its handler carries no metrics
collector; the gauge updates on advance/delete where size actually changes —
flagged as a possible follow-up.

Closes #9

### DIFF
--- a/src/domain/simulator.rs
+++ b/src/domain/simulator.rs
@@ -23,7 +23,8 @@ use rand::RngExt;
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
+use std::time::Instant;
 use tokio::sync::Mutex;
 use tracing::{debug, error, info, instrument, warn};
 use uuid::Uuid;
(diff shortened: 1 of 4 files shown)


src/domain/simulator.rs
```diff
@@ -33,10 +34,80 @@ const DEFAULT_CHAIN_SIZE: usize = 30;
...
+            }
+            None => break,
+        }
+    }
+}
+
+/// Simulator handles the generation of option chains based on simulation parameters.
+///
+/// It owns a bounded, per-session cache of [`RandomWalk`]s keyed by session id. The
+/// cache is evicted on three lifecycle triggers so it never outlives the sessions
```
joaquinbejar, line 126: 🔴 **Critical**: this can evict an active historical session, but this PR still rebuilds historical data using `rand::rng()` for symbol and date selection. The next request can therefore continue on a different tape despite the same session seed. Land the seeded historical-selection fix before enabling active LRU eviction, or exclude historical walks until rebuilds are deterministic.

src/domain/simulator.rs
```diff
@@ -99,35 +175,51 @@ impl Simulator {
...
+            );
         }
 
         // Get the current step data
         let step = {
-            let cache = self.simulation_cache.lock().await;
+            let mut cache = self.simulation_cache.lock().await;
 
-            let random_walk = cache.get(&session.id).ok_or_else(|| {
+            let entry = cache.get_mut(&session.id).ok_or_else(|| {
```
joaquinbejar: 🟠 **Warning**: another cold insertion can evict this session after its insert and before this lookup, causing a valid request to return an internal error at capacity. Clone the requested step in the same critical section that inserts or refreshes the entry, or add per-session in-flight coordination and a concurrent capacity regression test.

src/domain/simulator.rs
```diff
@@ -99,35 +175,51 @@ impl Simulator {
...
 
-            // Insert the new random walk into the cache
+            // Insert the new random walk into the cache, enforcing the LRU bound
+            // first. Evicting down to `max - 1` before inserting keeps the cache at
+            // or below `MAX_CACHED_WALKS` afterwards; the id being inserted is absent
+            // here (a Reinitialized session was removed above), so it is never evicted.
             let mut cache = self.simulation_cache.lock().await;
-            cache.insert(session.id, random_walk);
+            let max = *MAX_CACHED_WALKS;
+            enforce_capacity(&mut cache, max.saturating_sub(1));
```
joaquinbejar: 🟠 **Warning**: this introduces saturating arithmetic even though the repository rules prohibit it. `MAX_CACHED_WALKS` is already validated as at least 1, so use `max - 1` here, or `checked_sub` if the invariant should be surfaced explicitly.

joaquinbejar (commented):
The hard cache bound addresses growth, but the incremental eviction design can break reproducibility, fail concurrent cold requests, and violates the arithmetic rule. Intended decision REQUEST_CHANGES, submitted as COMMENT because GitHub does not allow requesting changes on a pull request authored by the current account.

src/domain/simulator.rs
joaquinbejar: Good ordering catch. Historical walks are now pinned against LRU eviction at this level (evictable flag on the cache entry, warn when the bound cannot be honored), and the seeded-historical PR stacked above lifts the pin once rebuilds are deterministic. Done in da1d859.

src/domain/simulator.rs
joaquinbejar: Fixed in da1d859, capacity enforcement, insert, range check, and step clone now happen in one critical section (the ClickHouse-touching build stays outside any lock), so a concurrent cold insert can no longer evict the entry between insert and lookup.

src/domain/simulator.rs
joaquinbejar: Done in da1d859, plain max - 1 with a debug_assert citing the parse-time >= 1 invariant, saturating_sub is gone.

joaquinbejar:
All three addressed in da1d859. Rebased onto main.

PR #35, opened 2026-07-13
Fix JumpDiffusion Bernoulli draw (#11)

### LINKED ISSUE #11: JumpDiffusion uses a normal sample as a Bernoulli draw

joaquinbejar:
## Summary

`jump_diffusion` tested jump occurrence with a STANDARD NORMAL sample against
the probability (`normal_sample() < λ·dt`), implementing probability Φ(λ·dt)
instead of λ·dt — for λ=1, dt=1/252 that's ~50.2% per step instead of ~0.4%,
i.e. a jump every two steps. The occurrence draw is now a uniform
Bernoulli(λ·dt) test from the seeded RNG, with λ·dt ≥ 1 rejected.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 13**
- **Base branch:** `stack/12-9-cache-eviction`
- **Stacked on:** #34 · **Blocks:** the next PR in the stack (#12)

## Changes

- `Walker::uniform_sample()` ([0,1) from the seeded RNG) and the testable
  `bernoulli_jump(λ·dt)` helper; `jump_diffusion` uses it for occurrence — the
  jump SIZE draw stays normal (correct).
- `λ·dt ≥ 1` fails fast with a walk error (Poisson jump counts documented as
  out of scope).
- **Documented intentional divergence from upstream**: optionstratlib's kernel
  has the same normal-as-Bernoulli bug; the seeded mirror now implements the
  documented model. Comment at the site + `CLAUDE.md` mirror-rule updated so
  upstream re-syncs preserve the fix. (Verified the rest of the kernel stays
  byte-for-byte with upstream, and `generate_with_vol` routes JumpDiffusion
  through this override.)

## Contract impact

**Tape-breaking for JumpDiffusion seeds vs the previous release** — the same
seed now yields a different (correct) tape. Same-seed ⇒ same-tape holds within
the build (tested). Other walk types' tapes are unaffected (occurrence draw is
the only behavioral change; verified in the diff).

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 330 + 2 passed, 0 failed
- [x] Deterministic empirical-frequency test: 100k draws at p=0.004 under a fixed seed → observed 0.00416 ∈ [0.002, 0.006] (the old code would give ≈0.5)
- [x] λ·dt ≥ 1 rejected; same-seed JumpDiffusion tape identical (walker + full Simulator path); different seeds differ
- [x] All pre-existing tape tests green; clippy `-D warnings` + fmt clean

Closes #11

### DIFF
--- a/src/domain/walker.rs
+++ b/src/domain/walker.rs
@@ -2,8 +2,8 @@ use optionstratlib::chains::OptionChain;
 use optionstratlib::error::SimulationError;
 use optionstratlib::simulation::{WalkParams, WalkPath, WalkType, WalkTypeAble};
 use positive::Positive;
-use rand::SeedableRng;
 use rand::rngs::StdRng;
+use rand::{RngExt, SeedableRng};
 use rand_distr::{Distribution, StandardNormal};
 use rust_decimal::prelude::{FromPrimitive, ToPrimitive};
 use rust_decimal::{Decimal, MathematicalOps};
@@ -37,6 +37,29 @@ impl Walker {
         let mut rng = self.rng.lock().unwrap();
         let z: f64 = StandardNormal.sample(&mut *rng);
         Decimal::from_f64(z).unwrap_or(Decimal::ZERO)
+    }
+
+    /// Draws a uniform sample in `[0, 1)` from the walker's own RNG, as a
+    /// `Decimal`. Used for Bernoulli-style occurrence tests where a probability
+    /// must be compared against a *uniform* variate — not a standard-normal one
+    /// (see [`Walker::bernoulli_jump`] and issue #11).
+    fn uniform_sample(&self) -> Decimal {
+        let mut rng = self.rng.lock().unwrap();
+        let u: f64 = rng.random::<f64>();
+        Decimal::from_f64(u).unwrap_or(Decimal::ZERO)
+    }
+
+    /// Bernoulli occurrence test for a jump-diffusion step: returns `true` with
+    /// probability `lambda_dt` (= λ·dt) by comparing a uniform `[0, 1)` variate
+    /// against the probability.
+    ///
+    /// This is the correct draw for a Bernoulli(λ·dt) event. It is split out of
+    /// [`jump_diffusion`](WalkTypeAble::jump_diffusion) so the empirical jump
+    /// frequency can be asserted directly and deterministically under a fixed
+    /// seed. Callers must guarantee `lambda_dt < 1` (validated once up front in
+    /// `jump_diffusion`); this helper does not re-check it.
+    fn bernoulli_jump(&self, lambda_dt: Decimal) -> bool {
+        self.uniform_sample() < lambda_dt
     }
 
     /// Ornstein-Uhlenbeck path drawn from the walker's RNG. Mirrors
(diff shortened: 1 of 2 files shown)


src/domain/walker.rs
```diff
@@ -39,6 +39,29 @@ impl Walker {
         Decimal::from_f64(z).unwrap_or(Decimal::ZERO)
     }
 
+    /// Draws a uniform sample in `[0, 1)` from the walker's own RNG, as a
+    /// `Decimal`. Used for Bernoulli-style occurrence tests where a probability
+    /// must be compared against a *uniform* variate — not a standard-normal one
+    /// (see [`Walker::bernoulli_jump`] and issue #11).
+    fn uniform_sample(&self) -> Decimal {
+        let mut rng = self.rng.lock().unwrap();
```
joaquinbejar: 🔴 **Critical**: this new `unwrap()` is reachable while serving JumpDiffusion requests, so a poisoned RNG mutex panics the worker instead of returning an error. The repository forbids production unwraps; recover the poisoned guard or propagate a typed simulation error.

joaquinbejar (commented):
The Bernoulli correction is mathematically sound, but it adds a forbidden request-path panic. Intended decision REQUEST_CHANGES, submitted as COMMENT because GitHub does not allow requesting changes on a pull request authored by the current account.

src/domain/walker.rs
joaquinbejar: Fixed in a093786, both sampling helpers recover the poisoned guard with into_inner since the RNG state stays valid after another thread's panic, no unwrap left on the walk path. Applied to the pre-existing normal_sample too.

joaquinbejar:
Addressed in a093786. Rebased onto main.

PR #36, opened 2026-07-13
Seed historical symbol and date selection (#12)

### LINKED ISSUE #12: Historical simulations ignore the session seed when selecting data

joaquinbejar:
## Summary

The seed-reproducibility contract did not cover Historical sessions: symbol
and start-date selection drew from a thread-local RNG BEFORE the walker was
built, so two sessions with identical parameters and seed could load different
tapes. Selection now derives from the session seed (domain-separated stream),
and the resolved source is persisted into the session parameters.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 14**
- **Base branch:** `stack/13-11-jumpdiffusion-bernoulli`
- **Stacked on:** #35 · **Blocks:** the next PR in the stack (#14)

## Changes

- `get_historical_data` takes a caller-provided `&mut StdRng`; the Historical
  arm of `create_random_walk` seeds it with
  `seed ^ HISTORICAL_STREAM_SALT` ("HISTORY") — domain-separated from the
  walker's stream, which is untouched. Seedless (programmatic) sessions fall
  back to entropy with a warn.
- The resolved source (chosen symbol + loaded prices + timeframe) flows back:
  `create_random_walk` returns it, `simulate_next_step` (now `&mut Session`)
  writes it into `session.parameters.method`, and the manager's existing
  `save_cas` persists it — exposed through `SessionParametersResponse.method`
  with the full price vector embedded, so replay survives eviction/restart.
- Reproducibility does not depend on the write-back: the seeded selection
  stream alone reloads the identical series.

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace`: 333 + 2 passed, 0 failed
- [x] Mock-repo: same seed twice ⇒ identical resolved symbol/prices AND identical tape; different seeds diverge (8-symbol universe, collision negligible)
- [x] Resolution persisted: post-simulate `parameters.method` carries `Some(symbol)` + non-empty prices
- [x] Non-Historical tape tests unchanged and green; clippy `-D warnings` + fmt clean

## Checklist

- [x] Walker streams untouched (salt-separated); the only remaining `rand::rng()` uses are documented entropy fallbacks for unseeded construction
- [x] Bounds-checked indexing on the selection path
- [x] No new dependencies; no `unsafe`; `tracing` only

Closes #12

### DIFF
--- a/src/domain/simulator.rs
+++ b/src/domain/simulator.rs
@@ -19,7 +19,8 @@ use optionstratlib::{
     },
 };
 use positive::{Positive, pos_or_panic};
-use rand::RngExt;
+use rand::rngs::StdRng;
+use rand::{RngExt, SeedableRng};
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use std::collections::HashMap;
@@ -36,6 +37,15 @@ const DEFAULT_SMILE_CURVE: Decimal = dec!(0.4);
 
 /// Default upper bound on the number of random walks held in the simulation cache.
 const DEFAULT_MAX_CACHED_WALKS: usize = 1000;
+
+/// Domain-separation salt (ASCII "HISTORY") mixed into the session seed to derive
+/// the RNG that selects the historical symbol and start date (issue #12).
+///
+/// XORing the seed with a fixed salt keeps the historical-selection stream
+/// deterministic per seed while decoupling it from the walker's own stream, so the
+/// same session seed always loads the same data series without perturbing — or being
+/// perturbed by — the walk's stochastic draws.
+const HISTORICAL_STREAM_SALT: u64 = 0x0048_4953_544F_5259;
 
 /// Hard bound on the number of random walks the simulation cache may hold
 /// (`OCS_MAX_CACHED_WALKS`).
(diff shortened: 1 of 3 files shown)


joaquinbejar (commented):
Seeded historical selection is correct, but the documented source persistence fails on the normal peek-then-advance flow. Intended decision REQUEST_CHANGES, submitted as COMMENT because GitHub does not allow requesting changes on a pull request authored by the current account.

src/domain/simulator.rs
```diff
@@ -99,35 +197,62 @@ impl Simulator {
...
+            let (random_walk, resolved_method) = self.create_random_walk(session).await?;
+
+            // A Historical source resolves its symbol/date range and loads prices on
+            // the first build. Persist that resolution back into the session
+            // parameters so the resolved method (symbol + embedded prices) rides along
+            // when the manager saves the session, keeping the historical tape
+            // replayable after a restart even if the database later changes.
+            // Reproducibility itself is guaranteed by the seeded selection stream and
+            // does not depend on this write-back.
+            if let Some(resolved) = resolved_method {
```
joaquinbejar: 🔴 **Critical**: this only writes the resolved history into the `Session` used for a cache-miss build. If GET peek builds the walk first, that copy is discarded while the walk remains cached, then the next POST hits the cache and saves the original unresolved method. Keep the resolution with the cache entry or otherwise apply it on cache hits so restart or eviction cannot refetch a different tape; add a peek-then-advance persistence regression.

joaquinbejar: Good catch, that peek-built copy really was dropped on the floor. The resolution now rides in the CacheEntry and gets written back to the served session on every serve including hits, with a peek-then-advance persistence regression, and since seeded selection makes rebuilds reproduce the same tape I also lifted the historical eviction pin here as promised in the eviction PR. Fixed in 6b9a106.

joaquinbejar:
Thanks for the review. The resolved-method persistence hole is fixed and the historical eviction pin is lifted, CI rerunning on the rebased branch.

PR #39, opened 2026-07-13
Hermetic default test suite with opt-in Mongo integration (#18)

joaquinbejar:
## Summary

`cargo test --workspace` on a clean checkout required a live MongoDB on
`localhost:27017`: one test hard-asserted a successful connection (failing
after ~30s of server selection without it), making the documented default
test command environment-dependent and slow. The default suite is now
hermetic and fast; live-service tests are opt-in.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 17**
- **Base branch:** `stack/16-17-mongo-timeout`
- **Stacked on:** #38 · **Blocks:** the next PR in the stack (#16)
- The diff is cumulative until the lower PRs merge — review against the base branch.

## Changes

- `test_mongodb_client_initialization` and `test_init_mongodb` are
  `#[ignore]`d with explanatory reasons (the latter tolerated absence but
  burned the full 30s server-selection timeout in hermetic runs); run them
  with `cargo test --workspace -- --ignored`.
- New hermetic test covering config construction + URI parsing without a ping.
- CI split: the "Tests" job drops the mongo service and runs the hermetic
  default suite; a new **"Integration (MongoDB)"** job keeps the `mongo:7`
  service container and runs `cargo test --workspace -- --ignored`.
- `CLAUDE.md` documents the hermetic-default / opt-in-integration split.

## Testing

- [x] Hermetic proof: with NO services running, `LOGLEVEL=WARN cargo test --workspace` → 334 + 2 passed, 0 failed, 2 ignored, in ~1s (was 1 failure + ~30s wait)
- [x] Integration proof: with mongo up, `-- --ignored` → live tests pass
- [x] clippy `-D warnings` + fmt clean

Closes #18

PR #40, opened 2026-07-13
Record session metrics only on success (#16)

### LINKED ISSUE #16: Active-session metrics are updated before operation success

joaquinbejar:
## Summary

Session metrics counted ATTEMPTED requests, not successful transitions: POST
incremented the active-sessions gauge and creation counter before the create
could fail, and DELETE decremented the gauge (and counted a deletion) before
even parsing the UUID — so invalid POSTs inflated the gauge and repeated or
invalid DELETEs drove it negative. Metrics are now recorded only on success.

## Stack

- **Stack:** #15 → #13 → #7 → #10 → #21 → #5 → #4 → #6 → #20 → #19 → #8 → #9 → #11 → #12 → #14 → #17 → #18 → #16 → #22 — **this is PR 18**
- **Base branch:** `stack/17-18-hermetic-tests`
- **Stacked on:** #39 · **Blocks:** the next PR in the stack (#22)

## Changes

- Collector API split (metric names unchanged for dashboards):
  `record_session_created()` / `record_session_deleted()` replace the
  bundled increment/decrement methods; docs state they are success-only and
  that the gauge is a process-local approximation (TTL expiry / restarts not
  reflected — documented drift; store-derived gauge noted as future work).
- `create_session` records after a successful create (PR 4 had only moved it
  past validation); `delete_session` records only on `Ok(true)` — parse
  errors, not-found, and failures record nothing, so the gauge cannot be
  driven negative through the API.

## Testing

- [x] `LOGLEVEL=WARN cargo test --workspace` (hermetic): 338 + 2 passed, 0 failed
- [x] Collector pairing: created→deleted returns the gauge to zero; counters move independently; the delete-only negative case is documented and tested as the reason call-site gating is the guard

Closes #16

### DIFF
--- a/src/infrastructure/telemetry/collector.rs
+++ b/src/infrastructure/telemetry/collector.rs
@@ -175,14 +175,31 @@ impl MetricsCollector {
             .inc();
     }
 
-    /// Increments the active sessions counter
-    pub fn increment_active_sessions(&self) {
+    /// Records that a session was created: bumps the `active_sessions` gauge and
+    /// the `session_creations_total` counter.
+    ///
+    /// Call this ONLY after the create operation has succeeded — a rejected or
+    /// failed create must not inflate either metric. The gauge is a
+    /// process-local approximation of live sessions: Redis TTL expiry and
+    /// process restarts are not reflected here (documented known drift; a
+    /// store-derived gauge is future work).
+    pub fn record_session_created(&self) {
         self.active_sessions.inc();
         self.session_creation_counter.inc();
     }
 
-    /// Decrements the active sessions counter
-    pub fn decrement_active_sessions(&self) {
+    /// Records that a session was deleted: decrements the `active_sessions`
+    /// gauge and bumps the `session_deletions_total` counter.
+    ///
+    /// Call this ONLY after a session was actually deleted (the delete returned
+    /// `Ok(true)`). Invalid ids, not-found (`Ok(false)`), and errors must record
+    /// nothing. Prometheus allows a gauge to go negative, so repeated calls
+    /// without a matching create CAN drive `active_sessions` below zero at this
+    /// level — the protection against that is the call-site gating in the delete
+    /// handler, not this method. Like the creation counterpart, the gauge is a
+    /// process-local approximation (Redis TTL expiry and restarts are not
+    /// reflected — documented known drift; a store-derived gauge is future work).
+    pub fn record_session_deleted(&self) {
         self.active_sessions.dec();
         self.session_deletion_counter.inc();
     }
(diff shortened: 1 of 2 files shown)


src/infrastructure/telemetry/collector.rs
```diff
@@ -169,14 +175,31 @@ impl MetricsCollector {
...
+    /// Call this ONLY after a session was actually deleted (the delete returned
+    /// `Ok(true)`). Invalid ids, not-found (`Ok(false)`), and errors must record
+    /// nothing. Prometheus allows a gauge to go negative, so repeated calls
+    /// without a matching create CAN drive `active_sessions` below zero at this
+    /// level — the protection against that is the call-site gating in the delete
+    /// handler, not this method. Like the creation counterpart, the gauge is a
+    /// process-local approximation (Redis TTL expiry and restarts are not
+    /// reflected — documented known drift; a store-derived gauge is future work).
+    pub fn record_session_deleted(&self) {
         self.active_sessions.dec();
```
joaquinbejar: 🟠 **Warning**: this can still make `active_sessions` negative after a process restart. The gauge restarts at zero while Redis retains sessions, so deleting one of those valid pre-existing sessions reaches this unconditional decrement. Derive the gauge from the store or only decrement sessions counted by this process.

joaquinbejar (commented):
Success-only lifecycle accounting fixes invalid requests, but the active-session gauge can still report an impossible negative value after restart. Intended decision REQUEST_CHANGES, submitted as COMMENT because GitHub does not allow requesting changes on a pull request authored by the current account.

src/infrastructure/telemetry/collector.rs
joaquinbejar: True, call-site gating could not cover deletes of sessions that predate the process. The collector now keeps an atomic count of sessions it created and only decrements the gauge when that count backs the delete, so pre-restart deletions just bump the deletion counter, with tests for the restart scenario and the zero clamp. Fixed in 039b1af.

joaquinbejar:
Thanks for the review. The gauge can no longer go negative, a store-derived gauge stays noted as future work.

PR #50, opened 2026-08-14
docs: specify the v2 rolling-simulation contract (ADR 0001)

joaquinbejar:
## Stack

- **Base branch:** `main` (this is the bottom of the stack)
- **Stack:** **#42** → #43 → #44 → #45 → #46 → #47 → #48 → #49
- **Stacked on:** — (first PR)
- **Blocks:** #51, and through it the whole v0.2.0 rolling-simulation stack

Every PR in this chain targets `main`, but each branch is cut from the previous
one, so **the diff of the PRs above this one is cumulative until the lower ones
merge**. Review each against its base branch, not against `main`. Merge strictly
bottom-up.

## Summary

Defines the public v2 contract for deterministic rolling multi-expiration
simulations **before** implementation begins, so #43–#49 have one authority to
agree with rather than seven independent interpretations.

New file: `doc/adr/0001-v2-rolling-simulation-contract.md`.

Documentation only — no runtime code, no dependency changes, no behaviour
change. `README.md` is regenerated with `make readme` (not hand-edited).

## What the ADR specifies

| § | Content |
|---|---|
| 2.1 | Module homes, so v2 respects `api → session → domain → infrastructure` and `domain` stays private |
| 3 | Effective UTC start resolved exactly once; `simulated_at = effective_start + cursor × step_interval` with checked arithmetic; the `time_frame` → seconds derivation table and its `1..=31_536_000` range check; the `expires_at <= simulated_at` cutoff |
| 4 | The versioned `weekdays_v1` calendar — daily/0DTE, selected-weekday weekly, last-weekday monthly, yearly/LEAPS — with explicit DST **fold** (take the earlier instant) and **gap** (one-minute probe, shift-agnostic so 30-minute zones work) resolution, a holiday-adjustment hook without a bundled holiday database, and the full invalid-schedule rejection list |
| 5 | Overlap: counts are evaluated **per rule**, then coincident physical expirations are deduplicated, priced **once**, and carry every matching label |
| 6 | Resources and lifecycle; v2 is immutable after creation (no PATCH/PUT); reuses v1's `expected_step` precondition (**412**) and keeps it distinct from the optimistic-concurrency CAS conflict (**409**) |
| 7 | The snapshot shape and its checkable invariants |
| 8 | The **exhaustive** replay-input list, and the distinction between invisible in-process cache eviction and a session-store TTL expiry, which is not invisible |
| 9 | Retention, cache eviction, and the `OCS_MAX_*` request caps that stop one request asking for unbounded work |
[...]

Issue #49, opened 2026-08-14
Stream deterministic JSON and CSV backtest exports


joaquinbejar:
## Overview

Add bounded-memory, deterministic JSON and CSV downloads for the complete v2 simulation tape. Clients must be able to export underlying, base-volatility, and normalized option-chain datasets for frontend analysis and backtests without manually advancing the live session.

## Context

Depends on #45, #46, #47, and #48. The current API exposes only one snapshot at a time and MongoDB event logging is best-effort, insert-only, and not a complete replay source. Export should rebuild from the immutable effective parameters/factor tape and lazy snapshot iterator rather than persist every option contract.

Any new CSV serialization dependency requires explicit approval before it is added to `[workspace.dependencies]`.

## Tasks

- [ ] Add a read-only route such as `GET /api/v2/simulations/{id}/export?dataset=underlying|volatility|option_chains&format=json|csv&from_step=&to_step=`.
- [ ] Replay from step zero (or the requested validated range) using an immutable session/version snapshot without touching the live cursor, state, version, or cache identity.
- [ ] Define stable dataset schemas and deterministic row/column ordering.
- [ ] Stream JSON as a valid array and CSV as RFC 4180 with correct quoting, line endings, `Content-Type`, and attachment filename headers.
- [ ] Flatten option-chain rows with simulated/expiration timestamps, rule labels, DTE, strike, per-strike IV, call/put bid/ask/mid/delta, gamma, and any other documented Greeks.
- [ ] Enforce range/row/resource limits, backpressure, and cancellation; isolate CPU-heavy chain pricing from Actix worker threads.
- [ ] Add typed 400/404/409/410/500 behavior for invalid datasets/formats/ranges, missing/expired sessions, version conflicts, incomplete generation, and internal failures.
- [ ] Document all query parameters, response content types/headers, schemas, statuses, and examples in utoipa without `actix_extras`.

## Acceptance criteria

- [ ] `underlying`, `volatility`, and `option_chains` are separately downloadable in JSON and CSV.
- [ ] JSON and CSV contain equal row counts and semantically identical values for each dataset and range.
- [ ] The option-chain export includes every simulated step, active expiration, strike, IV, premium, and available Greek in stable order.
- [ ] Repeated export with the same effective parameters is byte-equivalent/canonical where the chosen format permits it.
- [ ] Export never changes `current_step`, session state/version, or the snapshot returned by peek.
[...]

Issue #48, opened 2026-08-14
Make session retention and simulation-cache cleanup configurable


joaquinbejar:
## Overview

Make v2 session idle retention and domain-cache cleanup configurable and consistent so real service sessions can safely represent long simulated horizons without expiring unexpectedly or leaving orphaned heavyweight cache entries.

## Context

Depends on #44. The service currently constructs Redis sessions with a hard-coded one-hour TTL (`src/main.rs`), store defaults are shorter, in-memory cleanup uses a fixed lifetime, and the cleanup trait returns only a count rather than the IDs needed to evict simulator cache state.

The real idle lifetime is operational and must remain independent from the months/years represented by simulated time.

## Tasks

- [ ] Add typed, validated configuration for v2 session idle retention and domain-cache idle/capacity limits.
- [ ] Give in-memory and Redis v2 stores the same documented idle-retention and renewal semantics.
- [ ] Change v2 cleanup coordination to return/emit expired session IDs so corresponding factor/snapshot cache entries are evicted.
- [ ] Evict cached state on delete, completion, TTL cleanup, and capacity/idle pressure without holding locks across external awaits.
- [ ] Ensure an in-flight export owns an immutable effective-parameter snapshot and is not invalidated by store expiry.
- [ ] Add metrics for v2 cache occupancy/eviction and session expiry where useful.
- [ ] Document every new environment variable in `.env.example` and `src/lib.rs`; validate invalid values at startup with clear errors.

## Acceptance criteria

- [ ] A v2 session configured with an appropriate idle TTL remains usable while its simulated clock spans months or years.
- [ ] In-memory and Redis tests prove matching expiry/renewal behavior.
- [ ] Store cleanup, delete, and completion evict the matching factor/snapshot cache entries; no orphan remains.
- [ ] Capacity and idle eviction keep cache memory bounded and preserve deterministic rebuilds.
- [ ] Concurrent cleanup/peek/advance/export behavior has no deadlocks, lock-across-await violations, or silent state corruption.
- [ ] Export started before TTL expiry completes from its immutable parameter snapshot without mutating the session.
- [ ] Invalid/missing config paths and relevant metrics are tested; credentials/connection strings are never logged.
- [ ] Old v1 TTL and cache behavior is unchanged unless a separately documented compatibility decision says otherwise.
[...]

Issue #47, opened 2026-08-14
Expose the v2 simulation lifecycle and snapshot API


joaquinbejar:
## Overview

Expose a new `/api/v2/simulations` lifecycle and snapshot API for rolling multi-expiration simulations while preserving the existing `/api/v1/chain` contract exactly. V2 must provide deterministic simulated timestamps, explicit underlying/base-volatility state, and ordered chains grouped by absolute expiration.

## Context

Depends on #44 and #46. Changing v1 `ChainResponse` from one chain to many or changing its timestamp behavior would break IronCondor and the published crate's REST DTO contract, so the new representation belongs in separate v2 DTOs/routes.

## Tasks

- [ ] Add v2 create, get-session, delete, repeatable snapshot peek, and CAS-protected advance endpoints under `/api/v2/simulations`.
- [ ] Add v2 request/response DTOs with f64 only at the REST boundary and typed conversion in one place.
- [ ] Return `simulated_at`, underlying symbol/price/base volatility, cursor information, and ordered `chains[]` with `expires_at`, fractional DTE, labels, contracts, per-strike IV, premiums, and available Greeks.
- [ ] Preserve serve-then-advance semantics and an expected-cursor precondition for retry safety.
- [ ] Map failures through `ChainError` and the single HTTP error boundary.
- [ ] Add accurate utoipa annotations, manual query/path params, schemas, examples, headers, and status codes without enabling `actix_extras`.
- [ ] Add golden v1 contract tests and valid v2 examples using the reference expiration schedule.

## Acceptance criteria

- [ ] Repeated v2 peek calls are semantically/byte stable for simulation data and do not advance or persist the cursor.
- [ ] Advance serves the current snapshot then advances exactly once; CAS races yield one winner and one documented conflict.
- [ ] V2 can create and walk the full reference configuration across multiple rolls.
- [ ] Invalid IDs, configurations, cursors, completed simulations, and internal failures return documented typed statuses/bodies.
- [ ] `/api/v1/chain` routes, DTO fields, timestamp behavior, statuses, and OpenAPI examples remain unchanged.
- [ ] The effective seed, effective start, timezone, calendar version, and normalized schedules are surfaced so clients can replay the tape.
- [ ] OpenAPI contains every v2 path/schema and no sensitive/environment-specific values.
- [ ] DTO serialization/deserialization tests cover optional omission, unknown-field rejection, and all error paths.
[...]

PR #55, opened 2026-08-14
feat(session): add deterministic v2 simulation parameters and persistence

### LINKED ISSUE #44: Add deterministic v2 session parameters and persistence
## Overview

Add separate v2 session/request types that persist the effective simulated clock and normalized rolling-expiration configuration without changing the published v1 DTO or stored-session shape. A v2 session must be independently replayable from its stored effective parameters.

## Context

Depends on #42. `SimulationParameters` and `CreateSessionRequest` are public and IronCondor-facing, while Redis stores v1 sessions as serde JSON. Adding rolling fields directly to those structs would be source-breaking even if serde defaults preserved old JSON, so v2 needs a separate versioned contract and Redis key/schema.

## Tasks

- [ ] Add v2 request/session parameter types for effective start, fixed step interval, timezone, calendar ID/version, expiration time, schedules, market path, chain shape, skew/smile, and seed.
- [ ] Generate `start_at` exactly once when omitted, normalize it to UTC, persist it, and return it alongside the effective seed.
- [ ] Make internal v2 seed and effective start non-optional after request conversion.
- [ ] Validate schedule counts, rule IDs, weekdays, local times, timezones, step/range limits, symbols, and numeric fields at the DTO conversion boundary.
- [ ] Derive `simulated_at = effective_start + cursor × fixed_interval` with checked arithmetic.
- [ ] Add a versioned v2 store key/schema to both in-memory and Redis stores; do not reinterpret old v1 JSON as rolling configuration.
- [ ] Keep v2 configuration immutable after creation; a changed configuration creates a new simulation rather than PATCH/PUT mutation.
- [ ] Preserve the effective-seed guarantee and optimistic concurrency semantics.

## Acceptance criteria

- [ ] The example configuration (1 rolling 0DTE, 3 Monday/Wednesday/Friday weeklies, 12 last-Friday monthlies, 17:00 local) serializes, deserializes, and round-trips through both stores.
- [ ] Omitting `start_at` generates one effective value that is persisted, surfaced, and unchanged after reads/restarts.
- [ ] The same v2 parameters derive the same `simulated_at` for every cursor without consulting wall time.
- [ ] Unknown request fields and every invalid field return a field-specific validation error; no request path panics.
- [ ] New v2 stored JSON has explicit schema/version coverage and old v1 JSON still deserializes with unchanged legacy behavior.
- [ ] Golden v1 DTO and stored-shape tests prove no v1 field, type, serde behavior, or status contract changed.
- [ ] Session/store tests cover create/get/delete, CAS conflicts, duplicate IDs, and error paths for v2.
- [ ] Utoipa/schema docs and `src/lib.rs` are updated; `README.md` is regenerated with `make readme`.

joaquinbejar:
## Stack

- **Base branch:** `stack/2-43-rolling-expiry-planner`
- **Stack:** #42 → #43 → **#44** → #45 → #46 → #47 → #48 → #49
- **Stacked on:** #51
- **Blocks:** #60, and through it #46 (consumes the schedule), #47 (surfaces the DTOs), #48 (retention), #49 (immutable parameter snapshot)

Targets `main` but is cut from `stack/2-43-rolling-expiry-planner`, so **the diff against `main` is cumulative** — it still contains #50's ADR and #51's planner. Review against the base branch. Merge bottom-up.

## Summary

Gives v2 rolling simulations their own request DTO, resolved parameters, session document and store key space — so the rolling contract can exist without touching anything `/api/v1/chain` publishes.

**Why separate types rather than new fields.** `SimulationParameters` and `Session` are public, IronCondor-facing and persisted as serde JSON. Adding rolling fields would be source-breaking even where serde defaults kept old documents loading, and it would tie the v1 stored shape to v2's evolution.

## What lands

| | |
|---|---|
| `CreateSimulationRequest` | `src/api/rest/requests_v2.rs`. f64 and strings at the boundary, **`deny_unknown_fields`** — a typo is a `400` naming the field, not a silently ignored parameter that changes the tape. |
| `SimulationParametersV2` | `src/session/model_v2.rs`. Its fields are exactly the replay-input list of ADR §8. |
| Resolution | Three values are *resolved*, not merely converted, and all three are surfaced so a client can replay from the creation response alone: the **seed** (generated when absent, as in v1), the **effective start** (generated when absent, normalised to whole-second UTC — the only wall-clock read in a v2 simulation's life), and the **step interval** (derived from `time_frame` when absent). |
| The simulated clock | `simulated_at(cursor) = effective_start + cursor × step_interval`, checked throughout. An overflow is a rejected request, never a wrapped timestamp. |
| `SessionV2` | Reuses the v1 `SessionState` machine but reaches only `Initialized → InProgress → Completed`, since a v2 simulation is immutable after creation. Carries an explicit `schema_version`. |
| `SimulationStore` | A separate trait, with in-memory and Redis implementations. The in-memory store is a **separate map**, the Redis store a **separate key prefix** (`optionchain:simulation:v2:`) — the v1/v2 isolation is structural, not conventional: a v2 id cannot resolve a v1 session because the two never share a container. Redis reuses v1's already-reviewed CAS Lua script, which is parameterised entirely by its keys. |
| `cleanup() -> Vec<Uuid>` | Returns the **expired ids**, not a count, because #48's caller owns heavyweight per-simulation domain caches and cannot evict them without knowing which sessions went away. On Redis that needs an index set, since key expiry is otherwise invisible to the process. v1's `cleanup() -> usize` is untouched. |

## Also: the expiry rule wire format

This PR is the first code to depend on it. The kind-specific fields are now declared explicitly instead of flattening an internally-tagged enum — flattening produced the same JSON but forfeits `deny_unknown_fields`, and with it both guarantees ADR §4.4 makes: that an unknown field is rejected, and that a field not valid for the rule's `kind` (`weekdays` on a `daily` rule) is rejected rather than silently ignored. **The stored shape is unchanged** and its golden test still passes.

The expiry *configuration* types become `pub` and are re-exported through `session`, since they are fields of the public `SimulationParametersV2`; the *planner* stays crate-internal. The ADR records the trade-off this accepts: `chrono-tz` enters the public API, so a major bump there is a semver event here. The alternative — a parallel primitive-typed representation in `session` — would mean two shapes and two validation paths that could drift.

## v1 is frozen, and now provably so

`tests/v1_contract.rs`, 12 new tests. They assert against **literal JSON** rather than round-trips, because a round-trip stays green when both sides move together — exactly the regression this work could introduce. They pin:

- the v1 request and stored shapes, **including a pre-`seed` document that must still load** (the `#[serde(default)]` guarantee §12.2 rests on);
- the **exact key set** of the stored parameters — a new key would be written by a new binary and read by an old one;
- the response DTO shapes, including `ChainResponse.timestamp`, whose wall-clock semantics v2 deliberately does not back-port;
- the tri-state PATCH semantics (absent / null / value are three distinct intents);
- the **rendered** value of `SessionState` — `"In Progress"` *with a space*. v2 uses `snake_case`; "normalising" v1 to match would break IronCondor.
- that v1 still **tolerates** unknown fields, which v2 rejects — making the difference deliberate rather than accidental.

## Verification

- `cargo fmt --all --check`
- `LOGLEVEL=WARN cargo test --workspace` — **474 passed, 0 failed, 13 ignored**, up from 418; plus **16** golden contract tests and 2 doctests
- CI's Integration job ran all **13 ignored tests against live services** — 4 MongoDB and the 9 new live-Redis ones
- `cargo build --release`
- `make readme` — no README change (no `src/lib.rs` doc change in this PR)

No new dependency, no new env var.

## Review round

`architect` audited the first commit and found two 🔴. Both are inert today — no store is wired into `main.rs` — and both go live the moment #46 builds on this, so they are fixed in the second commit.

**Stored documents were trusted completely.** `SimulationParametersV2` and `SessionV2` derived `Deserialize` directly with public fields, so every invariant the request path establishes was reachable *around* it. A probe against the public API loaded a hand-edited document with:

- `step_interval_seconds: 0` — which **freezes the simulated clock**: every snapshot carries the same instant, the expiration cutoff never advances, the rolling inventory never rolls. Silently degraded, not an error.
- `steps` past the cap, driving an unbounded factor tape in #46;
- a sub-second `effective_start`, breaking the whole-second rendering §10.2 needs for byte-comparable exports;
- a `symbol` carrying `,` `"` `|` — the exact CSV-corruption hazard the `rule_id` charset was narrowed to keep out, reachable through the field that was *not* re-validated;
- `state: "Reinitialized"`, which the code itself documents as unreachable for v2.

Both types now route `Deserialize` through a validating wire type, mirroring the expiry types. Redis is an outer layer, and `rules/global_rules.md` is explicit that a domain type must not trust one.

**Nothing read `schema_version`.** During a rolling deploy an old replica would read a newer document, drop what it does not understand, and write the truncated version back — with an intact revision, so the CAS *succeeds* and the data is simply gone. The stored shapes now carry `deny_unknown_fields`, and a future `schema_version` is refused on load.

Three Redis races, all 🟠:

- **`create` wrote three keys non-atomically**, and both torn states were silent: a document with no revision key, and a document absent from the index — which `cleanup` could then never report, so #48's caller would never evict its caches for it. Each operation is now one Lua script.
- **`cleanup` could report a live simulation.** It walked `SMEMBERS`, probed `EXISTS`, then deleted and removed in separate round trips. `UuidGenerator` is a deterministic counter that repeats across process restarts, so a `create` landing in that window made cleanup report an id that had just been recreated — evicting a live cache and dropping the id from the index *permanently*, so its real expiry was never reported.
- **The index was an unbounded `SET` with no TTL**, walked with `SMEMBERS` and N+1 `EXISTS` probes.

All three are gone: the index is a **sorted set scored by each simulation's expiry deadline**, read from Redis's own `TIME` rather than the client's clock, and cleanup is a bounded range query that reaps and reports in one unit.

Also folded in: the two retention constants become one (§9.1 requires the backends to agree; two names are how that drifts), `MIN_/MAX_STEP_INTERVAL_SECONDS` become `pub(crate)` so #48's env-knob conversion is not a breaking change, `requests_v2` imports `ExpiryRule` from `session` rather than reaching into `domain`, `from_parts` normalises a weekly set so a Rust caller cannot persist an un-normalised replay input, the expiry module doc stops describing the visibility policy this stack abandoned, and the dead-code expectation moves to per-item so #46 must remove each one.

### The Redis store now has real coverage

9 `#[ignore]`d live tests running against the `redis:7.4` service CI already provisions — verified in this run's Integration job, all 9 green. They cover what a mock cannot: that `create` writes all three keys together, that the CAS commits once and rejects the loser, that `delete` leaves no debris, that `cleanup` reports expired ids and spares live ones, that it is idempotent, and that a tampered document is refused on load.

### One gap handed to #47, explicitly

A schedule rule validates *during* deserialization, so its `ChainError::Validation` is flattened into a serde error and actix would render a plaintext `400` with no `field`. The v2 routes must register a `web::JsonConfig` error handler to recover it. That is now written into ADR §11 as part of #47's acceptance rather than left as an undocumented wart.

Closes #44

### DIFF
--- a/src/domain/expiry.rs
+++ b/src/domain/expiry.rs
@@ -52,18 +52,23 @@
 //! exactly like one built from a request. Without that, a stored
 //! `target_count` of `1e11` would reach the projection loop.
 
-// The planner is the bottom of the v0.2.0 stack and has no in-tree caller yet:
-// issue #44 persists an `ExpirationSchedule` in the v2 session parameters and
-// #46 calls `RollingPlanner::active_at` to build each snapshot. Landing it with
-// its own tests first is what keeps every PR in the stack independently sound.
+// The planner *evaluation* half — `RollingPlanner`, `ActiveExpiry`, and their
+// helpers — has no in-tree caller yet: #46 calls `active_at` to build each
+// snapshot. The *configuration* half is already live, wired into the v2 session
+// parameters by #44.
 //
-// `expect` rather than `allow` on purpose: the moment #44 or #46 wires a caller
-// in, the expectation goes unfulfilled and CI's `-D warnings` turns it into a
-// hard error, so removing this is enforced by the compiler rather than promised
-// in a commit message.
-#![expect(
-    dead_code,
-    reason = "no in-tree caller until #44 persists the schedule and #46 calls active_at"
+// `expect` rather than `allow` on purpose: once #46 exercises the last of these
+// items the expectation goes unfulfilled, and CI's `-D warnings` turns that into
+// a hard error. Removing this line is therefore enforced by the compiler rather
+// than promised in a commit message. It is scoped to non-test builds because the
+// module's own tests already exercise the evaluation path, which would make the
+// expectation unfulfilled under `cargo test` for the wrong reason.
+#![cfg_attr(
+    not(test),
+    expect(
+        dead_code,
+        reason = "the planner evaluation path has no in-tree caller until #46 builds snapshots"
+    )
 )]
 
 use crate::utils::ChainError;
@@ -97,6 +102,10 @@ pub(crate) const MAX_TARGET_COUNT: usize = 256;
 /// A plain constant for now; issue #48 turns it into an `OCS_MAX_*` knob.
 pub(crate) const MAX_EXPIRATIONS_PER_SNAPSHOT: usize = 512;
 
+/// The month a `yearly` rule expires in when the request omits it — December,
+/// the conventional LEAPS expiry month.
+pub(crate) const DEFAULT_YEARLY_MONTH: u32 = 12;
+
 /// Maximum length of a `rule_id`.
 ///
 /// Rule ids are echoed as labels on every chain of every step and are joined
@@ -114,10 +123,6 @@ const DAY_SCAN_SLACK: usize = 32;
 
 /// Upper bound on how many candidate periods a monthly or yearly rule may scan.
 const PERIOD_SCAN_SLACK: usize = 2;
-
-/// The month a `yearly` rule expires in when the request omits it — December,
-/// as ADR 0001 §4.1 specifies.
-const DEFAULT_YEARLY_MONTH: u32 = 12;
 
 /// Seconds in a day, as a `Decimal`, for the fractional days-to-expiration
 /// conversion.
@@ -143,7 +148,7 @@ pub(crate) fn tzdb_version() -> &'static str {
 /// set, so a future policy can be added without silently changing the tape of
 /// a simulation created under an older one.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
-pub(crate) enum CalendarVersion {
+pub enum CalendarVersion {
     /// Weekends are the only ineligible days; no exchange-holiday data is
     /// consulted. See [`CalendarVersion::eligible_date`].
     #[serde(rename = "weekdays_v1")]
@@ -165,7 +170,7 @@ impl CalendarVersion {
     /// candidates onto the same date, which is why per-rule projection counts
     /// **distinct instants** (see [`RollingPlanner::project_rule`]).
     #[must_use]
-    pub(crate) fn eligible_date(self, date: NaiveDate) -> Option<NaiveDate> {
+    pub fn eligible_date(self, date: NaiveDate) -> Option<NaiveDate> {
         match self {
             CalendarVersion::WeekdaysV1 => match date.weekday() {
                 Weekday::Sat | Weekday::Sun => None,
@@ -176,7 +181,7 @@ impl CalendarVersion {
 
     /// The stable wire name of this calendar version.
     #[must_use]
-    pub(crate) fn as_str(self) -> &'static str {
+    pub fn as_str(self) -> &'static str {
         match self {
             CalendarVersion::WeekdaysV1 => "weekdays_v1",
         }
@@ -185,13 +190,12 @@ impl CalendarVersion {
 
 /// What a rule expires on, independent of how many expirations it keeps.
 ///
-/// Serialised **internally tagged** under `kind` with `snake_case` variant
-/// names, and flattened into [`ExpiryRule`], so a stored rule reads exactly as
-/// ADR 0001 §14.2 shows it:
+/// The wire form is flat and tagged by `kind` — see [`ExpiryRuleWire`], which
+/// carries the serde derives so that `deny_unknown_fields` and per-kind field
+/// validity both hold. A stored rule reads exactly as ADR 0001 §14.2 shows it:
 /// `{"rule_id": "monthlies", "kind": "monthly", "target_count": 12, "weekday": "Fri"}`.
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(tag = "kind", rename_all = "snake_case")]
-pub(crate) enum ExpiryRuleKind {
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ExpiryRuleKind {
     /// Every eligible weekday expires — the rolling 0DTE.
     Daily,
     /// Each named weekday expires. The set is non-empty and contains no
@@ -228,7 +232,7 @@ impl ExpiryRuleKind {
     /// [`ExpirationSchedule::validate`], which knows the owning rule id and can
     /// therefore name the offending field.
     #[must_use]
-    pub(crate) fn weekly(weekdays: impl IntoIterator<Item = Weekday>) -> Self {
+    pub fn weekly(weekdays: impl IntoIterator<Item = Weekday>) -> Self {
         // `chrono::Weekday` is not `Ord`, so deduplicate and order through the
         // Monday-based index, which round-trips losslessly.
         let ordered: BTreeSet<u8> = weekdays
@@ -253,26 +257,6 @@ impl ExpiryRuleKind {
         }
     }
 
-    /// The wire tag this kind serialises under.
-    ///
-    /// Nothing calls this at runtime — it exists so that adding a variant to
-    /// [`ExpiryRuleKind`] fails to compile until [`ExpiryRuleKindTag`] gains
-    /// the matching variant. Without it the two lists drift, and a kind that
-    /// serialises fine becomes a stored schedule that cannot be read back.
-    #[must_use]
-    #[cfg_attr(
-        not(test),
-        expect(dead_code, reason = "compile-time exhaustiveness link")
-    )]
-    fn tag(&self) -> ExpiryRuleKindTag {
-        match self {
-            ExpiryRuleKind::Daily => ExpiryRuleKindTag::Daily,
-            ExpiryRuleKind::Weekly { .. } => ExpiryRuleKindTag::Weekly,
-            ExpiryRuleKind::Monthly { .. } => ExpiryRuleKindTag::Monthly,
-            ExpiryRuleKind::Yearly { .. } => ExpiryRuleKindTag::Yearly,
-        }
-    }
-
     /// A short, stable description of the rule kind, used in validation
     /// messages.
     #[must_use]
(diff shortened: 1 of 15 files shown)


joaquinbejar (commented):
The v2 boundary and Redis CAS implementation are strong, but stored numeric invariants, lifecycle consistency, and validation before writes still have bypasses that can persist or load invalid simulations. REQUEST_CHANGES.

src/session/model_v2.rs
```diff
@@ -0,0 +1,1435 @@
...
+        if self.effective_start.nanosecond() != 0 {
+            return Err(ChainError::Validation {
+                field: "effective_start".to_string(),
+                reason: format!(
+                    "must be on a whole second, got {}",
+                    self.effective_start.to_rfc3339()
+                ),
+            });
+        }
+        self.schedule.validate()?;
```
joaquinbejar, line 386: 🟠 **Warning**: Stored-document validation does not recheck the strict-positive constraints for `initial_price`, `volatility`, or `strike_interval`, and it does not revalidate the `WalkType` invariants enforced by `ApiWalkType::try_into`. A tampered Redis document with a zero price or `dt` therefore loads successfully. Reuse the same typed validators here and add those cases to the tamper test.

src/session/model_v2.rs
```diff
@@ -0,0 +1,1435 @@
...
+        if self.current_step > self.total_steps {
+            return Err(ChainError::Validation {
+                field: "current_step".to_string(),
+                reason: format!(
+                    "must not exceed total_steps ({}), got {}",
+                    self.total_steps, self.current_step
+                ),
+            });
+        }
+        match self.state {
```
joaquinbejar, line 707: 🟠 **Warning**: This accepts `Error` and cursor/state combinations that the documented v2 lifecycle cannot produce, such as `Completed` at step 0 or `Initialized` at step 5. Validate `state` against `current_step` and `total_steps`, and reject `Error`, `Modified`, and `Reinitialized`, so corrupted documents cannot enter the manager.

src/session/store/v2_interface.rs
```diff
@@ -0,0 +1,80 @@
...
+    ///
+    /// Never overwrites: a colliding id is rejected rather than silently
+    /// clobbering a live simulation, which is what makes a fresh manager safe
+    /// after a restart or on a second replica.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ChainError::AlreadyExists`] on an id collision, or another
+    /// [`ChainError`] on a storage or serialization failure.
+    async fn create(&self, simulation: SessionV2) -> Result<(), ChainError>;
```
joaquinbejar, line 45: 🟠 **Warning**: Both store implementations accept this public, field-mutable `SessionV2` without calling `validate` on `create` or `save_cas`. A crate caller can persist invalid parameters or state; the in-memory backend keeps serving it while Redis only fails on the next load. Validate before every write or make model construction invariant-preserving.

src/session/model_v2.rs
joaquinbejar: Added both. The three Positive fields are rechecked strictly positive on load, and rather than mirror the walk invariants a second time in the session layer the stored WalkType now round-trips through the ApiWalkType mirror, so it faces exactly the checks a request does. The tamper table grows four cases covering a zero initial_price, volatility, strike_interval and dt.

src/session/model_v2.rs
joaquinbejar: Error now joins Modified and Reinitialized in the rejected set, and state is validated against the cursor: Initialized only at 0, Completed only at total_steps, InProgress only strictly between them. A new test walks the four contradictions, and the two store tests that advanced the cursor now move the state with it, which is what the manager will do.

src/session/store/v2_interface.rs
joaquinbejar: Both backends now call validate on create and on save_cas, and the trait spells that out as a requirement on any future implementation rather than leaving it to each one. The in-memory store has tests for both write paths, including that nothing is stored when the document is refused.

src/session/model_v2.rs
```diff
@@ -0,0 +1,1016 @@
...
+        let parameters = parameters(request);
+
+        assert_eq!(parameters.effective_start.nanosecond(), 0);
+        // Reading it again yields the same value: it is stored, not recomputed.
+        assert_eq!(parameters.effective_start, parameters.effective_start);
+        match parameters.simulated_at(0) {
+            Ok(at) => assert_eq!(at, parameters.effective_start),
+            Err(error) => panic!("cursor 0 must resolve: {error}"),
+        }
+    }
```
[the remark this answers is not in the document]
joaquinbejar: Removed, the simulated_at(0) check below it already carries the meaning.

src/session/store/v2_redis.rs
```diff
@@ -0,0 +1,368 @@
...
+            .map_err(Self::map_redis_error)?;
+
+        let mut conn = self.client.connection_manager();
+        let _: () = redis::cmd("SADD")
+            .arg(self.index_key())
+            .arg(id.to_string())
+            .query_async(&mut conn)
+            .await
+            .map_err(Self::map_redis_error)?;
+
```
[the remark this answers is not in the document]
joaquinbejar: Already atomic as of the last commit on the branch. Create runs as a single Lua script over the document, the version key and the index set, so the three either all land or none do and there is no window where a live document is missing from the cleanup index.

src/domain/expiry.rs
```diff
@@ -0,0 +1,2229 @@
...
+//! # Visibility
+//!
+//! Every type here is `pub(crate)` and stays inside the private `domain`
+//! module. It is deliberately **not** part of the published crate's API: doing
+//! so would put `chrono_tz::Tz`, `chrono::Weekday` and `chrono::NaiveTime` into
+//! the public surface, and a `chrono-tz` major bump would then be a breaking
+//! change for downstream consumers such as IronCondor. Issue #44 owns the
+//! public representation — primitive-typed schedule fields on the v2 session
+//! parameters — and converts into these types at the one boundary that already
+//! does f64 → typed conversion.
```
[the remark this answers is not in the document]
joaquinbejar: The module docs were rewritten on this branch and now say the opposite of what you read: the configuration types are pub and re-exported through session, only the planner and ActiveExpiry stay crate-private, and there is a paragraph on why publishing them puts chrono-tz on the public surface.

src/session/model_v2.rs
```diff
@@ -0,0 +1,1016 @@
...
+    /// The resolved interval between simulated steps, in seconds.
+    pub step_interval_seconds: u64,
+    /// Time frame the stochastic model is scaled by.
+    pub time_frame: TimeFrame,
+    /// The normalised rolling expiration schedule.
+    pub schedule: ExpirationSchedule,
+    /// The IANA time-zone database release the expirations were resolved
+    /// against, e.g. `2025b`. A replay against a different release is still a
+    /// replay — it is just one the client can detect (ADR 0001 §8).
+    pub tzdb_version: String,
```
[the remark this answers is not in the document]
joaquinbejar: Deliberate, and it is your option b. This type is the stored shape, and the ADR only pins the wire shape of the creation response in §14.2, which the REST DTOs in #47 own and will match field for field. Keeping the schedule nested here lets the schedule validate as one object with its calendar, timezone and expiration time attached, and a client replaying from a recorded response never sees this type. Worth a line in the ADR when #47 lands so the two shapes are documented as intentionally different.

joaquinbejar:
All three bypasses are closed in c3bc381, and the branch is rebased on main now that the planner merged. Stored parameters recheck the strict-positive constraints and round-trip the walk through the ApiWalkType mirror instead of duplicating its invariants here, state is validated against the cursor with Error joining the rejected set, and both stores validate on create and on save_cas with the trait documenting that for future backends. The Redis atomicity point was already handled by the Lua create script, and the expiry module docs already say the configuration types are public. The nested schedule shape is intentional and I answered it inline: this is the stored shape, the ADR pins the response shape, and #47 owns matching it. Gate is green with 482 unit tests plus the 16 v1 contract tests.

Issue #46, opened 2026-08-14
Build lazy rolling multi-expiration snapshots


joaquinbejar:
## Overview

Build lazy rolling multi-expiration option snapshots from one factor row and the active expiries returned by the planner. Each snapshot must show how premiums and per-strike volatility change with simulated price, base volatility, skew/smile, and exact time to expiration while keeping memory bounded.

## Context

Depends on #43 and #45. The current simulator builds one chain with one relative DTE and materializes all chain steps up front. Optionstratlib must remain the source of option pricing, Greeks, spreads, skew, and smile through `OptionChain::build_chain`.

Do not use `generator_optionseries` as the rolling engine: it ages and drops expiries but does not replenish them, and the current `OptionSeries::build_series` path does not preserve the requested strike interval. Use a thin local aggregate around upstream `OptionChain` instead.

## Tasks

- [ ] Add a domain series snapshot with `step`, `simulated_at`, factor state, and ordered chains keyed by absolute `expires_at` plus schedule labels.
- [ ] Compute positive fractional DTE from the same `simulated_at`/`expires_at` pair used by the planner.
- [ ] Build each active chain with upstream `OptionChain::build_chain` and the configured size, interval, spread, risk-free rate, dividend yield, base IV, skew, and smile.
- [ ] Replenish expired schedule slots in the same snapshot; never emit an expired chain.
- [ ] Deduplicate coincident physical expirations while preserving all labels.
- [ ] Build only requested steps and add a bounded snapshot cache with deterministic eviction/rebuild.
- [ ] Strengthen reproducibility comparisons to include timestamps, expiries, labels, spot, base/per-strike IV, strikes, quotes, and Greeks.

## Acceptance criteria

- [ ] The reference configuration returns 1 0DTE, 3 weekly, and 12 monthly rule slots at every step; coincident dates follow the documented deduplication semantics.
- [ ] Every chain shares the step spot/base volatility and has its own exact expiration and fractional DTE.
- [ ] Crossing a 17:00 cutoff removes and replaces the expired chain without a missing-inventory step.
- [ ] Multi-month/year simulations complete all requested steps despite repeated 0DTE/weekly/monthly rolls.
- [ ] Increasing base IV raises representative ATM premiums; decreasing DTE reduces representative ATM extrinsic value; spot changes reprice all strikes.
- [ ] Negative skew slope and positive smile curvature produce the expected per-strike IV shapes using upstream calculations.
[...]

Issue #43, opened 2026-08-14
Implement a pure rolling-expiry planner


joaquinbejar:
## Overview

Implement a pure, deterministic planner that converts a versioned expiration schedule and a simulated timestamp into the active rolling expiry inventory. This planner is the source of truth for keeping 0DTE, weekly, monthly, and yearly/LEAPS contracts continuously available across long simulations.

## Context

Depends on #42. The current domain has one `days_to_expiration` and upstream `Xstep::next` stops when it reaches zero; neither can replenish an expired chain. The planner belongs in the domain layer but must not perform option pricing, I/O, RNG draws, or wall-clock reads.

## Tasks

- [ ] Add typed schedule rules for eligible daily/0DTE expiries, selected-weekday weeklies, last-weekday monthlies, and yearly/LEAPS expiries.
- [ ] Support stable rule IDs/labels, `target_count`, local expiration time, IANA timezone, and a versioned calendar policy.
- [ ] Implement deterministic `weekdays_v1` eligibility and holiday-adjustment hooks without bundling a full exchange-holiday database.
- [ ] Define and implement cutoff behavior: `expires_at <= simulated_at` is expired and its replacement is present in the same planner result.
- [ ] Evaluate counts per rule, then deduplicate coincident physical expirations while retaining every matching label.
- [ ] Return stable chronological ordering and typed validation/arithmetic errors through `ChainError`.
- [ ] Document any proposed timezone dependency and obtain explicit approval before adding it to workspace dependencies.

## Acceptance criteria

- [ ] At 16:59 local, an eligible same-day 17:00 0DTE exists; at exactly 17:00 it is gone and the next eligible daily expiry is present.
- [ ] A weekly rule with Monday/Wednesday/Friday and `target_count = 3` always returns the next three rule expirations.
- [ ] A monthly last-Friday rule with `target_count = 12` remains at twelve across month, year, and leap-year boundaries.
- [ ] A yearly rule can maintain expirations at or beyond one year for LEAPS-style scenarios.
- [ ] Overlapping weekly/monthly rules produce one expiry with multiple labels, while each rule still satisfies its pre-deduplication count.
- [ ] DST gaps/folds, weekends, invalid local times, zero counts, duplicate rule IDs, impossible schedules, and checked-arithmetic overflow have tests.
- [ ] Results are independent of machine date, locale, RNG seed, call order, and process restart.
- [ ] Unit tests cover every rule's happy and error paths.
[...]

Issue #42, opened 2026-08-14
Specify the v2 rolling-simulation contract


joaquinbejar:
## Overview

Define the public v2 contract for deterministic rolling multi-expiration simulations before implementation begins. The contract must support long simulated horizons, exact expiration timestamps, configurable 0DTE/weekly/monthly/LEAPS inventories, complete-tape replay, and JSON/CSV backtest exports without changing the IronCondor-facing v1 API.

## Context

The current service accepts one relative `days_to_expiration` (`src/api/rest/requests.rs`, `src/session/model.rs`), builds one `OptionChain` per step (`src/domain/simulator.rs`), and stamps responses with `Utc::now()` (`src/api/rest/handlers.rs`). Relative expirations and wall-clock response timestamps are insufficient for replayable schedules such as one 0DTE, three Monday/Wednesday/Friday weeklies, and twelve last-Friday monthlies expiring at 17:00.

This issue is the architecture/contract prerequisite for the v0.2.0 implementation stack. It must not publish the crate or mutate `/api/v1/chain`.

## Tasks

- [ ] Write an ADR in the crate documentation defining `/api/v2/simulations` resources, lifecycle operations, snapshot DTOs, export datasets, and typed error semantics.
- [ ] Define deterministic time semantics: optional request `start_at`, persisted effective UTC start, IANA timezone, local expiry cutoff, fixed simulation step interval, and `expires_at <= simulated_at` as expired.
- [ ] Define versioned `weekdays_v1` calendar behavior, DST handling, daily/0DTE, selected-weekday weekly, last-weekday monthly, and yearly/LEAPS rules.
- [ ] Define overlap semantics: counts are evaluated per rule, while coincident expirations are priced once and carry every matching schedule label.
- [ ] Define replay semantics: the effective seed, effective start, calendar version, timezone, and normalized schedules are required to reproduce the complete tape.
- [ ] Define stable JSON and normalized CSV schemas for `underlying`, `volatility`, and `option_chains` datasets, including ordering and range semantics.
- [ ] Record compatibility and release policy: v1 stays byte/behavior compatible, v2 targets crate milestone v0.2.0, and publication remains separately approved.
- [ ] Record that adding an IANA-timezone or CSV crate requires explicit dependency approval before implementation.

## Acceptance criteria

[...]

PR #60, opened 2026-08-14
feat(domain): add the reproducible market-factor tape

### LINKED ISSUE #45: Generate a reproducible underlying and volatility factor tape
## Overview

Create a lightweight, first-class market-factor tape containing simulated time, underlying price, and base volatility for every requested step. The factor tape must continue for the complete simulation horizon independently of any option expiration and become the deterministic source for both chain pricing and exports.

## Context

Depends on #42. Today the cache stores `RandomWalk<Positive, OptionChain>`, so the whole chain tape is prebuilt and the walk truncates when its single expiry reaches zero (`src/domain/simulator.rs`). Optionstratlib already exposes aligned seeded price/volatility paths via `WalkTypeAble::generate_with_vol`; that upstream behavior should be reused.

There is also an ambiguity to remove: top-level `SimulationParameters.volatility` prices step zero, while the selected `WalkType` may carry another volatility for later steps.

## Tasks

- [ ] Introduce a domain factor row containing `step`, `simulated_at`, underlying price, and canonical base implied volatility.
- [ ] Generate exactly the requested number of rows using the seeded `Walker` and upstream `generate_with_vol` semantics.
- [ ] Define one authoritative base-volatility source and validate/eliminate disagreement between initial and walk-model volatility.
- [ ] Preserve constant volatility for constant-vol models and aligned stochastic volatility for GARCH, Heston, Custom, Telegraph, and historical paths.
- [ ] Keep historical volatility free of look-ahead and historical source selection seed-reproducible.
- [ ] Ensure expiration schedules and lazy chain requests consume no factor-tape RNG draws and cannot perturb the spot/volatility stream.
- [ ] Store/cache only the lightweight factor tape, not contracts for every step.

## Acceptance criteria

- [ ] Every row exposes the exact base volatility used to price that step's chains.
- [ ] A simulation spanning months/years produces exactly `steps` factor rows even though many short-dated options would expire during it.
- [ ] Same effective parameters/start/seed produce identical complete factor tapes; different seeds produce different tapes.
- [ ] GARCH/Heston tests prove volatility varies and stays index-aligned with spot; constant-vol tests prove it remains constant.
- [ ] Historical tests prove no look-ahead and deterministic rebuild after cache eviction/restart.
- [ ] Adding/removing/reordering expiration rules leaves the factor tape unchanged.
- [ ] No `Utc::now()`, thread-local RNG, or expiration-relative formatting affects deterministic factor output.
- [ ] Memory scales as `O(steps)` and configured step limits remain enforced.
- [ ] Same-seed and different-seed tape tests cover full rows, not only spot prices.
- [ ] Domain docs are updated and `README.md` is regenerated via `make readme` when the public surface changes.

joaquinbejar:
## Stack

- **Base branch:** `stack/3-44-v2-session-params`
- **Stack:** #42 → #43 → #44 → **#45** → #46 → #47 → #48 → #49
- **Stacked on:** #55
- **Blocks:** #61, and through it #46 (builds snapshots from a row) and #49 (exports the tape)

Targets `main` but is cut from `stack/3-44-v2-session-params`, so **the diff against `main` is cumulative** — it still contains #50, #51 and #55. Review against the base branch. Merge bottom-up.

## Summary

Adds the market path a v2 rolling simulation walks: one lightweight row per step carrying the simulated instant, the underlying price, and the base implied volatility that will price every chain at that step.

**Why it is separate from the chains.** v1 caches a `RandomWalk<Positive, OptionChain>`, so the whole chain tape is materialised up front and the walk truncates when its one expiry reaches zero. Neither survives a rolling simulation. Splitting the two lets the market path run for the **entire** requested horizon regardless of what expires along the way, and lets chains be built on demand from a row — memory is `O(steps)` in four-field rows rather than `O(steps × expiries × strikes)` in contracts.

## It reuses the kernels rather than reimplementing them

The tape asks the same seeded `Walker` for the same `WalkParams` v1 builds and reads the path from upstream's `generate_with_vol`. There is no second copy of the walk mathematics to diverge from the mirrored ones. Exactly one option chain is built, to seed the walk; the tape stores none.

Per-step volatility comes from the same call: `generate_with_vol` reports an index-aligned volatility path for `Garch`, `Heston`, `Custom` and `Telegraph`, and `None` for the constant-volatility models — so a row's `base_volatility` is always the number that step's chains will actually be priced at.

## It removes an ambiguity v1 carries

Today a request may name a top-level `volatility: 0.18` **and** a `Brownian { volatility: 0.35 }`, and the service accepts it: step zero is priced at one number while the path is driven by the other. The tape resolves **one** authoritative base volatility — the model's, for the nine synthetic types — and rejects a request that specifies two different ones, naming the field.

`Historical` has no model volatility (upstream returns `None`, since it would have to be estimated from the series), so the parameters' value is used as a constant. Estimating it per step is deliberately left out: a trailing estimate that includes step *i* is look-ahead, and a causal estimator is its own decision, not a side effect of building a tape.

## Why `build` takes the method as an argument

A `Historical` walk must be resolved against ClickHouse first, and that seeded symbol-and-date selection already lives in `Simulator`. Keeping it outside leaves the tape builder **pure and synchronous** — no I/O, no wall clock — and therefore directly testable for the properties that matter. #46 wires the resolution.

## Tests — 21 new

The three reproducibility properties, each pinned:

- the same parameters rebuild an **identical** tape, so an eviction or a restart is invisible;
- a different seed diverges, while the shared starting state does not;
- **the expiration schedule cannot perturb the tape at all.** Adding twelve monthlies and three weeklies, or merely reordering the rules, leaves every row byte-identical. If chain building ever started consuming the walker's RNG, that is the test that catches it.

The rest: a three-year horizon producing every requested row despite hundreds of intervening expirations; row zero being the starting state; instants following the simulated clock rather than the wall clock; constant volatility staying constant; GARCH and Heston volatility varying and staying reproducible in **both** series; historical replay in order without look-ahead and rebuilding deterministically; the volatility-disagreement rejection; the step cap enforced at the boundary; and a row's size not depending on the chain shape.

## Verification

- `cargo fmt --all --check`
- `LOGLEVEL=WARN cargo test --workspace` — **495 passed, 0 failed, 13 ignored**, up from 474; plus 16 golden contract tests and 2 doctests
- `cargo build --release`
- `make readme` — no README change

No new dependency, no new env var, no change to any v1 path. `reproducibility-reviewer` is auditing the diff; findings land as follow-up commits before merge.

Closes #45

### DIFF
--- a/doc/adr/0001-v2-rolling-simulation-contract.md
+++ b/doc/adr/0001-v2-rolling-simulation-contract.md
@@ -301,7 +301,14 @@ simulation:
   `1..=12`;
 - an empty `schedules` array, or more rules than the configured cap (§9.3);
 - a projected inventory larger than the configured per-snapshot cap (§9.3);
-- any date arithmetic that would overflow while projecting the requested count.
+- any date arithmetic that would overflow while projecting the requested count;
+- a top-level `volatility` that disagrees with the walk model's own
+  `volatility`. A simulation has exactly one base volatility: v1 accepts the
+  pair and silently prices step zero at one while walking on the other, and v2
+  refuses the contradiction at the boundary rather than letting the domain pick
+  a winner. `Historical` carries no model volatility, so there is nothing to
+  disagree with. The check runs on the stored document too, so a hand-edited
+  simulation cannot smuggle the contradiction back in.
 
 ---
 
--- a/src/api/rest/requests_v2.rs
+++ b/src/api/rest/requests_v2.rs
@@ -57,6 +57,11 @@ pub struct CreateSimulationRequest {
     /// Initial price of the underlying.
     pub initial_price: f64,
     /// Initial (and, for constant-volatility models, the only) volatility.
+    ///
+    /// Must equal the walk model's own `volatility` where the model carries
+    /// one: a simulation has exactly one base volatility, and a disagreement is
+    /// a `400` naming this field rather than a silent choice made later
+    /// (ADR 0001 §4.4). `Historical` carries none, so nothing has to match.
     pub volatility: f64,
     /// Annualised risk-free rate, as a decimal fraction.
     pub risk_free_rate: f64,
--- a/src/domain/factors.rs
+++ b/src/domain/factors.rs
@@ -1,0 +1,822 @@
+//! The market-factor tape for v2 rolling simulations.
+//!
+//! A **factor row** is the whole market state at one step that is *not* an
+//! option: the simulated instant, the underlying price, and the base implied
+//! volatility that will price every chain at that step. The tape is the ordered
+//! sequence of those rows, and it is the deterministic source both the snapshot
+//! builder (#46) and the export (#49) read from.
+//!
+//! # Why a separate tape at all
+//!
+//! v1 caches a `RandomWalk<Positive, OptionChain>`, so the entire chain tape is
+//! materialised up front and the walk stops when its one expiry reaches zero
+//! (`src/domain/simulator.rs`). Neither works for a rolling simulation. The
+//! factor tape splits the two concerns: the market path runs for the **whole**
+//! requested horizon regardless of what expires along the way, and chains are
+//! built on demand from a row. Memory is `O(steps)` in small rows rather than
+//! `O(steps × expiries × strikes)` in contracts.
+//!
+//! # What makes it reproducible
+//!
+//! Three properties, each of them tested:
+//!
+//! - **The tape is a pure function of the effective parameters.** Same seed,
+//!   same start, same model ⇒ the same rows, on any machine.
+//! - **Expiration schedules cannot perturb it.** The planner draws no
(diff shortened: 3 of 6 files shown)


joaquinbejar:
Rebased onto main now that the session entry merged, and since this PR carried no review I ran the reproducibility audit the domain requires. It came back clean on the contract itself, every draw is walker-owned, no clock value reaches a row and no derived RNG stream is introduced, but it found four tests that could not fail and several doc claims that were not true, all fixed in cf17b96. The memory test asserted an identity twice over instead of comparing the wide and narrow tapes, the two volatility alignment tests only checked that the column varied so a reversed series passed, build trusted a steps >= 1 invariant that only validate establishes and would spin on steps 0, and the volatility agreement rule was only tested on the request path. The new 400 is now in the DTO doc and ADR §4.4 as well.

PR #61, opened 2026-08-14
feat(domain): build lazy rolling multi-expiration snapshots

### LINKED ISSUE #46: Build lazy rolling multi-expiration snapshots

joaquinbejar:
## Stack

- **Base branch:** `stack/4-45-factor-tape`
- **Stack:** #42 → #43 → #44 → #45 → **#46** → #47 → #48 → #49
- **Stacked on:** #60
- **Blocks:** #64, and through it #47 (serves snapshots) and #49 (exports them)

Targets `main` but is cut from `stack/4-45-factor-tape`, so **the diff against `main` is cumulative**. Review against the base branch. Merge bottom-up.

## Summary

Turns a factor row plus the planner's active expirations into the thing a client actually reads: the whole simulated market at one step — one clock, one spot, one base volatility, and the ordered chains alive at that instant.

**Built on demand, never stored for every step.** v1 materialises the whole chain tape up front at `O(steps × strikes)` and stops when its single expiry reaches zero. A rolling simulation multiplies that by the number of live expirations and runs for years — the reference configuration alone is fifteen chains a step. Building from a factor row keeps the resident cost at `O(steps)` small rows plus a bounded cache.

**Nothing here prices anything.** Every chain comes from upstream `OptionChain::build_chain` through the one wiring point the factor tape already uses. `generator_optionseries` is deliberately not used: it ages expirations and drops them but never replenishes them, and its `build_series` path does not preserve the requested strike interval.

## The subtle decision

Chains are built with `ExpirationDate::Days(fractional_dte)`, computed from the same `(simulated_at, expires_at)` pair the planner used — never `ExpirationDate::DateTime`, whose resolution goes through the wall clock and would make every premium depend on when the request arrived.

## One field this cannot make deterministic — and therefore does not surface

Upstream stamps each chain with a `YYYY-MM-DD` string from `get_date_string()`, which for a relative expiration reads `Utc::now()` and **ignores** the thread-local reference. There is no hook to change it.

A guard pinning that reference was written, **proved ineffective by a test that failed**, and removed rather than kept as reassurance. Nothing that reaches a price depends on the stamp — `get_years()` divides the `Days` value directly and `get_days()` returns it verbatim — so it is documented as upstream metadata in the module and in ADR §7, and excluded from the v2 DTOs. The expiration a client sees is the planner's absolute `expires_at`.

Reproducibility is therefore asserted over the **surfaced content** rather than `PartialEq` on the whole chain: timestamps, expirations, fractional DTE, labels, spot, base and per-strike volatility, quotes and Greeks — exactly what #47 will put on the wire. That comparison cannot pass vacuously, and cannot fail across midnight for a reason that is not a regression.

## The snapshot cache

Bounded, least-recently-accessed, keyed by `(simulation, step)` so one simulation cannot evict another wholesale and #48 can drop every entry belonging to a session that has gone away. **Eviction is unobservable by construction** — a rebuilt snapshot is identical because it is a pure function of the factor row and the planner — which is what lets the bound be small. A test asserts exactly that.

## Tests — 24 new

**Inventory:** the reference configuration satisfies all sixteen rule slots at every step while coincident dates collapse to fifteen chains carrying both labels; chains chronological and unique; every chain shares the step's spot and base volatility while carrying its own expiration and fractional DTE; an expired chain is never emitted; crossing a 17:00 cutoff replaces the rolling expiration **in the same snapshot** with no gap; a ninety-step horizon builds throughout.

**Pricing:** a higher base volatility raises the ATM premium; a nearer expiration carries less extrinsic value; skew and smile vary the volatility across strikes.

**Cache:** bound respected, least-recently-accessed evicted, per-simulation eviction, re-insert replaces, zero capacity raised to one, and eviction unobservable.

## Verification

- `cargo fmt --all --check`
- `LOGLEVEL=WARN cargo test --workspace` — **519 passed, 0 failed, 13 ignored**, up from 495; plus 16 golden contract tests and 2 doctests
- `cargo build --release`

No new dependency, no new env var beyond the documented `OCS_MAX_CACHED_SNAPSHOTS` knob (which #48 folds into the typed v2 configuration), and no change to any v1 path.

Closes #46

### DIFF
--- a/doc/adr/0001-v2-rolling-simulation-contract.md
+++ b/doc/adr/0001-v2-rolling-simulation-contract.md
@@ -438,6 +438,15 @@ Invariants a reviewer can check:
   the ladder, not a copy of it.
 - All numeric fields are `f64` **at the REST boundary only**; the domain works
   in `Positive`/`Decimal`.
+- `expires_at` is the **only** expiration a client sees. Upstream's
+  `OptionChain` also carries a `YYYY-MM-DD` string that it stamps from the host
+  clock — `get_date_string()` reads `Utc::now()` for a relative expiration and
+  ignores the thread-local reference, and there is no hook to change that. It is
+  upstream metadata, it is not part of this contract, and the v2 DTOs must not
+  surface it. Nothing that reaches a price depends on it: premiums, Greeks and
+  the derived strike interval all come from the fractional days value directly,
+  so the snapshot's *surfaced* content stays a pure function of the effective
+  inputs.
 - The identifier field is `id` and the resource is a *simulation*; "session" is
   v1 vocabulary and is not used on the v2 wire.
 - `state` is serialised **`snake_case`** — `initialized`, `in_progress`,
--- a/src/domain/factors.rs
+++ b/src/domain/factors.rs
@@ -122,8 +122,9 @@ impl FactorTape {
     /// Returns [`ChainError::Validation`] when the parameters are invalid, when
     /// the model's volatility disagrees with the parameters' (see
     /// [`resolve_base_volatility`]), when a `Historical` series is too short
-    /// for the horizon, or when the simulated clock overflows; and
-    /// [`ChainError::Internal`] when the resolved method is not the one the
+    /// for the horizon, when a stochastic-volatility path exceeds the 1.0 an
+    /// option chain can be priced at, or when the simulated clock overflows;
+    /// and [`ChainError::Internal`] when the resolved method is not the one the
     /// parameters name, when the initial chain cannot be built, or when the
     /// walk returns fewer points than requested.
     #[instrument(skip(parameters, method), level = "debug")]
@@ -140,6 +141,7 @@ impl FactorTape {
         ensure_method_matches(parameters, method)?;
         ensure_historical_series_covers_the_horizon(parameters, method)?;
         let base_volatility = resolve_base_volatility(parameters, method)?;
+        reject_unpriceable_volatility(base_volatility, None)?;
         let walker = Walker::new_with_seed(parameters.seed);
 
         // The walk starts from an `OptionChain` because that is the shape v1's
@@ -206,6 +208,8 @@ impl FactorTape {
                 })?,
                 None => base_volatility,
             };
+
+            reject_unpriceable_volatility(row_volatility, Some(step))?;
 
             rows.push(FactorRow {
                 step,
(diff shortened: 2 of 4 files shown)


joaquinbejar:
Rebased onto main now that the factor tape merged, and since this PR carried no review I ran the reproducibility audit the domain requires. The isolation property holds structurally, snapshot building consumes no randomness at all, but three tests were not testing it: upstream's OptionChain PartialEq compares only the expiration string and the symbol, so deriving it on ExpiryChain gave the rebuild, same-seed and eviction tests a comparison that passes with completely different chains. It is hand-written now, and the schedule-isolation test this PR owed is added. Two reachable failure modes are closed too, a Garch or Heston volatility path crossing 1.0 would have 500'd halfway through a run and is now a 400 at tape build naming the step, and SeriesBuilder validates its parameters the way FactorTape::build does. The per-snapshot work cap and a weight-aware cache bound are product decisions I did not invent, they are tracked in #62. All in ec89f25, gate green with 534 unit tests.

Issue #63, opened 2026-08-14
Restore a causal volatility estimate for v2 historical walks


joaquinbejar:
## Overview

A v2 rolling simulation driven by a `Historical` walk prices every step at the constant `volatility` the request supplied. v1 prices the same walk with a rolling realized-volatility estimate. Close that gap, or decide deliberately that v2 does not offer one.

## Context

`WalkType::volatility()` returns `None` for `Historical` — the volatility is not a parameter of the model, it has to be estimated from the price series. v1 gets one: `walk_steps_par` prices its chains with upstream's `expanding_window_vols`, whose estimate at index `i` uses only the log returns of `prices[..=i]`, so it is causal and carries no look-ahead.

v2 does not walk through that driver. The whole point of the factor tape (#45) is to stop materialising a chain per step, and `expanding_window_vols` is private to the driver v2 no longer uses. So `resolve_base_volatility` falls back to the requested constant (`src/domain/factors.rs`), which is documented there and in #45's commit as a deliberate simplification.

The consequence is a behaviour difference a client can see: the same historical series, the same seed, run through v1 and v2, produces different premiums — not because the price path differs (it does not, #45 pins that) but because the volatility used to price it does.

## Options

1. **Ask upstream to expose the estimator.** `expanding_window_vols` is already written, already causal, and already the reference behaviour. Making it public in `optionstratlib` is the smallest change that gives v2 parity without a second copy of the mathematics. Preferred.
2. **Port it into the factor tape.** Fastest, but it duplicates upstream mathematics inside a module whose stated premise is not to — the same divergence risk the `Walker` already carries for the walk kernels, and CLAUDE.md's re-sync burden grows by one more mirror.
3. **Reject `Historical` for v2.** `POST /api/v2/simulations` returns a `400` naming `method` until an estimator exists. Honest, and it stops shipping a quiet difference, but it removes a model v1 offers.

## Tasks

- [ ] Decide between the three options above; if it is (1), open the upstream issue and link it here.
- [ ] Whichever is chosen, make the per-step base volatility of a historical v2 tape explicit in the module docs and in ADR 0001, so the v1/v2 relationship is stated rather than inferred.
- [ ] If an estimator lands: feed it into `FactorRow.base_volatility` and keep the row's value the exact number that step's chains are priced at, which is what #45's contract already promises.

## Acceptance criteria

[...]

Issue #56, opened 2026-08-14
Persist v2 simulation snapshots in ClickHouse


joaquinbejar:
## Overview

Persist generated v2 simulation snapshots in ClickHouse as an immutable, queryable tape instead of storing complete nested chain documents in MongoDB. Store snapshot metadata separately from flattened option-quote rows so the service can reconstruct a full chain efficiently while also serving contract histories, backtests, and JSON/CSV exports without unwinding BSON arrays.

MongoDB remains available for operational events/audit data, and Redis/in-memory stores continue to own the live session cursor and state. `/api/v1/chain` and its current MongoDB logging behavior remain unchanged.

## Context

The current v1 advance handler sends the complete `ChainResponse` to `MongoDBRepository::save_chain_step` (`src/infrastructure/repositories/mongo_repo.rs`), which inserts one nested document per step. That shape is convenient for whole-document retrieval but is not suitable for the v2 workload: one simulated timestamp can contain many expirations and strikes, while frontend and backtest consumers primarily query one contract across time or scan normalized rows.

The repository already uses ClickHouse for historical-market-data input under `src/infrastructure/clickhouse/*`. Reuse that infrastructure and introduce a dedicated snapshot repository rather than adding another time-series database.

This issue depends on the canonical domain snapshot from #46 and must coordinate with the v2 lifecycle in #47 and exports in #49. It also amends the warehouse-persistence decision recorded by #42/#49: persisted ClickHouse rows become the primary source for materialized snapshots and exports, while deterministic replay remains the recovery, validation, and backfill mechanism.

The persistence boundary must accept domain/session snapshot data or dedicated typed persistence records. Infrastructure must not import REST DTOs from `src/api`.

## Storage model

Add two logical ClickHouse tables (exact engine/partition choices must be documented against the query patterns):

### `simulation_snapshots`

One row per materialized simulation step, including at least:

- `snapshot_id` — deterministic identity derived from simulation id/version and step
- `simulation_id`, `simulation_version`, `step`
- `simulated_at`, `symbol`
- `underlying_price`, `base_volatility`
- expected quote-row count and persistence/completion metadata required to reject partial snapshots

### `simulation_option_quotes`

One row per snapshot + physical expiration + strike, including at least:

- `snapshot_id`, `simulation_id`, `simulation_version`, `step`, `simulated_at`
- absolute `expires_at`, fractional DTE, and all matching schedule labels
- `strike`, per-strike implied volatility
[...]

PR #68, opened 2026-08-15
Bound the work one snapshot prices and the cache by contracts (#62)

### LINKED ISSUE #62: Bound the per-snapshot work and the snapshot cache by contracts, not by entries
## Context

Found while auditing #61 (rolling snapshots). Each individual cap holds, but nothing bounds their product, and the snapshot cache counts entries rather than weight. Neither is exploitable today — the snapshot layer has no caller until #47 serves it — but both must be closed before it does.

## The two gaps

**Per-snapshot work is a product of two caps.** `MAX_EXPIRATIONS_PER_SNAPSHOT` is 512 (`src/domain/expiry.rs`) and `MAX_CHAIN_SIZE` is 500, which is 1001 strikes (`src/api/rest/limits.rs`). A single snapshot request can therefore price 512 × 1001 ≈ 512 000 contracts, each a Black-Scholes evaluation with Greeks — roughly a million evaluations and hundreds of MB, from one request that violates no current validation. ADR 0001 §9.3 anticipates the work bound for export but not for a single snapshot.

**The snapshot cache is bounded in entries.** `DEFAULT_MAX_CACHED_SNAPSHOTS` is 256 (`src/domain/series.rs`). The doc reasons about a snapshot being heavy, but the number does not: 256 worst-case snapshots is tens of GB resident.

## What needs deciding

The cap itself is a product decision, which is why this is an issue rather than a patch. Two shapes worth considering:

- a combined limit validated at creation — `chain_size × Σ target_count <= N`, rejected with a `400` naming `chain_size`, so the request that cannot be served is refused before a tape is built;
- a cache bounded by total `OptionData` count rather than entry count, which is the honest unit for a memory bound.

Both belong with #48, which turns the v2 bounds into validated `OCS_MAX_*` environment knobs following the `LazyLock` pattern in `api::rest::limits` — so whatever number is chosen should arrive as a knob, not a constant.

## Acceptance

- A request whose expiration count times chain size exceeds the chosen bound is rejected at creation with a field-specific `400`, with a test.
- The snapshot cache's bound is expressed in contracts, with a test that a cache of large snapshots evicts before it would have on entry count.
- ADR 0001 §9.3 records both numbers alongside the export bound it already documents.

joaquinbejar:
## Summary

Each cap held individually; their product did not. `OCS_MAX_CHAIN_SIZE` of 500 is 1 001 strikes, a schedule may keep 512 expirations alive, and one `/snapshot` call could therefore price half a million contracts — about a million Black-Scholes evaluations and hundreds of MB — from a request that violated neither bound. Separately, the snapshot cache counted entries, which says nothing about memory when one entry is a few hundred contracts in the reference configuration and hundreds of thousands in a large one.

## Changes

**Per-snapshot work.** `SimulationParametersV2::validate` now bounds `strikes × Σ target_count` against `OCS_MAX_SNAPSHOT_CONTRACTS` (default 200 000).

- Checked once at creation, not per step.
- Uses the **pre-deduplication** sum, the same tight upper bound §9.3 already uses for the expiration cap — rules claiming the same date are priced once, so a configuration this accepts can never exceed the cap.
- Reported against `chain_size`, the field a client can lower without changing what the simulation means.
- `strikes_per_chain` is `2n + 1`, because upstream's `chain_size` is a per-side half-width.

**Cache weight.** `SnapshotCache` gains a contract budget beside its entry bound (`OCS_MAX_CACHED_SNAPSHOT_CONTRACTS`, default 4 000 000 ≈ a gigabyte of `OptionData`), evicting on whichever binds first and sharing one victim rule. A snapshot heavier than the whole budget is still cached — refusing to hold anything would turn a large-but-legal configuration into a permanently cold cache, and the per-snapshot cap is what bounds that case.

**Where the knobs live.** `OCS_MAX_SNAPSHOT_CONTRACTS` follows the `limits.rs` `LazyLock` pattern, next to the two caps it multiplies with and reachable from `validate` without threading config through the session layer. `OCS_MAX_CACHED_SNAPSHOT_CONTRACTS` follows the startup-validated v2 config from #48, next to the entry bound it complements. Both are documented in `.env.example`, ADR 0001 §9.3 and the crate docs.

## Testing

- [x] A configuration at both caps (500 chain size, two rules at the per-rule cap) is refused, naming `chain_size`
- [x] The reference configuration validates comfortably, so the bound costs a realistic client nothing
- [x] Eviction fires on the contract budget while the entry bound of ten is nowhere near, and the cache never exceeds its budget
- [x] A snapshot larger than the whole budget is still served
- [x] `make pre-push` — 613 unit tests plus 16 v1-contract tests, clippy and fmt clean; `cargo build --release` clean

## Stack

- **Stack:** #63 → #62 → #56 — **this is PR 2**
- **Base branch:** `stack/1-63-historical-volatility-parity` (PR #67)
- **Stacked on:** #67 · **Blocks:** the ClickHouse PR
- The diff is cumulative until #67 merges — review it against the base branch, not `main`.

Closes #62

### DIFF
--- a/src/api/rest/limits.rs
+++ b/src/api/rest/limits.rs
@@ -11,6 +11,7 @@
 //! | `OCS_MAX_STEPS`            | `10_000`  | Max simulation steps per session         |
 //! | `OCS_MAX_CHAIN_SIZE`       | `500`     | Max option-chain size per request        |
 //! | `OCS_MAX_HISTORICAL_PRICES`| `100_000` | Max historical prices in a walk request  |
+//! | `OCS_MAX_SNAPSHOT_CONTRACTS`| `200_000`| Max contracts one v2 snapshot may price |
 
 use std::sync::LazyLock;
 use tracing::warn;
@@ -21,6 +22,16 @@ pub(crate) const DEFAULT_MAX_STEPS: usize = 10_000;
 pub(crate) const DEFAULT_MAX_CHAIN_SIZE: usize = 500;
 /// Default cap on the number of historical prices in a `Historical` walk request.
 pub(crate) const DEFAULT_MAX_HISTORICAL_PRICES: usize = 100_000;
+/// Default cap on the contracts one v2 snapshot may price.
+///
+/// A snapshot prices every strike of every live expiration, so its cost is the
+/// product of two caps that each look reasonable alone: `OCS_MAX_CHAIN_SIZE`
+/// of 500 is 1 001 strikes, and a schedule may keep 512 expirations alive, for
+/// half a million contracts in one request. The default admits any realistic
+/// configuration — ADR 0001's reference schedule prices about 500 contracts a
+/// snapshot, and 200 000 still allows 200 expirations at 1 001 strikes — while
+/// refusing the shapes that exist only to exhaust the service.
+pub(crate) const DEFAULT_MAX_SNAPSHOT_CONTRACTS: usize = 200_000;
 
 /// Maximum number of simulation steps a session may request (`OCS_MAX_STEPS`).
 pub(crate) static MAX_STEPS: LazyLock<usize> =
@@ -42,6 +53,26 @@ pub(crate) static MAX_HISTORICAL_PRICES: LazyLock<usize> = LazyLock::new(|| {
         DEFAULT_MAX_HISTORICAL_PRICES,
     )
 });
+
+/// Maximum number of contracts one v2 snapshot may price
+/// (`OCS_MAX_SNAPSHOT_CONTRACTS`).
+pub(crate) static MAX_SNAPSHOT_CONTRACTS: LazyLock<usize> = LazyLock::new(|| {
+    parse_limit(
+        std::env::var("OCS_MAX_SNAPSHOT_CONTRACTS").ok(),
+        DEFAULT_MAX_SNAPSHOT_CONTRACTS,
+    )
+});
+
+/// The number of strikes a chain of `chain_size` carries.
+///
+/// `chain_size` is upstream's per-side half-width — it counts strikes above
+/// *and* below the money — so the ladder is `2n + 1` wide. Saturating, because
+/// the caller may be validating a value that has not been range-checked yet and
+/// an overflow here would be the wrong error to report.
+#[must_use]
+pub(crate) fn strikes_per_chain(chain_size: usize) -> usize {
+    chain_size.saturating_mul(2).saturating_add(1)
+}
 
 /// Parses a raw environment value into a positive `usize` limit.
 ///
--- a/src/api/rest/limits.rs
+++ b/src/api/rest/limits.rs
@@ -11,7 +11,6 @@
 //! | `OCS_MAX_STEPS`            | `10_000`  | Max simulation steps per session         |
 //! | `OCS_MAX_CHAIN_SIZE`       | `500`     | Max option-chain size per request        |
 //! | `OCS_MAX_HISTORICAL_PRICES`| `100_000` | Max historical prices in a walk request  |
-//! | `OCS_MAX_SNAPSHOT_CONTRACTS`| `200_000`| Max contracts one v2 snapshot may price |
 
 use std::sync::LazyLock;
 use tracing::warn;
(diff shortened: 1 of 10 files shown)


joaquinbejar:
Blocks #69, which is stacked on this branch.

src/api/rest/limits.rs
```diff
@@ -43,6 +54,26 @@ pub(crate) static MAX_HISTORICAL_PRICES: LazyLock<usize> = LazyLock::new(|| {
     )
 });
 
+/// Maximum number of contracts one v2 snapshot may price
+/// (`OCS_MAX_SNAPSHOT_CONTRACTS`).
+pub(crate) static MAX_SNAPSHOT_CONTRACTS: LazyLock<usize> = LazyLock::new(|| {
```
joaquinbejar: 🟠 **Warning**: this new operational bound is read lazily through parse_limit, so a malformed OCS_MAX_SNAPSHOT_CONTRACTS silently falls back after startup. New environment configuration must be loaded through a typed config and fail startup with the variable name; otherwise the deployed work bound can differ from what the operator requested.

src/api/rest/limits.rs
```diff
@@ -43,6 +54,26 @@ pub(crate) static MAX_HISTORICAL_PRICES: LazyLock<usize> = LazyLock::new(|| {
...
+
+/// The number of strikes a chain of `chain_size` carries.
+///
+/// `chain_size` is upstream's per-side half-width — it counts strikes above
+/// *and* below the money — so the ladder is `2n + 1` wide. Saturating, because
+/// the caller may be validating a value that has not been range-checked yet and
+/// an overflow here would be the wrong error to report.
+#[must_use]
+pub(crate) fn strikes_per_chain(chain_size: usize) -> usize {
+    chain_size.saturating_mul(2).saturating_add(1)
```
joaquinbejar: 🟠 **Warning**: saturating arithmetic is explicitly forbidden by the repository rules because it hides overflow. Return a checked result and map overflow to the existing chain_size validation error instead of silently changing the requested width.

src/domain/series.rs
```diff
@@ -377,6 +408,16 @@ impl SnapshotCache {
...
         self.evict_to(self.capacity - 1);
+
+        // Make room for this snapshot's own weight before inserting it, so the
+        // cache never exceeds its budget afterwards. A single snapshot larger
+        // than the whole budget is still cached — evicting everything and then
+        // refusing to hold anything would turn a large-but-legal configuration
+        // into a permanently cold cache — and the per-snapshot contract cap in
+        // `SimulationParametersV2::validate` is what bounds that case.
+        let incoming = snapshot_contracts(&snapshot);
+        self.evict_to_contracts(self.contract_budget.saturating_sub(incoming));
```
joaquinbejar: 🟠 **Warning**: when the incoming weight exceeds contract_budget, saturating_sub yields zero, all existing entries are evicted, and the oversized snapshot is inserted anyway. That violates OCS_MAX_CACHED_SNAPSHOT_CONTRACTS, which is documented as the maximum resident contracts; skip caching an overweight entry or reject incompatible limits at startup.

src/domain/series.rs
```diff
@@ -397,20 +438,38 @@ impl SnapshotCache {
         before - self.entries.len()
     }
 
+    /// Evicts least-recently-accessed entries until the cache holds at most
+    /// `max` contracts.
+    ///
+    /// Shares the victim rule with [`SnapshotCache::evict_to`], so which entry
+    /// goes is the same question either bound asks.
+    fn evict_to_contracts(&mut self, max: usize) {
+        while self.contracts() > max {
```
joaquinbejar: 💡 **Suggestion**: every eviction iteration calls contracts(), rescanning every remaining quote, then scans the entries again to find the least recently used item. At the documented multi-million-contract budget this runs under the manager mutex and can stall unrelated requests. Store each entry weight plus a running total so eviction is linear in the number of victims.

joaquinbejar (commented):
The new resource limits target the correct failure modes, but their configuration can silently default, overflow is hidden, and the snapshot cache can exceed its declared contract budget while repeatedly rescanning it during eviction. REQUEST_CHANGES.

src/api/rest/limits.rs
joaquinbejar: Moved it into SimulationV2Config with the other v2 knobs, so it is parsed and range-checked once at startup and a malformed value fails the boot naming the variable. The validator has no config handle, so the parsed cap reaches it through a OnceLock the config publishes, which keeps the parsing in exactly one place.

src/api/rest/limits.rs
joaquinbejar: Agreed, and it was the wrong tool here specifically: the caller is a validator whose job is to reject what it cannot serve, so a saturating width would have answered a different question than it was asked. It returns an Option now and an overflow is a validation error naming chain_size.

src/domain/series.rs
joaquinbejar: Fixed by skipping it, which is the option you named first. An oversized snapshot is no longer cached at all, so the budget holds and the entries that do fit are left alone; the step is still served, since an uncached step is a rebuild and that is what every miss already is.

src/domain/series.rs
joaquinbejar: Done as you described. Each entry stores the weight it was measured at on insert and the cache keeps a running total, so the budget check is O(1) and eviction is linear in the victims; every removal path goes through one helper that keeps the total honest.

src/domain/series.rs
```diff
@@ -155,6 +155,19 @@ impl SeriesSnapshot {
...
+///
+/// The unit the cache budgets in: one priced option per strike per live
+/// expiration, which is what makes a snapshot heavy.
+#[must_use]
+fn snapshot_contracts(snapshot: &SeriesSnapshot) -> usize {
+    snapshot
+        .chains
+        .iter()
+        .map(|chain| chain.chain.iter().count())
+        .sum()
```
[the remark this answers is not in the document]
joaquinbejar: Right, and it mattered twice over since eviction called it per victim. Counting goes through the chain's own length now, and the cache stores each entry's weight plus a running total so the budget check no longer walks anything.

joaquinbejar:
All four are fixed in 2780579, and the branch is rebased on main now that the volatility entry merged. The cap moved into SimulationV2Config so a malformed value fails the boot instead of silently defaulting, strikes_per_chain is checked with the overflow reported against chain_size, an oversized snapshot is no longer cached at all so the budget genuinely holds, and each entry now carries its weight with a running total so eviction stops rescanning under the lock. Gate is green with 613 unit tests plus the 16 contract tests.

PR #69, opened 2026-08-15
Persist v2 simulation snapshots in ClickHouse (#56)

### LINKED ISSUE #56: Persist v2 simulation snapshots in ClickHouse

joaquinbejar:
## Summary

v2 snapshots had exactly one read path: replay. That is the right canonical answer and a poor operational one — a client wanting one contract's premium across a year had to walk the whole tape, and a restart lost everything a run had produced.

Snapshots are now filed in ClickHouse as an immutable, queryable tape. **Opt-in** (`OCS_SNAPSHOT_PERSISTENCE_ENABLED`, default off) and absent by construction when off: the manager holds `Option<Sender>`, so a deployment without ClickHouse does not have to name the feature to not use it.

## Storage model

Two tables, because the two query patterns are different shapes: one metadata row per step in `simulation_snapshots`, and one flattened row per `(step, expiration, strike)` in `simulation_option_quotes` — which is what turns "this contract over time" into a range scan rather than an unwind of nested documents.

- `ReplacingMergeTree(inserted_at_ms)`, ordered on the coordinate that identifies a row; `simulation_option_quotes` adds a minmax index on `(expires_at, strike)` for the contract-history pattern.
- `PARTITION BY toYYYYMM(simulated_at)` — **content**-derived on purpose. The engine only deduplicates within a partition, so an ingestion-derived key would let a backfill written a month later survive as a permanent duplicate. Retention is a row-level TTL anchored on ingestion time, so a backfill still keeps its full window.
- Money and Greeks are `Decimal(38, 28)`, timestamps `DateTime64(9)`, so a value reconstructs exactly rather than approximately.
- DDL is checked in under `src/infrastructure/clickhouse/schema/` and `include_str!`-ed, so the SQL and the Rust row structs are one artifact — the driver validates the struct against the live schema on every first insert.

## The four properties that make it safe to read

**Identity is derived, not generated.** A row's id comes from `(simulation, tape generation, step)`, and the tables collapse on that sorting key, so retrying a step replaces its rows instead of adding a copy. The generation is the *tape's*, not the session's compare-and-swap revision — filing under a revision that changes on every advance would scatter one simulation across as many generations as it has steps.

**A partial write never reads as complete.** Quote rows go first, in one batch; the metadata row carrying the expected count goes last. Every read compares what it got against that count and treats a mismatch as absent, so a torn write reads as a missing step rather than a short chain.

**Filing cannot fail or delay an advance.** It happens after the cursor commits, through a bounded queue that one writer task drains. A full queue drops the record with a `WARN` naming the step — the same trade as a failed write, made explicit: the step stays reproducible, replay rebuilds it, a retry writes the same rows.

**Nothing about v1 moves.** MongoDB stays event and audit only; no v2 chain is written as a nested document; `/api/v1/chain` routes, DTOs and logging are untouched (the 16 contract tests still pass).

## v2 simulation ids become random

`UuidGenerator` derives from a process-local counter starting at zero — v1 abandoned it for exactly this reason. That was survivable while an id only had to be unique among live sessions. It is not now that the id is a durable row key for the retention window: a second run would write the first run's coordinates, and either the engine collapses two different tapes into one or the completeness guard marks the step permanently absent.

The id reaches nothing seeded — the tape, the planner and the snapshot builder are functions of the parameters alone — so the same-seed ⇒ identical-tape contract is untouched. A test now pins that: two simulations sharing every parameter get different ids and produce identical snapshots, compared premium by premium.

`SessionV2::new` drops the generator argument it was ignoring rather than keeping a signature that lies to an external caller.

## Reviews run before this was opened

Both were on the diff, both are reflected in it.

- **architect** — found the id collision above as blocking, plus: docs claiming "every served snapshot" when only an advance files; the write being inline and able to hold a response for the insert timeout; a `version` field that would be confused with the CAS revision on the wire; the read half of the API being unusable externally because its generation constant was crate-private. All fixed.
- **reproducibility-reviewer** — confirmed the id change does not touch the seed contract, the stored-session shape or any test's meaning. Its findings: the detached write was unbounded (now a bounded queue), an empty-quotes expiration could disappear on read, the constructor lying about its argument, and the missing test for the property the id change rests on. Fixed.

## Testing

- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean
- [x] `cargo fmt --all --check` clean
- [x] `LOGLEVEL=WARN cargo test --workspace` — 692 unit tests, 16 v1-contract tests, 3 doctests, 18 ignored
- [x] `cargo build --release` clean
- [x] The 5 `#[ignore]`d live tests pass against the **exact CI image** (`clickhouse/clickhouse-server:26.3.17.110`): round-trip, double-persist idempotence, torn write, short batch, contract history. A column rename is precisely what the driver's schema validation catches, so this was run after the rename, not before.
- [x] CI needs no change — the integration job already runs `cargo test --workspace -- --ignored` with a ClickHouse service whose credentials match `ClickHouseConfig::default()`.

## Left deliberately

The export path (#49) still replays rather than reading the warehouse. Persistence is off by default, so replay is the only source that is always correct today; making the warehouse preferred is a follow-up that needs its own decision about what a partially-persisted range should return. Noted rather than half-done.

## Stack

- **Stack:** #63 → #62 → #56 — **this is PR 3 (top)**
- **Base branch:** `stack/2-62-snapshot-work-caps` (PR #68)
- **Stacked on:** #68 · **Blocks:** — (last in the chain)
- The diff is cumulative until #67 and #68 merge — review it against the base branch, not `main`.

Closes #56

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -133,10 +133,14 @@ OCS_MAX_EXPORT_ROWS=100000
 # Values: true/false/1/0/yes/no/on/off.  Default: false
 OCS_SNAPSHOT_PERSISTENCE_ENABLED=false
 
-# Rows per insert batch. Every quote row of one snapshot goes in ONE insert, so
-# this bounds a single statement, not the snapshot.
-# Range: 1 .. 5000000.  Default: 100000
-OCS_SNAPSHOT_BATCH_ROWS=100000
+# Maximum quote rows one snapshot may carry into storage. Every quote row of a
+# snapshot goes in ONE insert, so this bounds the snapshot and the statement
+# alike: a snapshot above it is rejected before a row is written. It must be at
+# least OCS_MAX_SNAPSHOT_CONTRACTS, or the API would accept simulations the
+# warehouse could never persist — startup refuses that combination rather than
+# leaving it to be found in the logs.
+# Range: 1 .. 5000000.  Default: 200000 (matches OCS_MAX_SNAPSHOT_CONTRACTS)
+OCS_SNAPSHOT_BATCH_ROWS=200000
 
 # Maximum rows one read may return, so a range query cannot be unbounded.
 # Range: 1 .. 10000000.  Default: 1000000
--- a/.env.example
+++ b/.env.example
@@ -118,6 +118,40 @@ OCS_MAX_CACHED_SNAPSHOT_CONTRACTS=4000000
 # Range: 1 .. 10000000.  Default: 100000
 OCS_MAX_EXPORT_ROWS=100000
 
+# ---------------------------------------------------------------------------
+# v2 snapshot persistence (ClickHouse)
+# ---------------------------------------------------------------------------
+# Files every ADVANCED v2 step in ClickHouse as an immutable, queryable tape:
+# one metadata row per step plus one flattened row per expiration and strike.
+# A peek and an export replay and persist nothing.
+# OFF by default — a deployment without ClickHouse does not have to name the
+# feature to not use it, and with it off nothing is added to the serving path.
+# Turning it ON makes ClickHouse a HARD STARTUP DEPENDENCY: the tables are
+# created at boot, so a schema or connectivity problem fails the boot rather
+# than surfacing later as a warning. Writes themselves are detached from the
+# request, so a warehouse that degrades later cannot delay a response.
+# Values: true/false/1/0/yes/no/on/off.  Default: false
+OCS_SNAPSHOT_PERSISTENCE_ENABLED=false
+
+# Rows per insert batch. Every quote row of one snapshot goes in ONE insert, so
+# this bounds a single statement, not the snapshot.
+# Range: 1 .. 5000000.  Default: 100000
+OCS_SNAPSHOT_BATCH_ROWS=100000
+
+# Maximum rows one read may return, so a range query cannot be unbounded.
+# Range: 1 .. 10000000.  Default: 1000000
+OCS_SNAPSHOT_MAX_READ_ROWS=1000000
+
+# How long an insert may take before it is an error the advance tolerates.
+# Range: 1 .. 600.  Default: 30
+OCS_SNAPSHOT_INSERT_TIMEOUT_SECS=30
+
+# How long persisted snapshots are kept, as a row-level TTL. Anchored on the
+# ingestion timestamp, so a backfilled step is kept for its full window.
+# Changing this on a live deployment needs an explicit ALTER TABLE MODIFY TTL.
+# Range: 1 .. 3650.  Default: 90
+OCS_SNAPSHOT_RETENTION_DAYS=90
+
 
 # ---------------------------------------------------------------------------
 # Redis — session and simulation storage
--- a/src/infrastructure/config/snapshot.rs
+++ b/src/infrastructure/config/snapshot.rs
@@ -1,0 +1,476 @@
+//! Operational configuration for persisting v2 simulation snapshots.
+//!
+//! The snapshot tape is written to ClickHouse (issue #56) as two flat tables
+//! rather than to MongoDB as nested documents. Everything an operator can tune
+//! about that write path lives here: whether it happens at all, how large one
+//! snapshot may be, how long the service waits for the insert, how much one
+//! read may return, and how long the rows are kept.
+//!
+//! # Why this validates instead of falling back
+//!
+//! Same reasoning as [`crate::infrastructure::SimulationV2Config`]: a
+//! silently-defaulted retention would delete a tape a client still expects to
+//! be able to export, and a silently-defaulted row bound would change the
+//! service's memory profile without anyone noticing. So an unset knob takes its
+//! documented default and a **set-but-invalid** one fails startup with a
+//! message naming the variable.
+//!
+//! # Why persistence is off by default
+//!
+//! ClickHouse is optional in this service — `Simulator::new` already tolerates
+//! its absence — and a deployment without it must not log a failed write on
+//! every advance. Persistence is therefore opt-in: an operator who has the
+//! warehouse turns it on explicitly.
+
+use crate::utils::ChainError;
+use std::env;
+use std::time::Duration;
+use tracing::info;
+
+/// Whether snapshot persistence is on when `OCS_SNAPSHOT_PERSISTENCE_ENABLED`
+/// is unset.
+///
+/// Off: see the module docs.
+pub const DEFAULT_SNAPSHOT_PERSISTENCE_ENABLED: bool = false;
+
+/// Default cap on the quote rows one snapshot may carry, in rows.
+///
+/// A snapshot is `expirations × strikes` rows; ADR 0001's reference
+/// configuration is a few hundred, and the per-snapshot contract budget in
+/// `SimulationParametersV2::validate` already bounds what the domain will
+/// price. This is the *storage* bound: a snapshot larger than it is rejected
+/// before a single row is written, so one pathological simulation cannot turn
+/// into an unbounded insert.
+pub const DEFAULT_SNAPSHOT_BATCH_ROWS: usize = 100_000;
+
+/// Default cap on the rows one read may return, in rows.
+///
+/// Bounds `stream_range` and `contract_series`, which #49's exports page
+/// through. A range that would exceed it is a typed error naming the knob
+/// rather than a truncated answer, because a silently short tape is
+/// indistinguishable from a gap in the data.
+pub const DEFAULT_SNAPSHOT_MAX_READ_ROWS: usize = 1_000_000;
+
+/// Default insert timeout, in seconds.
+///
+/// Covers both sending the batch and waiting for the server to accept it. A
+/// snapshot that cannot be written in half a minute is a degraded warehouse,
+/// and the advance path must not block on it.
+pub const DEFAULT_SNAPSHOT_INSERT_TIMEOUT_SECS: u64 = 30;
+
+/// Default retention for persisted snapshot rows, in days.
+///
+/// Applied as a row-level `TTL` on the ingestion timestamp when the schema is
+/// created — real time, not simulated time, because a simulation's clock may
+/// sit years in the future.
+pub const DEFAULT_SNAPSHOT_RETENTION_DAYS: u32 = 90;
+
+/// The largest per-snapshot row bound that can be configured.
+///
+/// Five million rows is far past any snapshot the domain will price, so a
+/// larger value is a typo rather than an intent.
+const MAX_BATCH_ROWS: usize = 5_000_000;
+
+/// The largest per-read row bound that can be configured.
+const MAX_READ_ROWS_CEILING: usize = 10_000_000;
+
+/// The longest insert timeout that can be configured, in seconds.
+const MAX_INSERT_TIMEOUT_SECS: u64 = 600;
+
+/// The longest retention that can be configured, in days — ten years.
+const MAX_RETENTION_DAYS: u32 = 3_650;
+
+/// Operational limits for the v2 snapshot tape in ClickHouse.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SnapshotPersistenceConfig {
+    /// Whether generated snapshots are written to ClickHouse at all.
+    ///
+    /// Consumed at construction:
+    /// [`crate::infrastructure::ClickHouseSnapshotRepository::from_env`] returns
+    /// `None` when this is false, so a disabled deployment never holds a
+    /// repository and never attempts a write.
+    pub enabled: bool,
+    /// The most quote rows one snapshot may carry.
+    pub batch_rows: usize,
+    /// The most rows one read may return.
+    pub max_read_rows: usize,
+    /// How long an insert may take, sending and server-side combined.
+    pub insert_timeout: Duration,
+    /// How long persisted rows are kept, in days, measured from ingestion.
+    pub retention_days: u32,
+}
+
+impl Default for SnapshotPersistenceConfig {
+    fn default() -> Self {
+        Self {
+            enabled: DEFAULT_SNAPSHOT_PERSISTENCE_ENABLED,
+            batch_rows: DEFAULT_SNAPSHOT_BATCH_ROWS,
+            max_read_rows: DEFAULT_SNAPSHOT_MAX_READ_ROWS,
+            insert_timeout: Duration::from_secs(DEFAULT_SNAPSHOT_INSERT_TIMEOUT_SECS),
+            retention_days: DEFAULT_SNAPSHOT_RETENTION_DAYS,
+        }
+    }
+}
+
+impl SnapshotPersistenceConfig {
+    /// Reads the configuration from the environment.
+    ///
+    /// An unset variable takes its documented default. A set-but-invalid one
+    /// fails, rather than falling back — see the module docs for why.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ChainError::Validation`] naming the environment variable when
+    /// its value does not parse, is zero, or exceeds its documented bound.
+    pub fn from_env() -> Result<Self, ChainError> {
+        let config = Self {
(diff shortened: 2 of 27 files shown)


src/infrastructure/config/snapshot.rs
```diff
@@ -0,0 +1,476 @@
...
+
+/// Default cap on the quote rows one snapshot may carry, in rows.
+///
+/// A snapshot is `expirations × strikes` rows; ADR 0001's reference
+/// configuration is a few hundred, and the per-snapshot contract budget in
+/// `SimulationParametersV2::validate` already bounds what the domain will
+/// price. This is the *storage* bound: a snapshot larger than it is rejected
+/// before a single row is written, so one pathological simulation cannot turn
+/// into an unbounded insert.
+pub const DEFAULT_SNAPSHOT_BATCH_ROWS: usize = 100_000;
```
joaquinbejar: 🟠 **Warning**: the default storage cap is 100,000 rows while OCS_MAX_SNAPSHOT_CONTRACTS defaults to 200,000, so a session accepted by the API can systematically fail every ClickHouse persist with the default configuration. Validate these knobs together at startup or make the storage bound at least the maximum accepted snapshot.

src/session/manager_v2.rs
```diff
@@ -197,6 +242,47 @@ impl SimulationManager {
...
+    /// The trade this accepts: a snapshot filed after the response means a
+    /// client that advances and immediately queries the warehouse may not find
+    /// the step yet. Deterministic replay is the read path that is always
+    /// current; the warehouse is the one that is durable.
+    fn file_snapshot(&self, simulation: &SessionV2, snapshot: &SeriesSnapshot) {
+        let Some(warehouse) = &self.warehouse else {
+            return;
+        };
+
+        let record = snapshot_record(simulation.id, &simulation.parameters.symbol, snapshot);
```
joaquinbejar, line 339: 🟠 **Warning**: snapshot_record walks and clones every quote synchronously before try_send, so filing is not actually off the request clock. The channel is also bounded by 1,024 records rather than weight; legal 200,000 quote snapshots can retain more than 200 million quotes while the warehouse is slow. Move materialization to bounded background work and enforce a total row or byte budget.

src/main.rs
```diff
@@ -147,7 +148,23 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    // it is off the manager never learns the feature exists; when it is on, the
+    // tables are created here rather than on the first advance, so a schema
+    // problem is a startup failure with a message rather than a warning buried
+    // in the serving path.
+    let mut simulation_manager = SimulationManager::new(simulation_store, v2_config);
+    match ClickHouseSnapshotRepository::from_env()? {
+        Some(warehouse) => {
+            warehouse.ensure_schema().await?;
+            info!("v2 snapshot persistence is enabled");
+            simulation_manager = simulation_manager.with_warehouse(Arc::new(warehouse));
```
joaquinbejar, line 166: 🟠 **Warning**: the repository is only handed to SimulationManager for writes. The v2 export path never receives it, so ClickHouse cannot be the preferred source for persisted ranges or backfill gaps as issue #56 requires, and the new read API is unreachable from the product surface. Wire the repository into the export and read flow with deterministic replay as fallback.

src/infrastructure/repositories/snapshot_repo.rs
```diff
@@ -0,0 +1,1616 @@
...
+    /// table that already exists, so changing `OCS_SNAPSHOT_RETENTION_DAYS`
+    /// against a live deployment needs an explicit
+    /// `ALTER TABLE ... MODIFY TTL`; the DDL documents that.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ChainError::ClickHouseError`] when the warehouse rejects the
+    /// DDL or is unreachable.
+    #[instrument(skip(self), level = "debug")]
+    pub async fn ensure_schema(&self) -> Result<(), ChainError> {
```
joaquinbejar, line 314: 🟠 **Warning**: no test in this PR calls ensure_schema or any ClickHouseSnapshotRepository read or write against the live ClickHouse service. The fake writer and pure assembly tests cannot validate the DDL, FINAL syntax, Decimal(38, 28), named parameters, or TTL behavior, leaving the issue's real ClickHouse integration acceptance criterion uncovered.

joaquinbejar (commented):
The storage model and completion protocol are thoughtfully structured, but the default bounds reject valid snapshots, record materialization and queue memory remain on the request path, the repository is not wired into exports or replay backfill, and no live ClickHouse test exercises the new schema and queries. REQUEST_CHANGES.

src/domain/series.rs
```diff
@@ -397,20 +438,38 @@ impl SnapshotCache {
...
+    fn evict_to_contracts(&mut self, max: usize) {
+        while self.contracts() > max {
+            match self.least_recently_used() {
+                Some(key) => {
+                    self.entries.remove(&key);
+                }
+                None => break,
+            }
+        }
+    }
```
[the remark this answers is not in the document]
joaquinbejar: This one is already gone from the branch: the same finding landed on the caps PR below this one, so each entry now carries the weight it was measured at and the cache keeps a running total. The budget check is O(1) and eviction is linear in the victims.

src/infrastructure/repositories/snapshot_repo.rs
joaquinbejar: There are five, in snapshot_repo.rs, and they do exactly that: each one calls ensure_schema and then persists and reads through the repository against a live server, covering the round trip, double-persist idempotence, a torn write, a short batch and a contract history. They carry the repo's usual ignore attribute so the default suite stays hermetic, and CI runs them in the Integration job that is green on this PR. I also ran them locally against the exact CI image after the column rename, since a rename is precisely what the driver's schema validation catches.

src/infrastructure/config/snapshot.rs
joaquinbejar: Fixed at both ends. The storage bound now defaults to the per-snapshot contract cap rather than half of it, and startup refuses a configuration where it is lower, naming both variables — a service that accepts simulations it can never persist is a configuration mistake, not something to discover in the logs.

src/session/manager_v2.rs
joaquinbejar: Both parts fixed. The decision to file now happens before anything is built, from a quote count taken off the domain snapshot, so a degraded warehouse costs a comparison rather than a clone of every quote. And the queue carries a row budget alongside the record depth, since a record is a few hundred quotes here and up to the per-snapshot cap in a large configuration; a new test drives a warehouse whose first write never returns and asserts it stops being queued.

src/session/manager_v2.rs
```diff
@@ -197,6 +242,47 @@ impl SimulationManager {
...
+        if let Err(error) = warehouse.try_send(record) {
+            warn!(
+                simulation_id = %simulation.id,
+                step = snapshot.step,
+                error = %error,
+                "The snapshot queue is full; the step was not filed and can be replayed"
+            );
+        }
+    }
```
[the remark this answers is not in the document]
joaquinbejar: Right, and it is fixed the way you describe: the queue is consulted first, from a quote count read off the domain snapshot, and the record is only materialised once there is room for it. The drop path is now a comparison.

.env.example
```diff
@@ -94,13 +94,62 @@ OCS_MAX_CACHED_TAPES=64
...
+# Turning it ON makes ClickHouse a HARD STARTUP DEPENDENCY: the tables are
+# created at boot, so a schema or connectivity problem fails the boot rather
+# than surfacing later as a warning. Writes themselves are detached from the
+# request, so a warehouse that degrades later cannot delay a response.
+# Values: true/false/1/0/yes/no/on/off.  Default: false
+OCS_SNAPSHOT_PERSISTENCE_ENABLED=false
+
+# Rows per insert batch. Every quote row of one snapshot goes in ONE insert, so
+# this bounds a single statement, not the snapshot.
+# Range: 1 .. 5000000.  Default: 100000
```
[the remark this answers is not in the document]
joaquinbejar: Reworded, and the default changed with it. The bound applies to the snapshot, not just the statement, so a snapshot above it is rejected before a row is written; it now defaults to the per-snapshot contract cap and startup refuses a value below it.

src/main.rs
joaquinbejar: Wired. The export now prefers a persisted step and replays whatever is missing, per step rather than all-or-nothing, reading in windows so neither a round trip per step nor a materialised range is needed. Both sources render through one adapter so a row is byte-identical either way, which is what the new tests pin — including one that persists rows a replay could never produce, so an export that ignored the warehouse fails instead of passing by coincidence. The routes take the repository off the manager rather than being handed a second one.

joaquinbejar:
All of it is in 37c2203, and the branch is rebased on main now that the two below it merged. The export reads persisted steps and replays the gaps, in windows, with both sources rendering through one adapter so a row is byte-identical either way. The storage bound now defaults to the per-snapshot cap and startup refuses a lower one naming both variables. Filing consults the queue before materialising anything, and the queue is bounded by rows as well as records. On the live coverage point, the five ClickHouse tests were already there and do call ensure_schema before persisting and reading against a real server; they carry the repo's ignore attribute so the default suite stays hermetic, and the Integration job runs them. Gate is green with 700 unit tests, 16 contract tests and 3 doctests, plus those five run locally against the exact CI image.

PR #71, opened 2026-08-15
feat(domain): price historical v2 steps at a causal volatility estimate

### LINKED ISSUE #63: Restore a causal volatility estimate for v2 historical walks

joaquinbejar:
## What

A `Historical` v2 walk carries no volatility of its own, and the factor tape priced every step of one at the constant `volatility` the request happened to send. v1 does not: its walk driver estimates the realized volatility over an **expanding window**, so step `i` is priced by what had been observed by step `i` and nothing later. Same series, same seed, two different volatility paths — and the v2 one is not causal, which is what makes it wrong rather than merely different: a backtest at step 3 was priced by the turbulence of step 900.

## How

The issue frames this as port-or-nothing, because upstream's `expanding_window_vols` is private to its driver. It is not. All three of its ingredients are public, and upstream's own comment records that its prefix-sum variance is *"algebraically identical to the two-pass form in `constant_volatility`"*:

| upstream function | path |
|---|---|
| `calculate_log_returns` | `utils/others.rs:250` |
| `constant_volatility` | `volatility/utils.rs:49` |
| `adjust_volatility` | `volatility/utils.rs:605` |

So `expanding_window_volatilities` writes an **indexing policy over upstream mathematics** — the window, the `>= 2` guard, the backfill of points 0 and 1 — and no mathematics of its own. There is no second copy of an upstream kernel in this repo, so nothing new has to be re-synced on an optionstratlib bump, and `CLAUDE.md` needs no new Key Decision. Upstream's own whole-series fallback (`walk_volatility`) composes the same three functions, so this matches its idiom rather than working around it.

Everything read stays inside the horizon, the seeding chain's volatility included — a historical walk replays `prices[..steps]`, and every reduction is taken over that prefix. That is a deliberate divergence from v1, whose fallback reduces the whole embedded series.

## Contract

`Historical` requests behave differently, with no schema change:

- **The request's `volatility` prices nothing.** The series prices itself. Not silently swallowed — the values that did price the chains are each snapshot's `base_volatility`, which the snapshot and export surfaces already report. Documented in the request DTO, the response DTO, `lib.rs` and ADR 0001 §8.1.
- **Two series are now refused that used to be served**: one whose realized volatility is zero or above the `1.0` a chain can be priced at, and a horizon of fewer than three steps, which has no dispersion to estimate from. Issue #63 lists refusing as an accepted answer, and the alternative was pricing from data the horizon could not have seen. Both arrive as a `400` naming `method.prices` at the **first peek**, not at creation, because creation deliberately does not build a tape — recorded in ADR §8.1 and the §11 error table.
- v1 and v2 agree on the volatility path **from step 1 on**, to a documented tolerance. Step 0 differs by construction: v1 never asks its estimator about step 0 and prices its first chain at the request's constant, while v2 serves step 0 as a snapshot.

No `pub` signature, DTO field, OpenAPI shape, stored-session field or dependency changed. IronCondor speaks only `/api/v1/chain`, which is untouched.

## Numbers

| | |
|---|---|
| max v1 ↔ v2 deviation | `3.4e-23` annualised (tolerance pinned at `1e-18`) |
| 10 000-step estimate, release | `3.33 s` |

Composing costs quadratic time where upstream's recurrence is linear, because that recurrence is not reachable from outside it. Only `Historical` pays it, once per tape build — and it was running inline on an actix worker, so the build moved to the blocking pool, where the export path already runs the identical call. It files its result from **inside** the blocking task: `spawn_blocking` cannot be cancelled but awaiting it can, and filing afterwards threw away a completed three-second build whenever a client disconnected.

## Tests

Eleven new, in `src/domain/factors.rs`:

- **no look-ahead** — for every cut of a series, the estimates over the prefix equal the first *k* of the estimates over the whole thing;
- **v1 parity** — drives `walk_steps_par`, the exact driver `generator_optionchain` uses, records the volatility it hands the chain builder, and compares it against the tape;
- backfill of points 0 and 1, the `< 3` guard, the zero-dispersion degenerate case;
- per-step variation and rebuild reproducibility on the tape;
- the four new refusals: short horizon, no dispersion, above `1.0`, and a zero price in the resolved series.

## Review

Audited before opening by `reproducibility-reviewer` and `architect`. Their findings — a non-causal fallback, a panic behind a false `# Errors` section, a zero volatility surfacing as a `500`, an uncancellable-but-discarded build, and five doc claims stronger than the code — are all addressed in the second commit rather than deferred.

## Checklist

- [x] Reproducibility intact, with a tape test — same parameters ⇒ identical rows, and the walk path itself is unchanged
- [x] Stored-session shape untouched; the v1 golden contract suite is green
- [x] DTO contract additive — text only, no schema change
- [x] Errors converge on `ChainError`; no upstream error type in any signature
- [x] utoipa descriptions updated on both the request and the response field
- [x] `make pre-push` clean — 711 + 16 + 3 tests, zero clippy/fmt/doc warnings
- [x] `README.md` regenerated via `make readme`

Also included, and called out because it is not this issue: a small commit fixing three broken intra-doc links that survived the v0.2.0 cleanup, so the gate is warning-free again.

Closes #63

### DIFF
--- a/src/domain/factors.rs
+++ b/src/domain/factors.rs
@@ -83,10 +83,9 @@ use optionstratlib::chains::{
 use optionstratlib::simulation::steps::{Step, Xstep, Ystep};
 use optionstratlib::simulation::{WalkParams, WalkTypeAble};
 use optionstratlib::utils::TimeFrame;
-use optionstratlib::utils::others::calculate_log_returns;
 use optionstratlib::volatility::{adjust_volatility, constant_volatility};
 use positive::Positive;
-use rust_decimal::Decimal;
+use rust_decimal::{Decimal, MathematicalOps};
 use tracing::{debug, instrument};
 
 /// One step of the market path: everything a snapshot needs that is not an
@@ -628,27 +627,65 @@ fn expanding_window_volatilities(
 
 /// The log returns of a price series as decimals.
 ///
-/// `Positive::ln` builds its result without revalidating, so a falling price
-/// yields a negative value wearing a `Positive`. Reading each one back as a
-/// `Decimal` — which is what upstream's own estimator does — recovers the true
-/// sign before any arithmetic touches it.
+/// Computed here rather than through upstream's `calculate_log_returns`, which
+/// divides two `Positive`s: that operator panics on a zero divisor **and on a
+/// ratio it cannot represent**, and it never returns an error for either. Both
+/// operands come straight from a request, and a `Decimal` holds about 29
+/// significant digits, so a series stepping from `1e-28` to `7e28` is a
+/// perfectly legal pair of strictly positive prices whose ratio is not
+/// representable. A request must not be able to abort the process, so the
+/// division is checked and an unrepresentable ratio is a rejection naming the
+/// field.
+///
+/// The log itself is taken on the `Decimal`, which is what recovers the true
+/// sign: `Positive::ln` builds its result without revalidating, so a falling
+/// price would otherwise yield a negative value wearing a `Positive`.
 ///
 /// # Errors
 ///
-/// Returns [`ChainError::Validation`] naming `method.prices` if upstream's
-/// log-return computation ever reports a failure.
-///
-/// It cannot report a **zero price**, and this is a precondition rather than an
-/// error case: the computation divides by the previous price and takes the log
-/// of the ratio, and both of those panic on a zero instead of returning.
-/// [`ensure_historical_series_covers_the_horizon`] is what makes that
-/// unreachable, and it has to run first.
+/// Returns [`ChainError::Validation`] naming `method.prices` when a ratio is
+/// not representable, when a price is zero, or when the log of a ratio is not
+/// representable.
 fn log_returns(prices: &[Positive]) -> Result<Vec<Decimal>, ChainError> {
-    let returns = calculate_log_returns(prices).map_err(|e| ChainError::Validation {
+    let unusable = |reason: String| ChainError::Validation {
         field: "method.prices".to_string(),
-        reason: format!("the historical series has no usable log returns: {e}"),
-    })?;
-    Ok(returns.iter().map(Positive::to_dec).collect())
+        reason,
+    };
+
+    let mut returns = Vec::with_capacity(prices.len().saturating_sub(1));
+    for (index, pair) in prices.windows(2).enumerate() {
+        let [previous, current] = pair else {
+            // `windows(2)` yields pairs; destructuring keeps it indexing-free.
+            continue;
+        };
+
+        let previous = previous.to_dec();
+        let current = current.to_dec();
+        if previous.is_zero() {
+            return Err(unusable(format!(
+                "the price at index {index} is zero, so the series has no return at {}",
+                index + 1
+            )));
+        }
+
+        let ratio = current.checked_div(previous).ok_or_else(|| {
+            unusable(format!(
+                "the price ratio at index {} is not representable ({current} over {previous})",
+                index + 1
+            ))
+        })?;
+
+        let log = ratio.checked_ln().ok_or_else(|| {
+            unusable(format!(
+                "the log return at index {} is not representable (ratio {ratio})",
+                index + 1
+            ))
+        })?;
+
+        returns.push(log);
+    }
+
+    Ok(returns)
 }
 
 /// Rescales a volatility from the series' timeframe to a year.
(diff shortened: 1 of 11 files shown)


joaquinbejar:
Copilot declined this review — "the user who requested the review has reached their quota limit" — so per the repo workflow the internal reviewers stood in: `reproducibility-reviewer` on the walk path and the seed plumbing, `architect` on layering and the contract surface.

They ran twice, once on the original change and once on the fixes, and between them landed ten findings that are all addressed rather than deferred. The two that mattered most were the same defect seen from different angles: **the estimate was causal but the reduction around it was not.** `resolve_base_volatility` reduced the whole embedded series, so a two-step horizon carrying five hundred prices was priced by all five hundred, and a request could be refused over turbulence it would never reach — with advice, *lower the model's volatility*, that does not apply to a walk whose volatility comes from a series. The second pass then caught that the zero-price guard had been left scanning the whole series while everything else moved to the walked prefix, and that the previous commit's message claimed a doc fix it had not made.

The rest: a zero price **panicked** rather than erroring (`Positive`'s division and `Decimal::ln` both panic on zero, and the *resolved* method — the argument that exists so a database-filled series can be passed — carries no validation); a zero volatility surfaced as a `500` from inside upstream's chain builder; a cancelled await discarded a completed 3.3 s build; the zero refusal fires on a flat *opening* rather than only a flat series, which the docs did not say; and five doc claims were stronger than the code, including the parity claim on the crate front page — v1 and v2 differ at step 0 by construction, which the ADR itself already recorded.

Two things the reviewers raised that are **not** in this PR, because they are decisions rather than defects:

- The `volatility` field is now inert for `Historical` and is already required-but-checked for the nine synthetic models, so it carries no information for any variant. Dropping it or making it optional is a DTO change and belongs in its own issue.
- No single-flight in front of a now-quadratic build (concurrent first-requests each build the same tape), and the snapshot build still runs on the request worker while only the tape build moved to the blocking pool.

src/domain/factors.rs
```diff
@@ -191,22 +217,35 @@ impl FactorTape {
...
+        // Where each step's volatility comes from, in v1's order of precedence
+        // (`walk_steps_par`): a stochastic-volatility model reports its own
+        // path, index-aligned with the prices; a historical walk reports none,
+        // so it is estimated causally from the path itself; everything else is
+        // the model's constant.
+        let step_volatilities = match path.vols {
+            Some(ref vols) => Some(vols.clone()),
+            None => match method {
+                SimulationMethod::Historical { timeframe, .. } => {
+                    expanding_window_volatilities(&path.prices, *timeframe)?
```
joaquinbejar, line 228: 🔴 **Critical**: this changes the deterministic volatility and premiums of historical snapshots while CURRENT_SNAPSHOT_GENERATION remains 1. The 0.2.0 image containing generation 1 persistence was already published from main, so a replay or retry after this deploy will address old and new tapes with the same key and may replace or mix their rows. Bump the generation and pin the old versus new behavior in a cross-generation test.

src/domain/factors.rs
```diff
@@ -387,41 +468,315 @@ fn resolve_base_volatility(
...
+/// Returns [`ChainError::Validation`] naming `method.prices` if upstream's
+/// log-return computation ever reports a failure.
+///
+/// It cannot report a **zero price**, and this is a precondition rather than an
+/// error case: the computation divides by the previous price and takes the log
+/// of the ratio, and both of those panic on a zero instead of returning.
+/// [`ensure_historical_series_covers_the_horizon`] is what makes that
+/// unreachable, and it has to run first.
+fn log_returns(prices: &[Positive]) -> Result<Vec<Decimal>, ChainError> {
+    let returns = calculate_log_returns(prices).map_err(|e| ChainError::Validation {
```
joaquinbejar: 🔴 **Critical**: calculate_log_returns uses Positive division, which panics not only on zero but also when current divided by previous overflows Decimal. Both operands can be valid strictly positive request values, for example 1e-28 followed by 7e28, so this new path still has a request-reachable panic. Compute the ratios with checked arithmetic or reject an unrepresentable ratio before calling upstream, and cover both overflow and underflow cases.

src/session/manager_v2.rs
```diff
@@ -462,18 +466,40 @@ impl SimulationManager {
...
-        // behind it.
-        let tape = FactorTape::build(&simulation.parameters, &simulation.parameters.method)?;
-        self.cache_tape(simulation.id, tape.clone());
-        Ok(tape)
+        let parameters = simulation.parameters.clone();
+        let id = simulation.id;
+        let tapes = Arc::clone(&self.tapes);
+        let max_cached_tapes = self.config.max_cached_tapes;
+
+        tokio::task::spawn_blocking(move || {
```
joaquinbejar, line 578: 🟠 **Warning**: every concurrent first peek can miss the cache and start its own quadratic 10,000-step build. The spawn_blocking pool is not a CPU work limiter and can grow to hundreds of threads, so one simulation can exhaust CPU before any build populates the cache. Add a per-simulation single flight or a bounded semaphore and have callers await the shared result.

joaquinbejar (commented):
The causal estimator is well tested and the v1 parity work is convincing, but the change reuses an already published snapshot generation, leaves an extreme value panic reachable from historical prices, and allows unbounded duplicate CPU builds on concurrent first reads. REQUEST_CHANGES.

src/domain/factors.rs
joaquinbejar: Right, and the collapse makes it worse than a mix, since the tables replace on that key rather than keeping both. Generation is 2 now, the constant carries the history of what each one priced, and a test pins that the same step under two generations is two rows.

src/domain/factors.rs
joaquinbejar: Confirmed against upstream, it divides two Positives and never returns an error for either failure, so the doc comment reasoning only about zero was covering half the hole. The ratios are computed here with checked arithmetic now and an unrepresentable one is a 400 naming method.prices, with your 1e-28 to 7e28 case and its mirror both covered, plus one violently volatile series that must still estimate.

src/session/manager_v2.rs
joaquinbejar: Added single flight per simulation, with the ownership decided under the lock so two callers cannot both start one and the rest wait on the owner's result. A test drives eight concurrent peeks of a cold simulation and asserts they agree and leave exactly one cached tape.

joaquinbejar:
All three are fixed in 0b05fb0. The generation is 2, so a historical tape priced the new way never lands on the coordinates a 0.2.0 deployment already wrote. The log returns are computed with checked arithmetic here rather than through upstream's Positive division, which panics on an unrepresentable ratio and never reports it, so an extreme but legal series is now a 400 instead of an abort. And tape builds are single flight, so concurrent first reads share one build instead of each starting their own. Gate is green with 719 unit tests, the 16 contract tests and 3 doctests.

Issue #76, opened 2026-08-28
Upgrade optionstratlib 0.18 to 0.20 and positive 0.5 to 0.6


joaquinbejar:
## Overview

This crate pins `optionstratlib = "0.18"` and `positive = "0.5"`
(`Cargo.toml:54-55`). Upstream has released 0.20.0, which carries the chain
greeks this crate needs to expose (#73) plus two correctness fixes that change
numbers this crate already serves. Getting there crosses 0.19.0, a deliberately
breaking release, so the upgrade is its own piece of work rather than a line in
another issue.

## Why upgrade

**What 0.20.0 gives us**

- `OptionData::greeks_call` / `greeks_put` as `Option<GreeksSnapshot>`, the full
  twelve-greek set per option style, computed at chain build time
  (`optionstratlib/src/chains/optiondata.rs:204-210`, upstream #430). This is
  what #73 needs, and it removes any need for this crate to convert and compute
  per request.
- `Greeks` now applies the `Side` sign in every greek, not only in `delta`
  (upstream #435, #437). Before this, an aggregate over a position with short
  legs returned a correctly signed delta next to gamma, theta, vega and rho
  carrying the long-position sign.
- `vomma` and `veta` were applying `quantity` twice (upstream #429).
- `d1` / `d2` were not carrying `b = r - q`, so analytic greeks diverged from the
  discounted terms whenever the dividend yield was non-zero (upstream #426).
- Shared Black-Scholes kernels are computed once per option instead of per greek
  (upstream #434), which matters because the chain builder is on the request
  path here.

The last three change values this service already returns. The upgrade is a
correctness fix, not only an enabler.

## The breaking surface

0.19.0 moved three in-house crates a breaking release each: `positive` 0.5 to
0.6, `expiration_date` 0.2 to 0.3, `option_type` 0.1 to 0.3. All three appear
throughout the upstream public API, so this crate moves in the same step.

**The one that matters here: the JSON wire format.** `positive` 0.6 serialises
`Positive` as the exact decimal in a **string** (`"42.5"` instead of `42.5`).
Every serialised type carrying a `Positive` changes shape, `OptionData` and
`OptionChainBuildParams` and `Step`/`Xstep`/`Ystep` among them.

This crate is exposed through `SimulationParameters`, whose `initial_price`,
`volatility` and `dividend_yield` are `Positive`
(`src/session/model_v2.rs:229-235`), and which is part of the stored session
[...]

Issue #75, opened 2026-08-27
Include the full greek set in the export datasets


joaquinbejar:
## Overview

The export endpoint streams the `option_chains` dataset with columns
`implied_volatility`, `call_delta`, `put_delta` and `gamma`. Once the full greek
set exists in the responses and in the warehouse, the export must carry it too,
otherwise a downstream backtester reading an exported tape gets a strictly poorer
chain than a client stepping the same simulation live.

This is the column set IronCondor's materialised tape is built from, so it
determines which greeks a backtest can attribute P&L to. Attribution beyond
delta and theta is impossible without vanna, vomma and charm on the tape.

## Context

- Header row and column order: `src/api/rest/export.rs:165-174`
- The exported row struct: `src/api/rest/export.rs:300-309`
- Two conversion paths feed it: from freshly built chain data
  (`export.rs:318-327`) and from a stored warehouse row (`export.rs:336-345`).
  Both must be extended, or the export silently differs depending on whether
  persistence is registered
- JSON serialisation of the same row: `src/api/rest/export.rs:946-955`

The export docstring guarantees byte-identical output on repeat for a given
simulation, so the column order must be fixed and deterministic.

The greek set is the twelve values of `optionstratlib::GreeksSnapshot`, per
option style.

Depends on: #73 and #74.

## Tasks

- [ ] Add a `greeks` query parameter to the export endpoint with the same
      `none` / `first` / `all` values and the same default as #73, so a tape and
      a live step agree on what a level means
- [ ] Append the new greek columns to the CSV header at `export.rs:165-174`,
      after `gamma`, so existing column positions do not shift. Naming follows
      the existing pattern: `call_theta`, `put_theta`, `call_vega`, `put_vega`,
      and so on
- [ ] Emit a fixed column set per level, not a variable one: `greeks=none` and
      `greeks=all` must each produce a stable, documented header, so a consumer
      can parse by position as well as by name
- [ ] Add the matching fields to the exported row struct at `export.rs:300-309`
- [ ] Extend both conversion paths (`export.rs:318-327` and `export.rs:336-345`)
- [ ] Extend the JSON serialisation at `export.rs:946-955`
- [ ] Update the utoipa annotation and the endpoint documentation for the
      `option_chains` dataset

## Acceptance criteria

- [ ] Exporting the same simulation twice at the same greek level still produces
      byte-identical output
- [ ] The CSV export of a simulation with persistence registered is identical to
[...]

Issue #74, opened 2026-08-27
Persist the full greek set in simulation_option_quotes


joaquinbejar:
## Overview

Once the chain responses can carry the full greek set, the persisted snapshot
must carry it too, or a replayed or exported tape silently loses everything past
delta and gamma relative to a live one. `simulation_option_quotes` currently
stores `implied_volatility`, `delta_call`, `delta_put` and `gamma` and nothing
else.

The asymmetry matters concretely: the export endpoint prefers a stored snapshot
over replaying it, so without this change the same simulation step returns
different greek coverage depending on whether persistence was registered.

## Context

- Schema: `src/infrastructure/clickhouse/schema/simulation_option_quotes.sql`
- Insert and select: `src/infrastructure/repositories/snapshot_repo.rs:118-120`
  (insert column list), `:164` (per-side projection tuple), `:185` (gamma
  projection), `:853-855` (row to domain mapping)
- The table is a `ReplacingMergeTree(inserted_at_ms)` ordered by
  `(simulation_id, simulation_generation, step, expires_at, strike)`, so adding
  nullable columns is a non-breaking `ALTER TABLE ... ADD COLUMN`

The greek set is the twelve values of `optionstratlib::GreeksSnapshot`: `delta`,
`gamma`, `theta`, `vega`, `rho`, `rho_d`, `alpha`, `vanna`, `vomma`, `veta`,
`charm`, `color`. Stored per option style, so twenty-four columns total, of which
`delta_call`, `delta_put` and `gamma` already exist.

Depends on: #73.

## Column layout

Follow the existing naming pattern (`delta_call` / `delta_put`), giving
`theta_call`, `theta_put`, `vega_call`, `vega_put`, and so on. Twenty-four
explicit nullable columns rather than one JSON blob per side: the columns stay
queryable, and they compress to almost nothing because a block holds one
simulation and consecutive steps, exactly the low-entropy case the file's CODEC
note already describes.

`gamma` today is a single shared column. Decide explicitly whether to keep it
shared or split it into `gamma_call` / `gamma_put` for consistency with the
upstream per-style snapshots. Splitting is the consistent choice; keeping it
shared avoids a migration on an existing column. Either way, document the
decision in the schema file.

## Tasks

- [ ] Add the missing greek columns as
      `Nullable(Decimal(38, 28)) CODEC(ZSTD(1))` to
      `simulation_option_quotes.sql`, matching the existing column style
[...]

PR #79, opened 2026-08-28
Emit the full greek set in the v1 and v2 chain responses (#73)

### LINKED ISSUE #73: Emit the full greek set in the v1 and v2 chain responses
## Overview

The chain responses expose implied volatility, gamma and per-side delta, and
nothing else (`src/api/rest/responses.rs:69-95`,
`src/api/rest/responses_v2.rs:177-196`). A client holding a position against a
simulated chain cannot show theta, vega or rho, and has no access at all to the
second-order greeks that determine how those first-order numbers will move.

Emit the full greek set, per option style, on both v1 and v2. Because the full
set is twelve values per style and a chain can be large, gate it behind a query
parameter so the default payload does not grow for clients that do not need it.

This is the enabling change for a live strategy viewer: pick a strategy, step the
simulation, and watch P&L move next to the position's greeks. Without it that
screen requires a second, divergent pricing implementation in the client.

### Today's surface

- v1: `OptionContractResponse` carries `implied_volatility` and `gamma`;
  `OptionPriceResponse` carries `bid`, `ask`, `mid`, `delta`
  (`src/api/rest/responses.rs:69-95`). The conversion is
  `src/api/rest/handlers.rs:43-60`, reading `contract.current_gamma()`.
- v2: `ContractResponse` carries `implied_volatility` and `gamma`;
  `OptionQuoteResponse` carries `delta` (`src/api/rest/responses_v2.rs:177-196`).
  The conversion is `src/api/rest/responses_v2.rs:244-255`.

### The greek set

Twelve values, all already computed upstream:

- first order: `delta` (∂V/∂S), `theta` (∂V/∂t), `vega` (∂V/∂σ), `rho` (∂V/∂r),
  `rho_d` (∂V/∂q)
- second order: `gamma` (∂²V/∂S²), `vanna` (∂²V/∂S∂σ), `vomma` (∂²V/∂σ²),
  `charm` (∂²V/∂S∂t), `veta` (∂²V/∂σ∂t)
- third order: `color` (∂³V/∂S²∂t)
- ratio: `alpha`

### Where they come from

Upstream 0.20.0 stores the whole set on the chain data:
`OptionData::greeks_call` and `OptionData::greeks_put`, both
`Option<GreeksSnapshot>`
(`optionstratlib/src/chains/optiondata.rs:204-210`). They are computed once at
chain build time by `calculate_greeks()` (`:1071`), so this crate reads them and
does not compute anything on the request path.

`GreeksSnapshot` (`optionstratlib/src/greeks/equations.rs`) derives `Serialize`,
`Deserialize` and `ToSchema`, so it crosses this boundary unchanged and needs no
local mirror type.

Two properties of the upstream snapshots to carry into the API documentation:

- `greeks_snapshot` builds via `get_option(Side::Long, style)`
  (`optionstratlib/src/chains/optiondata.rs:1089`), so every value is **per one
  long contract**. The client applies position sign and size. Upstream #435 made
  the `Greeks` trait apply the `Side` sign in every greek, so a consumer that
  also applies it would double-count.
- `alpha` is normalised to `None` when it would be `Decimal::MAX`
  (`optiondata.rs:1102`), so an absent alpha is a real absence, not an error.

`delta_call`, `delta_put` and `gamma` remain on `OptionData` as documented
convenience mirrors of the snapshots (`optiondata.rs:122-140`), computed
independently because `delta` and `gamma` return values at expiry and at zero
volatility where the full set does not. Keep reading the mirrors for the existing
fields so the default response is unchanged.

Note that `set_extra_params` and `set_volatility` **drop** the snapshots
(`optiondata.rs:455-456`) rather than leaving them stale. If this crate mutates
pricing inputs after a chain is built, it must call `calculate_greeks()` again or
the greeks come back `None`.

### Depends on

#76 moves the workspace pin from `optionstratlib` 0.18 to 0.20. That crosses
0.19.0, a breaking release that changes the JSON wire format for `Positive`, so
it is tracked separately. This issue lands on top of it.

### Payload size, and why the parameter is not optional

Twelve greeks on two styles is 24 numbers per strike on top of the existing
quote. The documented worst-case chain in `src/session/model_v2.rs:426-437` is
1001 strikes by 512 expirations. Even at one expiration, a full-greek snapshot of
1001 strikes is on the order of 24 000 extra numbers per response. A play-loop UI
stepping every 700 ms would be pulling that continuously. The set must therefore
be opt-in, and the default must stay what clients get today.

## Tasks

- [ ] Read `data.greeks_call` and `data.greeks_put` in the response conversions;
      no local greek computation in this crate
- [ ] Document in the OpenAPI description that every emitted greek is **per one
      long contract**: the client applies position sign and size
- [ ] Add a `greeks` query parameter to the v2 snapshot and step endpoints and to
      the v1 chain endpoints, with values `none`, `first`, `all`
  - `none`: current behaviour, the default. `implied_volatility`, `gamma` and
    per-side `delta` only, so no existing client sees a payload change
  - `first`: adds `theta`, `vega`, `rho`, `rho_d`
  - `all`: the full `GreeksSnapshot` per style
- [ ] v2: add an optional `greeks: Option<GreeksSnapshot>` to
      `OptionQuoteResponse` (per style), populated according to the parameter;
      map it in the conversion at `src/api/rest/responses_v2.rs:244-255`
- [ ] v1: the same on `OptionPriceResponse`, mapped in
      `src/api/rest/handlers.rs:43-60`
- [ ] Keep every new field optional and additive; a client that ignores unknown
      fields must be unaffected
- [ ] Reject an unknown `greeks` value with a typed 400 rather than silently
      falling back to the default
- [ ] Update the utoipa annotations for all four DTOs and for the new parameter
      (no `actix_extras`)
- [ ] Update the `Default` impls in `src/api/rest/responses.rs`

## Acceptance criteria

- [ ] With `greeks=all`, a v2 snapshot response carries all twelve greeks for
      both the call and the put on every non-degenerate strike; a test asserts it
- [ ] With `greeks=all`, a v1 chain response carries the same; a test asserts it
- [ ] With no parameter, the response is byte-identical to the response before
      this change; this is the regression test that protects existing clients
- [ ] `charm` differs between call and put and `rho` has opposite signs between
      them, so the per-style split is real and not a copied value
- [ ] An unknown `greeks` value returns 400 with a typed error body
- [ ] The OpenAPI document lists the parameter, its three values, and the new
      fields
- [ ] Response size is measured for a representative chain at each of the three
      levels and recorded in the PR
      snapshot tape
- [ ] `LOGLEVEL=WARN cargo test --workspace` clean (MongoDB on localhost:27017)
- [ ] `cargo build --release` clean (zero warnings)

## Consumer coordination

IronCondor consumes these DTOs. Its `RawQuote` already has `theta` and `vega`
fields filled with placeholder zeros for simulator-sourced tapes; once this
ships, `src/data/simulator.rs` there can carry real values and drop the
placeholder. Additive fields mean no coordinated release is required, but flag
the IronCondor issue when this merges.

joaquinbejar:
## What

Both chain-serving endpoints on each version — `GET /api/v1/chain`,
`POST /api/v1/chain/step`, `GET /api/v2/simulations/{id}/snapshot` and
`POST /api/v2/simulations/{id}/step` — accept an optional `greeks` level that
adds the rest of the greek set to each quoted side.

Closes #73.

| `greeks` | The `greeks` key on each quoted side |
|---|---|
| absent, or `none` | Not present at all. The response is what clients get today |
| `first` | `theta`, `vega`, `rho`, `rho_d` |
| `all` | The full twelve-value `GreeksSnapshot`: `delta`, `gamma`, `theta`, `vega`, `rho`, `rho_d`, `alpha`, `vanna`, `vomma`, `veta`, `charm`, `color` |

This is the enabling change for a live strategy viewer: pick a strategy, step
the simulation, and watch P&L move next to the position's greeks — without a
second, divergent pricing implementation in the client.

## Why it is opt-in

Twelve values on two styles is 24 numbers per strike on top of the existing
quote, and a snapshot can be 1 001 strikes wide across 512 expirations. A
play-loop UI stepping every 700 ms would pull that continuously.

Measured on a 29-strike 0DTE snapshot (debug build; the byte counts are
build-independent, the timings are not):

| Level | Body | Ratio | Marginal per contract |
|---|---|---|---|
| absent / `none` | 7 754 B | 1.00x | — |
| `first` | 17 819 B | 2.30x | +347 B |
| `all` | 36 926 B | 4.76x | +1 006 B |

`first` and `all` cost the same to compute — upstream builds the snapshot whole
either way — and differ only in what is written out.

## Rejection, not fallback

Any other value is a `400` naming `greeks`. A client that asked for `all` and
quietly received the default would price a position against greeks it never
got. On `POST /step` the level is resolved **before** the cursor moves, so a
rejected level cannot consume a step — a 400 that advanced the simulation would
make the error unrepeatable.

## Where the values come from, and where they run

`OptionData::greeks_call` / `greeks_put`, computed by upstream's
`calculate_greeks`. **No greek mathematics is added to this crate.**

They are computed lazily in `greeks_for` (`src/api/rest/greeks.rs`), on a clone
— `calculate_greeks` needs `&mut`, and the DTO layer has no business mutating
what it was handed to render. Upstream's `with_greek_snapshots` stays off:
turning it on globally would charge every client for a payload most never ask
for.

**The pricing runs on the blocking pool.** At roughly 40 µs per contract and a
`DEFAULT_MAX_SNAPSHOT_CONTRACTS` of 200 000, a single `?greeks=all` call can be
seconds of uninterrupted CPU; left on a tokio worker it would stall every other
request that worker holds. Both versions render above the default level inside
`spawn_blocking`, for the same reason and in the same way as the v2 tape build
(`session/manager_v2.rs`) and the export path. The default level stays inline —
it is a field-by-field copy of numbers the snapshot already holds.

**The MongoDB event log is written at the default level** regardless of what the
caller asked for. It records the chain at step N, not what the client who
happened to advance it wanted; without that, two clients advancing the same
session would write differently-shaped documents, and a `greeks=all` advance
would write roughly five times the BSON for no gain.

## The shape on the wire

`first` cannot be a partial `GreeksSnapshot` — nine of its twelve fields are
non-optional `Decimal` — so `GreeksResponse` is an untagged enum over the
upstream snapshot and a four-field first-order projection. Untagged keeps the
key a plain object at both levels.

`FirstOrderGreeks` carries `deny_unknown_fields`, which does two jobs at once:
it makes serde's untagged resolution independent of variant order, and utoipa
derives `additionalProperties: false` from it. That is what keeps the published
`oneOf` **satisfiable**. Without it a twelve-value payload would match both
branches, so every `greeks=all` response would be invalid against the document
this service publishes: strict validators reject it, and the `oneOf`
deserialisers openapi-generator emits for Java, C# and Python raise "multiple
matches found". A test now pins both the branch order and the closed shape.

Three things a consumer has to know, all of them in the OpenAPI description and
the crate docs:

- **Every value is per ONE LONG CONTRACT.** The client applies position sign
  and size exactly once. Upstream builds the snapshot as a long position and,
  since 0.20, applies the side sign inside *every* greek — so a consumer that
  applies it again double-counts.
- **They are decimals, so they arrive as JSON strings**, not numbers. That is
  the deliberate consequence of carrying the upstream type verbatim rather than
  an `f64` mirror that could drift from what upstream computed. Everything else
  on the wire stays `f64`.
- **`null` means not meaningful for these inputs**, never zero. Only `rho`,
  `rho_d` and `alpha` can be null; upstream normalises `alpha` to `null` where
  it would otherwise be `Decimal::MAX`. Distinct from that: a strike whose
  option cannot be built carries **no `greeks` key at all**, which on the wire
  looks like the default level. Its `implied_volatility`, `gamma` and `delta`
  are still there.

`delta` and `gamma` keep their existing places on the quote and the contract.
They are upstream's convenience mirrors, computed independently of the
snapshots and defined at expiry and at zero volatility where the full set is
not — so the default response is unchanged at **every** strike, degenerate ones
included.

## Tests

Thirty-two new tests. The load-bearing ones:

- the default response is identical to `greeks=none` and carries no `greeks`
  key at all, on v1 and v2 — the regression that protects existing clients
- `greeks=all` carries exactly the twelve documented values per style;
  `greeks=first` exactly the four the default lacks
- `charm` differs between call and put and `rho` carries opposite signs, while
  the six style-independent greeks agree exactly — so the split is real rather
  than one value copied twice
- the snapshot's `delta` and `gamma` agree with the `f64` mirrors the response
  duplicates, so a client never has to decide which to believe
- prices, delta and gamma are the same numbers at all three levels, and two
  simulations sharing a seed serve identical `greeks=all` snapshots
- an unknown level is a typed 400 naming the field on **both** versions, and on
  either `/step` it does not advance the cursor (the v1 half needs MongoDB, so
  it is `#[ignore]`d and runs in CI's integration job)
- `greeks_for` covers both branches, including the pre-populated one no handler
  reaches today
- the OpenAPI document advertises the parameter on all four chain-serving
  operations, names its three values and the per-one-long-contract convention,
  publishes the three schemas, and keeps the two `oneOf` branches mutually
  exclusive

Reproducibility is untouched: the greeks are read off the option and never fed
back into the walk, pinned both by the "same numbers at all three levels" test
and by the same-seed comparison at `greeks=all`.

### Documented-surface changes

`400` joins the documented codes on both v1 chain endpoints. That completes the
document rather than changing behaviour: a malformed `sessionid` has always
mapped to `400` through `ChainError::InvalidState`, and the frozen-code test now
records it. Every previously frozen code is still there. The entry deliberately
publishes **no** body schema: these operations answer 400 with two different
shapes — the typed `{error, field}` for a rejected level, `{error}` alone for a
malformed id — and promising one would document a `field` that is sometimes
absent. The description says which is which.

**ADR 0001 §7 is amended.** It stated that all v2 numeric fields are `f64` at
the REST boundary; the greeks are the documented exception, and the contract of
record now says so.

`utoipa` now declares `features = ["decimal"]` first-hand. The greek DTOs derive
`ToSchema` over `Decimal`, so that schema is a requirement of this crate rather
than a side effect of what optionstratlib happens to enable.

### Semver

`OptionPriceResponse` is publicly re-exported with all-public fields, so adding
`greeks` breaks any downstream struct literal — the edit to
`tests/v1_contract.rs` in this diff is the proof. The next release is **0.3.0**,
which #76 already established for the `positive` 0.6 public-dependency bump.
Worth considering separately: `#[non_exhaustive]` on the quote DTOs, so this is
the last time an additive field breaks construction.

## Consumer coordination

IronCondor's `RawQuote` already has `theta` and `vega` fields filled with
placeholder zeros for simulator-sourced tapes. Once this ships,
`src/data/simulator.rs` there can request `greeks=first` and carry real values.
The fields are additive and skipped when absent, so no coordinated release is
needed — but note the decimal-as-string rendering when wiring the parser.

## Reviewed before opening

`architect` and `api-expert` audited the diff. Their findings are folded in
rather than noted: the blocking-pool move, the unsatisfiable `oneOf`, the
duplicate `400` that silently dropped its own description, the over-promised
400 body schema, the client-shaped event log, the missing v1 HTTP coverage, the
unreachable `greeks_for` branch, the undeclared `utoipa/decimal` feature, the
ADR amendment, and several doc claims that were measurably wrong (the greek
build cost is 1.5x a chain build, not seven; the clone is needed for the `&mut`,
not for cache sharing).

## Checks

- `LOGLEVEL=WARN cargo test --workspace` clean — 760 lib tests, 16 contract
  tests, 3 doctests, 0 failures
- `cargo test --workspace -- --ignored` clean — 19 live-service tests against
  Redis, ClickHouse and MongoDB containers

## Stack

- **Base branch: `main`.** #77 merged, so this branch was rebased onto the
  squashed commit and the diff now reads cleanly against `main`.
- **Stacked on: #77 (merged).**
- **Stack: #76 → #73 → #74 → #75.** This one is #73; #74 stacks on it next.
- **Blocks: #85**, which stacks on this one.

### DIFF
--- a/src/api/rest/greeks.rs
+++ b/src/api/rest/greeks.rs
@@ -1,0 +1,415 @@
+//! The `greeks` query parameter and the payload it selects.
+//!
+//! A chain response carries implied volatility, gamma and per-side delta by
+//! default and nothing else. The full set is twelve values per option style,
+//! and a chain can be 1 001 strikes wide across 512 expirations, so emitting it
+//! unconditionally would multiply the payload a play-loop client pulls every
+//! few hundred milliseconds. It is therefore opt-in, and the default is exactly
+//! what clients get today.
+//!
+//! # Levels
+//!
+//! | `greeks` | What the quote carries |
+//! |---|---|
+//! | absent or `none` | Today's fields only. No `greeks` key at all |
+//! | `first` | `theta`, `vega`, `rho`, `rho_d` |
+//! | `all` | The full twelve-value [`GreeksSnapshot`] |
+//!
+//! # Sign and size
+//!
+//! Every emitted greek is **per one long contract**. Upstream builds the
+//! snapshot through `get_option(Side::Long, style)`, and since optionstratlib
+//! 0.20 the `Greeks` trait applies the `Side` sign in *every* greek rather than
+//! only in `delta` — so a consumer that applies the position sign again would
+//! double-count it. The client applies position sign and size exactly once, to
+//! these values.
+//!
+//! An absent `rho`, `rho_d` or `alpha` means **not meaningful for these
+//! inputs**, never zero; upstream normalises `alpha` to `None` where it would
+//! otherwise be `Decimal::MAX`.
+
+use crate::utils::ChainError;
+use optionstratlib::chains::OptionData;
+use optionstratlib::greeks::GreeksSnapshot;
+use rust_decimal::Decimal;
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+/// How much of the greek set a chain response should carry.
+///
+/// Parsed from the `greeks` query parameter by [`GreekLevel::parse`]; the
+/// default is [`GreekLevel::None`], which is the pre-existing response.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum GreekLevel {
+    /// Implied volatility, gamma and per-side delta only: the default, and
+    /// byte-identical to the response before the parameter existed.
+    None,
+    /// Adds the remaining first-order greeks: `theta`, `vega`, `rho`, `rho_d`.
+    First,
+    /// The full twelve-value snapshot per option style.
+    All,
+}
+
+impl GreekLevel {
+    /// Parses the raw `greeks` query value.
+    ///
+    /// An absent parameter is [`GreekLevel::None`], so a client that has never
+    /// heard of it keeps its current payload.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ChainError::Validation`] naming `greeks` for any other value.
+    /// An unrecognised level is rejected rather than silently downgraded: a
+    /// client that asked for `all` and quietly received the default would
+    /// price a position against greeks it never got.
+    pub(crate) fn parse(raw: Option<&str>) -> Result<Self, ChainError> {
+        match raw {
+            None => Ok(Self::None),
+            Some(value) => match value.trim() {
+                "none" => Ok(Self::None),
+                "first" => Ok(Self::First),
+                "all" => Ok(Self::All),
+                other => Err(ChainError::Validation {
+                    field: "greeks".to_string(),
+                    reason: format!("unknown greek level '{other}'; expected none, first or all"),
+                }),
+            },
+        }
+    }
+
+    /// Whether this level needs the option's greek snapshots at all.
+    #[must_use]
+    #[inline]
+    pub(crate) fn wants_greeks(self) -> bool {
+        !matches!(self, Self::None)
+    }
+}
+
+/// The first-order greeks the default response does not already carry.
+///
+/// `delta` is deliberately absent: it is already on the quote, and repeating it
+/// under a second key would let the two drift. `gamma` is likewise already on
+/// the contract.
+///
+/// Closed with `deny_unknown_fields`, which does two jobs at once: it makes
+/// serde's untagged resolution independent of variant order, and utoipa derives
+/// `additionalProperties: false` from it. That is what keeps the published
+/// `oneOf` satisfiable — without it a full twelve-value payload would satisfy
+/// this four-field shape as well, and "exactly one branch" would fail for
+/// every `greeks=all` response.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
+#[serde(deny_unknown_fields)]
+pub struct FirstOrderGreeks {
+    /// Sensitivity to the passage of time, per one long contract.
+    pub theta: Decimal,
+    /// Sensitivity to implied volatility, per one long contract.
+    pub vega: Decimal,
+    /// Sensitivity to the risk-free rate. `null` means not meaningful for
+    /// these inputs, never zero.
+    pub rho: Option<Decimal>,
+    /// Sensitivity to the dividend yield. `null` means not meaningful for
+    /// these inputs, never zero.
+    pub rho_d: Option<Decimal>,
+}
+
+impl From<&GreeksSnapshot> for FirstOrderGreeks {
+    fn from(snapshot: &GreeksSnapshot) -> Self {
+        Self {
+            theta: snapshot.theta,
+            vega: snapshot.vega,
+            rho: snapshot.rho,
+            rho_d: snapshot.rho_d,
+        }
+    }
+}
+
+/// The greek payload one quoted side carries, shaped by the requested level.
+///
+/// Serialised untagged, so the `greeks` key is a plain object at both levels
+/// and a client parses it by the fields it finds rather than by a discriminant.
+/// The full variant is the upstream [`GreeksSnapshot`] itself rather than a
+/// local mirror, so a value cannot drift between what upstream computed and
+/// what this service serves.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
+#[serde(untagged)]
+pub enum GreeksResponse {
+    /// `greeks=all`: the full twelve-value snapshot.
+    ///
+    /// Listed first so deserialisation prefers it — an untagged enum takes the
+    /// first variant that matches, and the four-field variant would otherwise
+    /// swallow a complete snapshot by ignoring the rest.
+    Full(GreeksSnapshot),
+    /// `greeks=first`: the remaining first-order greeks.
+    FirstOrder(FirstOrderGreeks),
+}
+
+impl GreeksResponse {
+    /// Shapes an upstream snapshot to the requested level.
+    ///
+    /// Returns `None` at [`GreekLevel::None`], which is what keeps the default
+    /// response free of the key entirely.
+    #[must_use]
+    fn from_snapshot(snapshot: &GreeksSnapshot, level: GreekLevel) -> Option<Self> {
+        match level {
+            GreekLevel::None => None,
+            GreekLevel::First => Some(Self::FirstOrder(FirstOrderGreeks::from(snapshot))),
+            GreekLevel::All => Some(Self::Full(snapshot.clone())),
+        }
+    }
+}
+
+/// The greek payloads for both sides of one strike, at the requested level.
+///
+/// Upstream can compute the snapshots at chain build time, but only when the
+/// build asks for them, and nothing in this crate does: `with_greek_snapshots`
+/// is off by default, and turning it on globally would charge every client for
+/// a payload most never request. So in practice **this function computes
+/// them**, through upstream's own `calculate_greeks`; the `is_some` branch is
+/// the guard for a chain that already carries them, not the normal path. No
+/// greek mathematics lives in this crate either way.
+///
+/// # An absent payload is not a zero
+///
+/// Upstream returns no snapshot for a strike whose option cannot be built, and
+/// only logs it at `debug`. Such a strike arrives with **no `greeks` key at
+/// all**, which on the wire is indistinguishable from the default level. A
+/// client that asked for a level and found the key missing on some strikes is
+/// looking at degenerate strikes, not at a downgraded response — the existing
+/// `implied_volatility`, `gamma` and `delta` mirrors are still there, because
+/// they are defined where the full set is not.
+///
+/// # Cost
+///
+/// Roughly 40 µs per contract in a release build. `first` and `all` cost the
+/// same: upstream builds the snapshot whole and the level only decides what is
+/// written out. Computing here rather than at build time is measurably more
+/// expensive per contract, because `calculate_greeks` recomputes delta and
+/// gamma that this response then reads from the mirrors instead; asking the
+/// builder for snapshots costs about 1.5x a plain chain build. The trade is
+/// deliberate — every client would pay the build-time cost, and only some ask
+/// for the payload.
+///
+/// Callers must keep this off the async runtime. Both versions render above the
+/// default level inside `spawn_blocking`, because
+/// `DEFAULT_MAX_SNAPSHOT_CONTRACTS` is 200 000 and seconds of uninterrupted CPU
+/// on a worker would stall every other request that worker holds.
+///
+/// # Why a clone
+///
+/// `calculate_greeks` takes `&mut self` and this function only has `&OptionData`
+/// — the DTO layer has no business mutating the snapshot it was handed to
+/// render. The clone is one option, not the chain, and is unmeasurable next to
+/// the pricing itself.
+#[must_use]
+pub(crate) fn greeks_for(
+    data: &OptionData,
+    level: GreekLevel,
+) -> (Option<GreeksResponse>, Option<GreeksResponse>) {
+    if !level.wants_greeks() {
+        return (None, None);
+    }
+
+    if data.greeks_call.is_some() || data.greeks_put.is_some() {
+        return (
+            data.greeks_call
+                .as_ref()
+                .and_then(|snapshot| GreeksResponse::from_snapshot(snapshot, level)),
+            data.greeks_put
+                .as_ref()
+                .and_then(|snapshot| GreeksResponse::from_snapshot(snapshot, level)),
+        );
+    }
+
+    let mut priced = data.clone();
+    priced.calculate_greeks();
+    (
+        priced
+            .greeks_call
+            .as_ref()
+            .and_then(|snapshot| GreeksResponse::from_snapshot(snapshot, level)),
+        priced
+            .greeks_put
+            .as_ref()
+            .and_then(|snapshot| GreeksResponse::from_snapshot(snapshot, level)),
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use optionstratlib::ExpirationDate;
+    use optionstratlib::chains::chain::OptionChain;
+    use optionstratlib::chains::{OptionChainBuildParams, utils::OptionDataPriceParams};
+    use positive::{Positive, pos_or_panic};
+    use rust_decimal_macros::dec;
+
+    /// A one-strike chain, optionally built with the greek snapshots already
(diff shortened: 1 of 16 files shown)


src/api/rest/greeks.rs
```diff
@@ -0,0 +1,415 @@
...
+//!
+//! | `greeks` | What the quote carries |
+//! |---|---|
+//! | absent or `none` | Today's fields only. No `greeks` key at all |
+//! | `first` | `theta`, `vega`, `rho`, `rho_d` |
+//! | `all` | The full twelve-value [`GreeksSnapshot`] |
+//!
+//! # Sign and size
+//!
+//! Every emitted greek is **per one long contract**. Upstream builds the
```
joaquinbejar: 🟠 **Warning**: this contract is wrong for `alpha`. Upstream defines it as `gamma / theta`, so side and quantity cancel; signing or scaling it gives clients the wrong position value. Exclude `alpha` from this rule and correct the repeated endpoint and crate documentation.

src/api/rest/greeks.rs
```diff
@@ -0,0 +1,415 @@
...
+/// what this service serves.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
+#[serde(untagged)]
+pub enum GreeksResponse {
+    /// `greeks=all`: the full twelve-value snapshot.
+    ///
+    /// Listed first so deserialisation prefers it — an untagged enum takes the
+    /// first variant that matches, and the four-field variant would otherwise
+    /// swallow a complete snapshot by ignoring the rest.
+    Full(GreeksSnapshot),
```
joaquinbejar: 🟠 **Warning**: carrying upstream `GreeksSnapshot` directly makes its `Decimal` serialization part of the public Rust and JSON contract, contrary to the binding `CLAUDE.md` rule that REST DTOs are local `f64` types with one boundary conversion. Use a local twelve-field DTO, or explicitly amend `CLAUDE.md` and the compatibility policy if this exception is intentional; the ADR alone leaves the authoritative rule stale.

src/api/rest/handlers_v2.rs
```diff
@@ -46,6 +48,49 @@ pub(crate) struct AdvanceQuery {
...
+async fn render_snapshot(
+    simulation: SessionV2,
+    snapshot: SeriesSnapshot,
+    level: GreekLevel,
+) -> Result<SnapshotResponse, ChainError> {
+    if !level.wants_greeks() {
+        return Ok(snapshot_response(&simulation, &snapshot, level));
+    }
+
+    tokio::task::spawn_blocking(move || snapshot_response(&simulation, &snapshot, level))
```
joaquinbejar: 🟠 **Warning**: this starts an unconstrained, non-cancellable CPU job for every `first` or `all` request. At the documented 200,000-contract cap, repeated peeks can fan out multi-second work and retain several huge snapshots. Add shared bounded admission or a bounded CPU pool, and apply the same protection to the v1 renderer.

src/api/rest/handlers_v2.rs
```diff
@@ -140,18 +188,25 @@ pub(crate) async fn peek_snapshot(
...
+        Ok(level) => level,
+        Err(error) => return map_error(error),
+    };
 
     match manager.peek(id).await {
-        Ok((simulation, snapshot)) => {
-            HttpResponse::Ok().json(snapshot_response(&simulation, &snapshot))
-        }
+        Ok((simulation, snapshot)) => match render_snapshot(simulation, snapshot, level).await {
+            Ok(response) => HttpResponse::Ok().json(response),
```
joaquinbejar: 🟠 **Warning**: `HttpResponse::json` calls `serde_json::to_string` synchronously after the pricing task returns. The measured `greeks=all` size scales to roughly 250 MB at the configured cap, so serialization can still stall the Actix worker. Serialize inside the bounded CPU pipeline and return JSON bytes, or stream the body; the step endpoint and v1 have the same issue.

src/api/rest/handlers_v2.rs
```diff
@@ -46,6 +48,49 @@ pub(crate) struct AdvanceQuery {
...
+    /// How much of the greek set the snapshot should carry: `none` (the
+    /// default), `first` or `all`. Kept as a raw string so an unknown value is
+    /// a typed `400` naming the field rather than actix's untyped query
+    /// rejection.
+    #[serde(default)]
+    pub(crate) greeks: Option<String>,
+}
+
+/// The query of a snapshot peek.
+#[derive(Debug, Deserialize)]
```
joaquinbejar, line 61: 🟠 **Warning**: this query DTO accepts unknown keys, so `?greek=all` is silently interpreted as no Greek request and returns the default payload, exactly the downgrade this feature is meant to prevent. Add `deny_unknown_fields` across the v1 and v2 query DTOs, map extraction failures to the typed 400, and use a schema enum in the four manual OpenAPI parameters instead of an unconstrained string.

joaquinbejar (commented):
The Greek values and regression coverage are solid, but the public alpha rule is wrong, the REST boundary is no longer controlled locally, large response work is unbounded and partly left on the Actix worker, and query validation does not match the rejection contract. These need fixing before merge, so REQUEST_CHANGES. GitHub does not let the PR author account set that event, so this review is posted as COMMENT. The diff is well over 500 lines, and splitting contract work from request execution would make the follow-up easier to verify.

src/api/rest/greeks.rs
joaquinbejar: Good catch, and upstream says it outright in the docs for alpha, that a short position negates both gamma and theta so the ratio is unchanged. I carved alpha out of the sign and size rule in the module docs, the four endpoint descriptions, the field doc and the crate docs, so nobody scales it by quantity.

src/api/rest/greeks.rs
joaquinbejar: Agreed, and I went with the local DTO rather than amending the rule. FullGreeks and FirstOrderGreeks are now this crate own f64 types converted once at the boundary, and the From impl destructures the upstream struct so a thirteenth greek breaks the build here instead of quietly disappearing from the API.

src/api/rest/handlers_v2.rs
joaquinbejar: Added a shared admission bound, OCS_MAX_CONCURRENT_GREEK_RENDERS with a default of 4, covering v1 and v2 together since they compete for the same cores. Requests above it wait on an async semaphore rather than being rejected, so a client that disconnects while queued drops its future and never takes a thread, which a bare spawn_blocking cannot do once its task has started.

src/api/rest/handlers_v2.rs
joaquinbejar: Right, the encoding was still landing on the worker after the pricing task returned. Pricing and serialisation now happen in the same admitted job and the handlers write bytes, on both step endpoints and on v1. The v1 renderer returns the served bytes plus the DTO to persist, because the event log is still written at the default level.

src/api/rest/handlers_v2.rs
joaquinbejar: Fixed across all five v1 and v2 query DTOs, with a QueryConfig error handler so the rejection comes back as the documented error and field shape instead of actix plaintext, and the four OpenAPI parameters now type greeks as the GreekLevel enum. The v1 request body keeps its unknown field tolerance since v1_contract freezes that deliberately, so this is the query string only.

joaquinbejar:
All five are fixed in add2f29. The alpha rule was the one I had plainly wrong, and upstream states it outright, so it is carved out everywhere the sign and size claim appears. The DTOs are local f64 types again rather than the upstream struct, which also drops the utoipa decimal feature and puts the greeks back on the wire as numbers. Rendering is bounded by a shared semaphore across v1 and v2 with serialisation moved inside the same job, and every query DTO now rejects unknown keys with the typed 400 while the parameter publishes as an enum. The v1 request body keeps its unknown field tolerance, since v1_contract freezes that on purpose and it is a different surface from the query string. Gate is green at 764 lib tests plus the 16 contract tests, clippy and release clean.

PR #85, opened 2026-08-28
Persist the full greek set in simulation_option_quotes (#74)

### LINKED ISSUE #74: Persist the full greek set in simulation_option_quotes

joaquinbejar:
## What

`simulation_option_quotes` stored `implied_volatility`, `delta_call`,
`delta_put` and `gamma` and nothing else, so a tape filed today and replayed
tomorrow lost everything past delta and gamma. Twenty-two new
`Nullable(Decimal(38, 28)) CODEC(ZSTD(1))` columns complete
`optionstratlib`'s twelve-value `GreeksSnapshot` per option style.

Closes #74.

Nothing serves these columns yet — the export still emits `call_delta`,
`put_delta` and `gamma` from both its branches. That is #75, the top of this
stack. This PR makes the tape ready for it.

## A panic closed on the way

`vomma` is `vega * d1 * d2 / iv`, so as the volatility goes to zero that
quotient leaves `Decimal`'s range and `Positive` **panics** inside
`OptionChain::build_chain`. Nothing stopped a request getting there: the model's
volatility was only checked for zero and for exceeding one, so
`"volatility": 1e-13` passed every gate and then aborted the request on a tokio
worker.

It fires only when the full greek set is computed, so it has been reachable
through `?greeks=` since #79 — and this change would have put it on the
**default** path of every warehouse-backed deployment.

`reject_unpriceable_volatility` now has a floor of `1e-9`, picked between two
measurements rather than guessed:

- `1e-11` panics while `3e-11` does not, so the boundary is not sharp — it
  depends on `d1 * d2` as well as on `iv` — and the floor keeps an order of
  magnitude of margin below the lowest observed overflow.
- A historical walk estimates its volatility from the client's series, and this
  repo's own fixture reaches `4.5e-7`. My first attempt at `1e-6` broke five
  historical tests, which is how that constraint was found; `1e-9` stays 450x
  below it.

One guard closes both the new exposure and the older one.

## Columns

`gamma` is **split** into `gamma_call` / `gamma_put`. For a European option the
two are always equal, so this is future-proofing rather than redundancy:
upstream's `gamma()` falls through to a numerical estimate for non-European
types, which *is* style-dependent, while the mirror is only ever built as a
Call. The shared `gamma` column is **kept and still written** — it is upstream's
convenience mirror, still defined at expiry and at zero volatility where the
full set is not, and it is the only greek a pre-#74 row carries.

`delta_call` / `delta_put` serve as **both** the mirror and the snapshot's delta
rather than storing one number twice. That rests on the two agreeing, which
upstream does not promise in writing, so a test pins it across five regimes
including a sub-day expiry — and also pins that no strike carries a mirror
without a snapshot, the one shape the shared column cannot represent.

Every greek column is nullable. A strike whose option cannot be built has no
snapshot, and NULL is that absence, never a zero. Reconstruction refuses to
invent a partial snapshot: if any of the nine values `GreeksSnapshot` requires
is NULL, that side reads back as `None`.

## Who pays, and when

`build_chain` takes the decision as an argument instead of hardcoding it, and
`SimulationManager` turns it on **exactly when a warehouse is registered**. A
deployment that files nothing and whose clients never ask pays nothing; the API
prices per request in that case, as v1 always has.

The flag selects a different upstream branch, so two deployments of the same
build take two paths to the same market. Three tests hold that down:

| Test | What it compares |
|---|---|
| `test_the_greek_snapshots_do_not_move_the_priced_chain` | the two upstream branches, strike by strike |
| `test_a_warehouse_does_not_change_the_served_market` | two managers, every served value |
| `test_the_greeks_are_the_same_whether_read_or_priced` | two HTTP bodies at `?greeks=all`, byte for byte |

Whether a step is filed cannot change what a client is served.

## Memory: the budgets were counts, and the unit got heavier

`size_of::<OptionData>()` is 848 bytes in this build and `size_of::<QuoteRow>()`
is 620, both up by 408 (two `Option<GreeksSnapshot>` at 204). Two bounds count
contracts while documenting a **byte** budget, so holding the count would have
tripled resident memory nobody asked for:

| Bound | Was | Now | Budget held |
|---|---|---|---|
| `DEFAULT_MAX_CACHED_SNAPSHOT_CONTRACTS` | 4 000 000 | 1 200 000 | ≈1.02 GB |
| `SNAPSHOT_QUEUE_CONTRACTS` | 4 000 000 | 1 350 000 | ≈837 MB |

Both docs now show the arithmetic, and note the size is measured in this crate
where feature unification decides which upstream fields exist.

## On-disk cost, measured

On a real tape — 9 115 rows, 40 steps, three expirations, 101 strikes, priced by
upstream rather than repeated from a fixture — the table grows from **121 to 363
compressed bytes per row**, a factor of three, about 11 bytes a greek.

That contradicted the schema file's claim that the greek columns "compress to
almost nothing", and the comment is corrected: a `Decimal(38, 28)` greek is a
16-byte integer whose low-order bytes carry real, uncorrelated precision, and
ZSTD recovers about a third.

The alternatives are priced there rather than left to the next reader:
`Delta`, `DoubleDelta`, `Gorilla` and `T64` are all **rejected by ClickHouse**
for a 16-byte type; higher ZSTD levels measure no better (11.98 B/value at
ZSTD(1), 12.13 at ZSTD(3)); and the only real lever, `Decimal64(12) CODEC(T64)`
at 4.63 B/value, trades away the exact round trip the whole `Decimal(38, 28)`
choice exists to guarantee. `.env.example` warns beside the retention knob that
the same window now buys a third of the history.

## Migration: run, not documented

`CREATE TABLE IF NOT EXISTS` does not alter an existing table. An elided `ALTER`
in a comment would mean a deployment that boots green and then fails every
insert while its warehouse silently stops filling — the opposite of what turning
persistence on is supposed to buy. `ensure_schema` therefore **runs**
`simulation_option_quotes_greeks.sql` right after the `CREATE`.
`ADD COLUMN IF NOT EXISTS` is idempotent and metadata-only.

Verified end to end rather than reasoned: the local ClickHouse was reduced to
its 21 pre-#74 columns, and the next live test brought it back to 43 and passed.
Two hermetic tests keep the files in step — the migration adds exactly the
columns the DDL declares, and the range query selects only columns the DDL
declares. That second one is the `rho_d_call` typo class, which would otherwise
surface only in the integration job.

Rollback is safe, and the schema file now says why: an older binary's statements
name a subset of the columns, so it keeps working against a migrated table.

## Tests

- the snapshot round-trip carries greeks on both sides, proving field-for-field
  equality through the real database, including a `charm` that differs between
  the styles and an `alpha` absent on one of them
- the per-side projection is asserted against live ClickHouse: a call query
  projects `charm_call`, a put query `charm_put`, they differ, `rho` flips sign.
  A projection naming one side's column for both would pass every round-trip
  test in the file and still be wrong
- a row with no greek columns reads back as a row with no greeks, live and
  hermetically
- a partially stored snapshot rebuilds as absent, not as a partial one
- the two sides store genuinely different values; the style-independent greeks
  store identical ones

## Other

The SQL projection selects literal column names from a closed match instead of
composing identifiers, and the reconstruction names its columns from a
`&'static` table instead of formatting one per greek per style per row — up to
24 million allocations on a max-size read. `QuoteRow` and `ContractQuote` are
`#[non_exhaustive]`: the greek fields already broke downstream struct literals,
and this is the one moment where marking them costs nothing extra. `with_greeks`
became `with_greeks_call` / `with_greeks_put`.

ADR 0001 §12.2b records the stored column set, the executed migration, why
`simulation_generation` deliberately does not move, and that one generation can
now mix coverage across the upgrade.

## Reviewed before opening

`infrastructure-expert`, `architect` and `reproducibility-reviewer` audited the
diff; their findings are folded in rather than noted — the panic, the memory
budgets, the executable migration, the conditional cost, the mirror/snapshot
assumption, the per-row allocations, the composed SQL identifiers, and several
doc claims that measurement contradicted.

## Checks

- `LOGLEVEL=WARN cargo test --workspace` clean — 773 lib tests, 16 contract
  tests, 3 doctests
- `cargo test --workspace -- --ignored` clean — 21 live-service tests against
  Redis, ClickHouse 26.3 and MongoDB containers

## Stack

- **Base branch: `stack/2-73-full-greek-set`** — read the diff against that
  branch, not `main`; it is cumulative until #79 merges.
- **Stacked on: #79.**
- **Stack: #76 → #73 → #74 → #75.** This one is #74.
- **Blocks: #86**, which stacks on this one.


src/domain/factors.rs
```diff
@@ -795,6 +827,18 @@ fn reject_unpriceable_volatility(
         });
     }
 
+    if volatility < MIN_PRICEABLE_VOLATILITY {
```
joaquinbejar: 🔴 **Critical**: this checks only the base/row volatility, but upstream applies the caller-controlled skew/smile adjustment per strike and clamps that multiplier as low as `0.01`. A base IV equal to the allowed `1e-9` boundary passes this strict `<` check and can become `1e-11`, exactly in the `vomma` overflow/panic region documented above. Validate the effective adjusted IV before full-Greek calculation (or derive an inclusive conservative base bound), and add an extreme skew/smile regression.

src/api/rest/greeks.rs
```diff
@@ -0,0 +1,422 @@
...
+                .as_ref()
+                .and_then(|snapshot| GreeksResponse::from_snapshot(snapshot, level)),
+            data.greeks_put
+                .as_ref()
+                .and_then(|snapshot| GreeksResponse::from_snapshot(snapshot, level)),
+        );
+    }
+
+    let mut priced = data.clone();
+    priced.calculate_greeks();
```
joaquinbejar, line 382: 🔴 **Critical**: v1 reaches this with arbitrarily small accepted volatility, while the new pricing floor runs only in v2 `FactorTape::build`. A v1 session created or updated with values such as `1e-13` can panic in upstream `vomma` on `greeks=first` or `greeks=all`; on the step endpoint the cursor has already been saved before rendering returns the resulting 500. Apply the same adjusted-IV validation to v1 before building or mutating state, and cover create, PUT/PATCH, peek, and step.

src/infrastructure/clickhouse/schema/simulation_option_quotes_greeks.sql
```diff
@@ -0,0 +1,47 @@
...
+--   Run AFTER the `CREATE TABLE IF NOT EXISTS` of the same table. On a fresh
+--   deployment every column already exists and this is a no-op; on an existing
+--   one it is the upgrade. The physical column order therefore differs between
+--   a migrated table and a fresh one, which is harmless: every INSERT and every
+--   SELECT in this crate names its columns.
+--
+-- The column list must stay in step with `simulation_option_quotes.sql`; a test
+-- asserts that it does.
+ALTER TABLE simulation_option_quotes
+    ADD COLUMN IF NOT EXISTS gamma_call Nullable(Decimal(38, 28)) CODEC(ZSTD(1)),
```
joaquinbejar: 🟠 **Warning**: `Nullable(...)` alone is not an explicit default to clickhouse-rs 0.15 validation. After this migration, a rolled-back binary whose old row struct omits these 22 fields gets `SchemaMismatch` on every insert because their `default_type` is unset, contrary to the rollback-safe contract above. Add `DEFAULT NULL` to every new column in both DDL files and cover an old-row insert against the migrated live schema.

src/session/manager_v2.rs
```diff
@@ -451,7 +457,13 @@ impl SimulationManager {
         }
 
         let tape = self.tape_for(simulation).await?;
-        let snapshot = SeriesBuilder::new(&simulation.parameters, &tape)?.snapshot(step)?;
+        // The greek snapshots are built only when something will read them:
+        // a registered warehouse files every step, and a filed step has to
+        // carry what a replayed one does (issue #74). Without a warehouse the
+        // API prices them per request instead, and only when asked.
+        let snapshot = SeriesBuilder::new(&simulation.parameters, &tape)?
+            .with_greek_snapshots(self.warehouse.is_some())
```
joaquinbejar: 🟠 **Warning**: registering a warehouse now makes even a default peek or advance synchronously price up to 200,000 contracts with both full Greek snapshots on the Actix/Tokio worker. Move the entire `SeriesBuilder::snapshot` call behind bounded CPU/blocking admission in this PR; a later stacked change does not make this independently mergeable head safe.

joaquinbejar (commented):
The persistence mapping and CI coverage are strong, but two tiny-volatility paths still reach an upstream panic, the migration breaks the documented old-binary rollback, and warehouse-backed snapshots now perform max-size Greek pricing on the async worker. These are merge blockers, so REQUEST_CHANGES. GitHub does not let the PR author account set that event, so this review is posted as COMMENT. The diff is well over 500 lines; splitting the volatility guard, schema migration, and execution-path changes would make the fixes safer to verify.

src/domain/factors.rs
joaquinbejar: You are right, the base is not what a strike is priced at. Upstream clamps the skew and smile multiplier at 0.01, so a base at the old boundary could reach 1e-11. The bound is now derived from that clamp, effective floor over 0.01, which puts the request check at 1e-7, and there is a regression that drives skew and smile to the edges of their ranges at the smallest admitted base and builds the chain with the full greek set.

src/api/rest/greeks.rs
joaquinbejar: Fixed. The guard now runs at the v1 request boundary, which covers create and PUT and PATCH through the two conversions, and again before pricing on the peek and the step. The step checks before the cursor moves, since a 500 after the advance costs the client a step it cannot get back. There is a test that writes a session straight into the store the way an older binary would and asserts the 400 names volatility while the default level is still served.

src/infrastructure/clickhouse/schema/simulation_option_quotes_greeks.sql
joaquinbejar: Confirmed and fixed, both DDL files now carry DEFAULT NULL on all 22 columns. I checked the test is actually load bearing rather than passing by accident, with the defaults removed by hand the old row struct fails with exactly the SchemaMismatch you described, and passes with them.

src/session/manager_v2.rs
joaquinbejar: Agreed that a later stacked change does not make this head safe, so it is in this PR. SeriesBuilder::snapshot now runs under the same bounded admission the renderers use, which I moved to utils::admission so the session layer and the API share one bound rather than two, since they compete for the same cores. The permit is taken before the blocking task spawns, so a client that disconnects while queued costs nothing.

joaquinbejar:
All four are fixed in 54a9ec1. The volatility one was the interesting catch, since the base a client sends is not what a strike is priced at, so the bound now comes from the skew and smile clamp itself rather than from a number I picked. v1 gets the same guard at the request boundary and again before pricing, with the step checking before the cursor moves. The DEFAULT NULL fix I verified both ways, removing the defaults by hand reproduces the SchemaMismatch exactly. And the snapshot build now shares one admission bound with the renderers in utils::admission, in this PR rather than the next one. Gate is green at 782 lib tests plus 22 live service tests, clippy, release and cargo doc all clean.

PR #86, opened 2026-08-28
Include the full greek set in the export datasets (#75)

### LINKED ISSUE #75: Include the full greek set in the export datasets

joaquinbejar:
## What

The `option_chains` export emitted `implied_volatility`, `call_delta`,
`put_delta` and `gamma`, so a backtester reading an exported tape got a strictly
poorer chain than a client stepping the same simulation live. `greeks` now takes
the same three values, with the same meaning and the same default, as the chain
endpoints.

Closes #75. Completes the stack.

| level | appends |
|---|---|
| `none` (default) | nothing — the seventeen columns are frozen |
| `first` | `call_theta`, `put_theta`, `call_vega`, `put_vega`, `call_rho`, `put_rho`, `call_rho_d`, `put_rho_d` |
| `all` | the above, then seven more per style — `gamma` through `color` — as `call_*` / `put_*` |

Each level's header is a **prefix** of the next, so raising it appends columns
and never moves one. The set is fixed per level, never per row: a strike whose
greeks upstream could not compute writes empty fields rather than fewer of them,
which is what keeps a row parseable by position.

`delta` is not repeated — `call_delta` / `put_delta` already carry it and the
two agree. `gamma` **is**, because the shared mirror stays defined at expiry and
at zero volatility where the per-style pair goes blank.

## Measured

5 996-row export, release build:

| Level | CSV | JSON | Wall time |
|---|---|---|---|
| `none` | 1.27 MB | 2.54 MB | 303 ms |
| `first` | 2.23 MB | 4.06 MB | 459 ms |
| `all` | 4.01 MB | 6.87 MB | 469 ms |

`first` and `all` cost the same to compute — upstream builds the snapshot whole
and the level only decides what is written — so that step is paid in bytes, not
in time.

## Why the two conversion paths cannot drift

Both `QuoteView` constructors go through one `GreeksView::of`, and the replay
path builds with `with_greek_snapshots(level.wants_greeks())` — the flag #74
added, which is why it was threaded rather than hardcoded.

`GreeksView::of` **destructures** the upstream snapshot, so a thirteenth greek
is a compile error here rather than a column the export quietly stops carrying
while the response and the warehouse both gain it. Same discipline `ApiWalkType`
uses for a new `WalkType` variant.

### The shape both reviewers caught

A stored step that cannot answer the level is now **replayed** rather than
served short. A row filed before the greek columns existed reconstructs with no
snapshots, so preferring it at `greeks=all` put empty greek columns for the
persisted prefix of a range and real ones for the replayed rest — one file, two
coverages, no signal, and a `200`. For IronCondor that is the worst possible
shape: a tape that looks complete and attributes P&L against blanks for an
arbitrary prefix. Reproduced, fixed, and pinned by a test that files a pre-#74
tape and asserts every row comes back with its greeks.

## An hour spent on a bug that was mine

The JSON-versus-CSV test failed on `gamma`: `0.009547174464993617` against
`0.009547174464993615` — a genuine one-ULP difference between two f64 values.
The encoders agreed when handed the same value, the CSV export repeated
byte-identically twelve times, and it reproduced on every run. All of it pointed
at the export computing different numbers per format.

It was not. The raw JSON bytes carry `...615`, exactly as the CSV does. The
shift came from the test: `serde_json`'s float parser is not bit-exact for that
value, so parsing the response and re-rendering it moved the last bit. Review
put a number on it — over 200 000 values in that magnitude band, the two parsers
disagree **16.6%** of the time.

I should have suspected the instrument before the subject. The test now compares
raw tokens, and parses both sides with one correctly-rounded parser where the
comparison has to be numeric — which is what the criterion asks for anyway,
since the encodings render the same value differently: JSON writes `4950.0`
where CSV writes `4950`, on every strike of every row, and takes exponent form
below `1e-5`. Both facts are now in the test, the crate docs and ADR §10.2.

## Two fixes this stack's review surfaced, outside #75's scope

Called out because they are not export changes:

- **`SimulationManager::snapshot_at` priced its chains inline on an Actix
  worker.** `tape_for` was moved to the blocking pool for exactly that reason,
  with a comment explaining why; the chain build beside it was not, and #74 made
  it ~1.54x heavier for every persistence-enabled deployment. Now on the
  blocking pool too.
- A public doc comment linked a private constant, putting back a `cargo doc`
  warning that a7d8d80 had cleared.

## Deliberately not done here

- **`deny_unknown_fields` on the v2 query DTOs.** `?greek=all` (typo) is
  silently the default level, which is the same failure the typed 400 exists to
  prevent, reached through the key instead of the value. ADR §11 says the v2
  request surface rejects unknown fields, so the code should follow — but it is
  a behaviour change to v2 query parsing and does not belong in a PR about
  export columns. Worth its own issue.
- **A degenerate strike's `greeks` key is omitted rather than `null`** on the
  chain responses, which is indistinguishable from `greeks=none`. Documented in
  `greeks.rs`; narrow in practice, since any non-degenerate strike in the same
  response settles it. Belongs to #73's surface.

## Tests

- the two sources agree at every level, in both encodings
- a stored step without greeks is replayed rather than served short, while the
  same record still answers the default level from storage
- the default export is byte-identical to `greeks=none` and carries no greek
  column; the default header is frozen literally rather than derived from the
  constant it guards
- exporting twice at the same level is byte-identical, at every level, both
  encodings
- the JSON export carries the same values as the CSV export at every level
- an unknown level is a typed 400 raised before a byte is streamed and before
  the store is read, so `?greeks=bogus` on a missing simulation is 400 here
  exactly as on the snapshot endpoint
- the export joins both OpenAPI greek tests — which is what would have caught
  the duplicate `400` entry that silently dropped its own description

ADR 0001 §10 is amended: the query grammar, the per-level column table, the sign
convention, why the representation differs by surface, the replay rule, and the
"no exponent" claim the greek columns falsified.

## Release

**The next release is `0.3.0`.** #76 made `positive` 0.6 a public-dependency
major bump; #73 added public types and a public field to two public structs.
IronCondor talks HTTP, not the crate, and never sends `greeks`, so it keeps
receiving byte-identical v1 payloads — no coordinated release needed, but its
`RawQuote` can drop its placeholder zeros once this lands.

## Checks

- `LOGLEVEL=WARN cargo test --workspace` clean — 781 lib tests, 16 contract
  tests, 3 doctests
- `cargo test --workspace -- --ignored` clean — 21 live-service tests

## Stack

- **Base branch: `stack/3-74-persist-greek-set`** — read the diff against that
  branch, not `main`; it is cumulative until #85 merges.
- **Stacked on: #85.**
- **Stack: #76 → #73 → #74 → #75.** This is #75, the top.

### DIFF
--- a/src/api/rest/export.rs
+++ b/src/api/rest/export.rs
@@ -808,7 +808,7 @@ impl Stream for RowStream {
     ),
     responses(
         (status = 200, description = "The exported rows, streamed", body = String),
-        (status = 400, description = "An invalid range or an unknown `greeks` level, as the typed `{error, field}` of ValidationErrorResponse — `field` is `from_step`, `to_step` or `greeks`. An unknown `dataset` or `format` is rejected by the query extractor before the handler and carries actix's own untyped body instead"),
+        (status = 400, description = "Every rejection carries the typed `{error, field}` of ValidationErrorResponse. `field` names the offender where it is known — `id` for a malformed identifier, `from_step` or `to_step` for a range that is reversed or past the tape, `greeks` for an unknown level. It is EMPTY when the query string fails to deserialise at all (an unknown `dataset` or `format`, a non-numeric step), because serde's message for those does not name the key; the `error` string carries the detail.", body = crate::api::rest::responses::ValidationErrorResponse),
         (status = 404, description = "Simulation not found"),
         (status = 500, description = "Internal server error")
     )
@@ -988,8 +988,16 @@ fn produce(
                 .ok_or_else(|| ChainError::Internal("the step counter overflowed".to_string()))?;
         }
 
-        let chunk = writer.rows(parameters, row.step, row, chains)?;
-        if !chunk.is_empty() && sender.blocking_send(Ok(chunk)).is_err() {
+        let delivery = writer.rows(parameters, row.step, row, chains, &mut |chunk| {
+            if chunk.is_empty() {
+                return Delivery::Sent;
+            }
+            match sender.blocking_send(Ok(chunk)) {
+                Ok(()) => Delivery::Sent,
+                Err(_) => Delivery::ClientGone,
+            }
+        })?;
+        if delivery == Delivery::ClientGone {
             return Ok(());
         }
     }
@@ -1067,13 +1075,17 @@ impl Writer {
     /// chains: it is the tape's instant, the same one a stored record was
     /// written from, and taking it from one place keeps the two sources
     /// rendering identically by construction.
-    fn rows(
+    fn rows<E>(
         &mut self,
         parameters: &SimulationParametersV2,
         step: usize,
         row: &crate::domain::factors::FactorRow,
         chains: Option<StepChains<'_>>,
-    ) -> Result<Vec<u8>, ChainError> {
+        emit: &mut E,
+    ) -> Result<Delivery, ChainError>
+    where
+        E: FnMut(Vec<u8>) -> Delivery,
+    {
         let simulated_at = render_instant(row.simulated_at);
         let symbol = parameters.symbol.as_str();
 
@@ -1085,6 +1097,7 @@ impl Writer {
             } => {
                 let values = json_rows(*dataset, *level, step, &simulated_at, symbol, row, chains);
                 let mut chunk = Vec::new();
+                let mut rows_buffered = 0_usize;
                 for value in values {
                     if !*first {
                         chunk.push(b',');
@@ -1094,12 +1107,28 @@ impl Writer {
                         ChainError::Internal(format!("failed to encode an export row: {e}"))
                     })?;
                     chunk.extend_from_slice(&encoded);
+                    rows_buffered += 1;
+
+                    if rows_buffered >= ROWS_PER_CHUNK {
+                        if emit(std::mem::take(&mut chunk)) == Delivery::ClientGone {
+                            return Ok(Delivery::ClientGone);
+                        }
+                        rows_buffered = 0;
+                    }
                 }
-                Ok(chunk)
+                if chunk.is_empty() {
+                    return Ok(Delivery::Sent);
+                }
+                Ok(emit(chunk))
             }
             Writer::Csv { dataset, level } => {
                 let records = csv_rows(*dataset, *level, step, &simulated_at, symbol, row, chains);
-                encode_csv(&records)
+                for batch in records.chunks(ROWS_PER_CHUNK) {
+                    if emit(encode_csv(batch)?) == Delivery::ClientGone {
+                        return Ok(Delivery::ClientGone);
+                    }
+                }
+                Ok(Delivery::Sent)
             }
         }
     }
@@ -1232,6 +1261,34 @@ fn json_rows(
             rows
         }
     }
+}
+
+/// How many rows one chunk carries.
+///
+/// The export streams, but a chunk is indivisible: the channel bounds how many
+/// are in flight, not how large they are, so a whole step in one chunk means a
+/// slow client can hold the entire step in memory no matter how small the
+/// channel is. At the per-snapshot cap of 200 000 contracts a `greeks=all` step
+/// is hundreds of megabytes, and sixteen of those in flight is gigabytes.
+///
+/// Rows rather than bytes because a row's width is fixed by the dataset and the
+/// greek level, so a row count IS a byte bound: 512 rows of the widest shape the
+/// `option_chains` dataset can produce is a few hundred kilobytes, and the
+/// narrower datasets are far less. Backpressure then applies within a step
+/// rather than only between steps.
+const ROWS_PER_CHUNK: usize = 512;
+
+/// Whether a chunk reached the client.
+///
+/// A `bool` would read as "success", and the interesting case here is the one
+/// that is neither success nor failure: the client hung up, which is not an
+/// error to report to anyone, just a reason to stop working.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Delivery {
+    /// The chunk was handed to the response stream.
+    Sent,
+    /// The receiver is gone. Stop producing; nothing is owed to anyone.
+    ClientGone,
 }
 
 /// The CSV records one step contributes, in the header's order.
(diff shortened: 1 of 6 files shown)


src/api/rest/export.rs
```diff
@@ -858,8 +1078,12 @@ impl Writer {
...
 
         match self {
-            Writer::Json { dataset, first } => {
-                let values = json_rows(*dataset, step, &simulated_at, symbol, row, chains);
+            Writer::Json {
+                dataset,
+                level,
+                first,
+            } => {
+                let values = json_rows(*dataset, *level, step, &simulated_at, symbol, row, chains);
```
joaquinbejar, line 1098: 🟠 **Warning**: this still materializes one whole step as `Vec<Value>` (and CSV as `Vec<Vec<String>>`) and then allocates a second whole-step encoded buffer before channel backpressure applies. Using the PR measurement, 200,000 `greeks=all` contracts are roughly a 229 MB JSON chunk before object overhead; a 16-slot channel bounds chunk count, not bytes, and can retain multiple GB for a slow client. Emit bounded row/byte subchunks directly in both encodings.

src/session/manager_v2.rs
```diff
@@ -451,7 +457,29 @@ impl SimulationManager {
...
+        let parameters = simulation.parameters.clone();
+
+        // Off the runtime, for the same reason `tape_for` is: pricing a
+        // snapshot is real synchronous CPU — up to
+        // `DEFAULT_MAX_SNAPSHOT_CONTRACTS` priced contracts, and about 1.54x
+        // that again once the greek snapshots are on — and a worker holding it
+        // stalls every other request that worker owns. This build predates the
+        // greek work; #74 made it heavy enough that leaving it inline stopped
+        // being defensible.
+        let snapshot = tokio::task::spawn_blocking(move || {
```
joaquinbejar: 🟠 **Warning**: concurrent cache misses for the same `(simulation, step)` each enqueue this non-cancellable, potentially 200,000-contract Greek build. The cache is populated only after `.await`, so a disconnected or cancelled owner also discards the completed result and leaves the cache cold. Mirror `tape_for` with one per-key owner and waiters, bounded admission, and cache publication from work that survives caller cancellation.

src/api/rest/export.rs
```diff
@@ -624,19 +793,22 @@ impl Stream for RowStream {
...
+        ("greeks" = Option<String>, Query, description = "How much of the greek set the option_chains dataset carries: `none` (default), `first` (appends call_theta, put_theta, call_vega, put_vega, call_rho, put_rho, call_rho_d, put_rho_d) or `all` (appends seven more per style, gamma through color: fourteen columns). Each level's header is a PREFIX of the next, so raising it appends columns and never moves one. Values are per ONE LONG CONTRACT. Accepted but immaterial for the underlying and volatility datasets, which carry no chains. An unknown value is a 400")
     ),
     responses(
         (status = 200, description = "The exported rows, streamed", body = String),
-        (status = 400, description = "Unknown dataset or format, or an invalid range; body carries `error` and `field`"),
+        (status = 400, description = "An invalid range or an unknown `greeks` level, as the typed `{error, field}` of ValidationErrorResponse — `field` is `from_step`, `to_step` or `greeks`. An unknown `dataset` or `format` is rejected by the query extractor before the handler and carries actix's own untyped body instead"),
```
joaquinbejar: 🟠 **Warning**: this `400` contract does not match the handler/extractor paths. An invalid UUID returns the typed validation body with `field = id`, which is omitted here, while a non-numeric `from_step` or `to_step` is rejected by `web::Query` before the handler with Actix’s untyped body. Document both cases accurately, or install typed extraction so the declared response schema is true.

joaquinbejar (commented):
The Greek export source-parity tests are strong, but whole-step materialization defeats bounded-memory streaming, new snapshot blocking jobs can stampede and lose completed work on cancellation, and the documented 400 shapes do not match extractor behavior. These need fixing before merge, so REQUEST_CHANGES. GitHub does not let the PR author account set that event, so this review is posted as COMMENT. The diff is well over 500 lines; separating export buffering, snapshot scheduling, and API-contract changes would make the follow-up easier to verify.

src/api/rest/export.rs
joaquinbejar: Fixed, the writer now emits at most ROWS_PER_CHUNK rows per chunk in both encodings, so backpressure applies within a step rather than only between steps. It reports delivery instead of returning bytes, since it emits several chunks per step and has to notice a disconnect itself. Two tests cover it, a wide step arrives as more than one bounded chunk in both encodings, and a refused chunk stops the step immediately.

src/session/manager_v2.rs
joaquinbejar: Both parts fixed by mirroring tape_for. There is now one owner per simulation and step with waiters on a broadcast, and the result is cached from inside the job rather than after the await, so a caller that disconnects mid build no longer discards a snapshot that ran to completion and leaves the cache cold. The build also runs under the shared admission bound that landed in #85.

src/api/rest/export.rs
joaquinbejar: You were right on both counts and I probed all four paths rather than reasoning about them. The id case answers with field id, and the extractor rejections now render the typed object with an empty field, since serde message for a bad key does not name it, so the description says exactly that and the operation declares the schema. A test pins all five cases, and the OpenAPI guard is now per operation, a schema may only be published where every rejection has one shape.

joaquinbejar:
All three are fixed in 0e13107. The export now chunks within a step so backpressure actually applies, and the writer reports delivery rather than returning bytes since it emits several chunks per step. Snapshot builds got the same single flight the tape already had, plus caching from inside the job so a cancelled caller no longer throws away completed work. And I probed all four ways of getting a 400 out of the export rather than reasoning about them, which confirmed both of your points, so the description now names the id case and says plainly that the extractor rejections carry an empty field. Gate is green at 794 lib tests and 22 live service tests, with clippy, release and cargo doc all clean.

Issue #87, opened 2026-08-28
Percent-encode Redis credentials in RedisConfig::url


joaquinbejar:
Overview

`RedisConfig::url` (`src/infrastructure/config/redis.rs`) interpolates the username and password into the URL verbatim, with no percent-encoding. A credential containing `/`, `#`, `?` or a space therefore produces a URL that is wrong rather than one that fails loudly: a `/` shifts what the driver reads as the database index, and a `#` truncates everything after it.

The `Display` impl in the same file already names "delimiter-containing credentials" as a hazard it sidesteps, so the gap is known; it has simply never been closed.

Context

Found while reviewing #83. A test there originally asserted that a password of `s3cr:t@pass` reaches the URL verbatim, which would have frozen the missing encoding as intended behaviour. That assertion was narrowed to a plain password, and the punctuated case now only asserts the resolved config field, so the gap stays a gap rather than becoming a contract.

Tasks

- [ ] Percent-encode the username and password when building the URL, using the userinfo character set
- [ ] Keep the resolved config fields unencoded, so a caller reading `RedisConfig.password` still sees what the operator set
- [ ] Extend the redaction in `Display` and `redact_userinfo` to the encoded form

Acceptance criteria

- [ ] A password containing `/`, `#`, `?`, `@`, `:` and a space produces a URL that parses to the same credential
- [ ] The database index is unaffected by a credential containing `/`
- [ ] An alphanumeric credential produces a byte-identical URL to today, so nothing already working changes
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean
- [ ] `cargo fmt --all --check` clean
- [ ] `LOGLEVEL=WARN cargo test --workspace` clean

Issue #84, opened 2026-08-28
Add /health and /ready endpoints


joaquinbejar:
## Overview

There is no health endpoint. `/health`, `/healthz`, `/api/health` and `/ping` all
return 404, so the only liveness signal is `GET /metrics`, which returns 200
`text/plain`.

That works as a probe by accident, not by contract: `/metrics` is a Prometheus
surface whose availability is not a statement about the service being ready to
serve simulations, and anything written against it breaks the moment the metrics
surface is moved or gated.

## Context

The gap shows up immediately in orchestration. `Docker/docker-compose.yml`
defines healthchecks for redis (lines 58-62) and mongodb (83-87) but **none for
the backend service**, so `docker compose up -d` reports the simulator healthy the
instant the container starts, well before it has connected to Redis and Mongo.
Anything that waits on that signal races.

The distinction matters here because startup does real work: `RedisClient::new`
at `src/main.rs:121` and `init_mongodb` at `:132` both abort the process on
failure, and with `OCS_SNAPSHOT_PERSISTENCE_ENABLED=true` so does
`warehouse.ensure_schema()` at `:164`.

Consumer context: Chainwalk
(`gitlab.taunais.com/capitaldelta/chainwalk`) plans to run several OCS instances
and shard tape materialisation across them, which needs a real readiness signal to
know which instances can take work.

## Tasks

- [ ] Add `GET /health` as a liveness probe: the process is up, no dependency
      checks, always cheap and always 200
- [ ] Add `GET /ready` as a readiness probe: returns 200 only once Redis and
      MongoDB are reachable, and ClickHouse too when snapshot persistence is
      enabled. Return 503 with a body naming the failing dependency otherwise
- [ ] Keep both unauthenticated and out of the metrics middleware, so probing does
      not pollute request metrics
- [ ] Add a healthcheck for the backend service in
      `Docker/docker-compose.yml` and `docker-compose.dev.yml`, pointing at
      `/ready`
- [ ] Document both endpoints in the README and the OpenAPI document

## Acceptance criteria

- [ ] `GET /health` returns 200 while dependencies are down
- [ ] `GET /ready` returns 503 naming the dependency while Redis is unreachable,
      and 200 once it is back, without a restart
- [ ] Neither endpoint appears in the request metrics
- [ ] `docker compose up -d` reports the backend healthy only after it can serve a
      v2 create request
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean
[...]

Issue #82, opened 2026-08-28
Make the bind address and port configurable


joaquinbejar:
## Overview

The listen address and port are compile-time constants:

```rust
let listen_on = ListenOn::All;
let port = 7070;
```

(`src/main.rs:198-199`). There is no environment variable for either. If 7070 is
taken, the only options are patching `main.rs` or publishing a different host
port through Docker.

## Context

This blocks a use case that is now concrete: running **several OCS instances in
parallel on one host**. Tape materialisation is embarrassingly parallel because
each simulation is independent and shares no state, so a batch consumer scales by
sharding work across N instances. With a hardcoded port that requires N
containers where N processes would do.

Consumer context: Chainwalk
(`gitlab.taunais.com/capitaldelta/chainwalk`) materialises one tape per Monte
Carlo path. A study of 200 paths is 200 independent materialisations, and its
scaling plan is to fan them out across several OCS instances.

Binding to all interfaces by default is also worth revisiting. This service has
no authentication and no rate limiting, so the safe default is loopback, with
exposure an explicit choice.

## Tasks

- [ ] Read `OCS_BIND_ADDRESS` (default `127.0.0.1`) and `OCS_PORT` (default
      `7070`)
- [ ] Validate both at startup and fail with a `ChainError::Validation` naming the
      variable, matching how the v2 and snapshot knobs already behave
- [ ] Log the effective bind address and port at startup
- [ ] Document both in `.env.example` and the README
- [ ] Note the default change in the changelog: a deployment relying on the
      implicit `0.0.0.0` must now set `OCS_BIND_ADDRESS=0.0.0.0`

## Acceptance criteria

- [ ] Two instances started with different `OCS_PORT` values run concurrently on
      one host and serve independent simulations
- [ ] An invalid port or address fails startup with an error naming the variable,
      not a panic
- [ ] With neither variable set, the service binds `127.0.0.1:7070`
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean
- [ ] `cargo fmt --all --check` clean
- [ ] `LOGLEVEL=WARN cargo test --workspace` clean

Issue #81, opened 2026-08-28
Honour LOGLEVEL instead of hardcoding DEBUG


joaquinbejar:
## Overview

The service always logs at `DEBUG`. `src/main.rs:116` calls
`setup_logger_with_level("DEBUG")` with a literal, and `LOGLEVEL` is read
**nowhere** in `src/`: `grep -rn LOGLEVEL src/` returns zero matches.

`.env.example` documents the variable, so an operator sets it, sees no change,
and reasonably concludes the logging config is broken rather than absent.

## Context

The cost is not cosmetic. At `DEBUG` the service emits hyper connection traces on
every request, which in a batch consumer that materialises hundreds of tapes
turns into a large volume of log output that hides anything worth reading and
costs real IO.

## Tasks

- [ ] Read `LOGLEVEL` in `main.rs`, defaulting to `INFO`, and pass it to
      `setup_logger_with_level`
- [ ] Accept the usual level names case-insensitively; an unrecognised value
      warns once and falls back to the default rather than aborting startup
- [ ] Log the effective level at startup so the setting is observable
- [ ] Reconcile `.env.example` with the implemented behaviour

## Acceptance criteria

- [ ] `LOGLEVEL=WARN` produces no `INFO` or `DEBUG` records; a test asserts the
      resolved level rather than scraping output
- [ ] An unset `LOGLEVEL` resolves to `INFO`
- [ ] An invalid `LOGLEVEL` warns and falls back without aborting
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean
- [ ] `cargo fmt --all --check` clean
- [ ] `LOGLEVEL=WARN cargo test --workspace` clean

PR #88, opened 2026-08-28
Update every dependency, and treat a blank environment variable as unset (#83)

### LINKED ISSUE #83: Treat a blank environment variable as unset, consistently
## Overview

Blank environment variables are treated inconsistently, and the inconsistent half
produces a failure that looks like something else entirely.

`src/infrastructure/config/redis.rs` reads credentials with
`env::var("REDIS_PASSWORD").ok()`, so `REDIS_PASSWORD=` yields `Some("")` rather
than `None`. `RedisConfig::url()` then builds `redis://:@host:6379`, which is an
AUTH attempt with an empty password against a server that has no password set.
The connection fails, and the message points at authentication rather than at an
empty variable.

`.env.example` writes exactly that blank form, so the shipped example produces the
failure.

## Context

The codebase already knows the right answer and applies it elsewhere.
`src/infrastructure/config/simulation_v2.rs` carries a helper documented as
"Reads a variable, treating an empty or whitespace-only value as unset. A blank
value in a `.env` file is how a knob gets commented out in practice; treating it
as unset is friendlier than failing startup over it." `snapshot.rs` does the same.

So the v2 and snapshot knobs trim blank to unset while the Redis, Mongo and
ClickHouse configs do not. That split is invisible until it bites, and it bites
during first-run setup, which is the worst moment.

## Tasks

- [ ] Extract the blank-is-unset helper from `simulation_v2.rs` into a shared
      config utility
- [ ] Apply it to every `env::var(...).ok()` in the Redis, Mongo and ClickHouse
      configs, credentials included
- [ ] Audit `.env.example` so no variable is shown blank where blank now means
      unset; comment them out instead
- [ ] State the rule once in the config module docs: a blank value is an unset
      value, everywhere

## Acceptance criteria

- [ ] `REDIS_PASSWORD=` connects successfully to a password-less Redis; a test
      asserts the resolved config carries `None`, not `Some("")`
- [ ] The same holds for whitespace-only values, and for the Mongo and ClickHouse
      credential variables
- [ ] A genuinely set password still reaches the URL unchanged

joaquinbejar:
## What

Two things, both at the bottom of the stack because everything above compiles
against them.

Closes #83.

### Dependencies

Every workspace dependency moved to its latest stable, pinned `major.minor` as
`[workspace.dependencies]` requires. Five moved; the other twenty-three were
already there.

| crate | from | to |
|---|---|---|
| `uuid` | 1.23 | 1.26 |
| `tokio` | 1.52 | 1.53 |
| `actix-web` | 4.14 | 4.15 |
| `actix-files` | 0.6 | 0.7 |
| `redis` | 1.3 | 1.6 |

**None crosses a stable major.** `actix-files` 0.6 → 0.7 is the only
semver-breaking one, since for a `0.x` crate the minor *is* the major. Nothing
here needed a change for it, and the favicon route it serves keeps its existing
test.

### Blank environment variables

`REDIS_PASSWORD=` read as `Some("")`, which built `redis://:@host:6379` — an
AUTH attempt with an empty password against a server that has none. The
connection failed with an authentication error pointing nowhere near the cause,
and `.env.example` shipped exactly that line, so the failure was reachable by
following the setup instructions.

The codebase already knew the answer. `simulation_v2.rs` and `snapshot.rs`
carried a helper treating blank as unset and documented why; Redis, Mongo and
ClickHouse did not. That split is invisible until first-run setup.

`utils::env::read_var` is now that helper, with the rule stated once: **a blank
value is an unset value**. It lives in `utils` rather than in
`infrastructure::config` because `utils::admission` reads a knob of its own and
must not import a layer above it. Every knob in the service routes through it —
the three service configs, the three v1 request caps, `OCS_MAX_CACHED_WALKS` and
`OCS_MAX_CONCURRENT_PRICING_JOBS` — so a blank knob now falls back in silence
instead of warning that the operator's config is invalid. Values are trimmed, so
padding never reaches a host name, a credential or a parser.

### One consequence, stated rather than buried

`ClickHouseConfig.password` is a plain `String` whose default is the literal
`"password"`, so applying the rule to it makes an **empty ClickHouse password
inexpressible through the environment** — and a stock ClickHouse `default` user
has exactly that. `CLICKHOUSE_PASSWORD=` used to be the only way to say it.

I kept the rule uniform anyway: the alternative is one variable in the whole
service behaving differently from every other, which is the split #83 exists to
remove. The consequence is documented at the field, in `.env.example` and here,
and a password-less deployment sets `ClickHouseConfig` directly, which is public.
Say the word if you would rather have the exception.

### Also fixed, found by review

`Docker/docker-compose*.yml` used `${VAR-default}`, which substitutes only when a
variable is **unset**. A blank one reached the container as `""`, where the
service applied its own default — `localhost`, which inside a container points at
nothing. Every interpolation is now `:-`.

`rust-version = "1.88"` is declared: actix-web 4.15 and redis 1.6 both require
it, so the bump moved the MSRV. Without it a consumer on an older toolchain gets
a failure inside a dependency instead of a clear message about this crate.

`.env.example` states the rule at the top and shows no variable blank: the
optional Redis credentials are commented out, which is what "leave unset" has to
look like once blank means unset.

## Tests

Six new. The load-bearing ones:

- a blank or whitespace-only Redis credential resolves to `None` **and** produces
  a URL with no userinfo at all, which is the actual bug
- a real password still reaches the URL verbatim, punctuation included, so the
  rule cannot quietly eat a legitimate value
- blank Mongo and ClickHouse variables fall back to their documented defaults
  rather than connecting to the empty string
- `read_var` itself: unset, empty, spaces and a tab all read as absent; a padded
  real value survives trimmed

The credential test deliberately does **not** assert that an unencoded
`s3cr:t@pass` round-trips through the URL. `RedisConfig::url` does no
percent-encoding, so asserting that would freeze a known gap as intended
behaviour; the punctuated case asserts the resolved field only, and the gap is
now issue #87.

## Checks

- `LOGLEVEL=WARN cargo test --workspace` clean — 801 lib tests, 16 contract
  tests, 3 doctests
- `cargo test --workspace -- --ignored` clean — 22 live-service tests against
  Redis, ClickHouse and MongoDB containers
- `make readme` regenerated

Reviewed by `architect` before opening; its findings are folded in above rather
than noted.

## Stack

- **Base branch: `main`** — this is the bottom.
- **Stack: #83 → #81 → #82 → #84 → #80 → #78.** This one is #83.
- Honest note on the topology: #83, #81 and #82 are logically independent of one
  another. They are stacked because they overlap in `main.rs` and `Cargo.toml`,
  not because one needs another; #80 → #78 is the only hard ordering, since both
  rewrite the export column lists.
- **Blocks: #89**, which stacks on this one.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,10 @@
 name = "optionchain_simulator"
 version = "0.2.0"
 edition = "2024"
+# Declared because the dependency bump moved it: actix-web 4.15 and redis 1.6
+# both require 1.88. Without this a consumer on an older toolchain gets a
+# failure inside a dependency rather than a clear message about this crate.
+rust-version = "1.88"
 authors = ["Joaquin Bejar <jb@taunais.com>"]
 description = "OptionChain-Simulator is a lightweight REST API service that simulates an evolving option chain with every request. It is designed for developers building or testing trading systems, backtesters, and visual tools that depend on option data streams but want to avoid relying on live data feeds."
 license = "MIT"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,10 +2,12 @@
 name = "optionchain_simulator"
 version = "0.2.0"
 edition = "2024"
-# Declared because the dependency bump moved it: actix-web 4.15 and redis 1.6
-# both require 1.88. Without this a consumer on an older toolchain gets a
-# failure inside a dependency rather than a clear message about this crate.
-rust-version = "1.88"
+# The maximum `rust-version` across the RESOLVED graph, not the highest direct
+# dependency: clickhouse 0.15.1, nalgebra 0.35, statrs 0.19.1 and wide 1.7 all
+# require 1.89, so declaring 1.88 would fail inside a dependency instead of
+# saying clearly that this crate needs a newer toolchain. Re-derive it after a
+# bump with `cargo metadata` over the resolved packages' rust_version.
+rust-version = "1.89"
 authors = ["Joaquin Bejar <jb@taunais.com>"]
 description = "OptionChain-Simulator is a lightweight REST API service that simulates an evolving option chain with every request. It is designed for developers building or testing trading systems, backtesters, and visual tools that depend on option data streams but want to avoid relying on live data feeds."
 license = "MIT"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ rust_decimal = { version = "1.42", features = ["maths", "serde"] }
 rust_decimal_macros = "1.40"
 serde_json = "1.0"
 serde = "1.0"
-uuid = { version = "1.23", features = ["v4", "v5", "serde"] }
+uuid = { version = "1.26", features = ["v4", "v5", "serde"] }
 chrono = { version = "0.4", features = ["serde"] }
 # IANA timezone database for the v2 rolling-expiry calendar: the local
 # expiration cutoff and its DST fold/gap resolution (ADR 0001, issue #43)
@@ -67,11 +67,11 @@ chrono-tz = { version = "0.10", features = ["serde"] }
 # Write sink so a multi-year download stays bounded-memory (ADR 0001, issue #49)
 csv = "1.4"
 clickhouse = "0.15"
-tokio = { version = "1.52", features = ["full"] }
+tokio = { version = "1.53", features = ["full"] }
 async-trait = "0.1"
-actix-web = { version = "4.14", features = ["rustls"] }
-actix-files = "0.6"
-redis = { version = "1.3", features = ["tokio-comp", "connection-manager"] }
+actix-web = { version = "4.15", features = ["rustls"] }
+actix-files = "0.7"
+redis = { version = "1.6", features = ["tokio-comp", "connection-manager"] }
 # actix_extras is unavailable: optionstratlib enables utoipa/axum_extras,
 # and utoipa's framework extras are mutually exclusive
 utoipa = "5.5"
--- a/src/infrastructure/config/clickhouse.rs
+++ b/src/infrastructure/config/clickhouse.rs
@@ -1,6 +1,5 @@
 //! This module defines the `ClickHouseConfig` struct and its default implementation.
 use serde::{Deserialize, Serialize};
-use std::env;
 use std::fmt;
 
 /// Configuration structure for connecting to a ClickHouse server.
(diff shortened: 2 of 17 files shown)


joaquinbejar (commented):
REQUEST_CHANGES. The incremental dependency and environment update needs its declared MSRV aligned with the resolved graph, must preserve nonblank credential bytes, and must retain a usable path for passwordless ClickHouse. The cumulative stacked diff is large and should remain split for review.

Cargo.toml
```diff
@@ -2,6 +2,10 @@
 name = "optionchain_simulator"
 version = "0.2.0"
 edition = "2024"
+# Declared because the dependency bump moved it: actix-web 4.15 and redis 1.6
+# both require 1.88. Without this a consumer on an older toolchain gets a
+# failure inside a dependency rather than a clear message about this crate.
+rust-version = "1.88"
```
joaquinbejar: 🟠 Warning: this MSRV is lower than the dependency graph Cargo resolves from these ranges. A clean `cargo +1.88.0 check` fails because ClickHouse 0.15.1, nalgebra 0.35, statrs 0.19.1 and wide 1.7 require Rust 1.89. Declare 1.89 and test it in CI, or constrain the graph to 1.88 compatible releases.

src/utils/env.rs
```diff
@@ -0,0 +1,98 @@
...
+/// The returned value is trimmed, so surrounding whitespace never reaches a
+/// host name, a credential or a parser.
+#[must_use]
+pub fn read_var(variable: &str) -> Option<String> {
+    let raw = std::env::var(variable).ok()?;
+    let trimmed = raw.trim();
+    if trimmed.is_empty() {
+        None
+    } else {
+        Some(trimmed.to_string())
```
joaquinbejar: 🟠 Warning: trimming every nonblank value changes opaque credentials. A password with leading or trailing whitespace now authenticates with different bytes. Use trimming only to detect blank values, return the original value otherwise, and trim individual numeric or host fields where appropriate.

src/infrastructure/config/clickhouse.rs
```diff
@@ -55,17 +54,29 @@ impl Default for ClickHouseConfig {
...
+            // ClickHouse password is no longer expressible through the
+            // environment. `CLICKHOUSE_PASSWORD=` used to be the only way to
+            // say "this user has no password", which is how a stock
+            // ClickHouse `default` user is configured; it now falls back to
+            // the default below. The rule is applied uniformly on purpose —
+            // the alternative is one variable in the whole service behaving
+            // differently from the rest — and a password-less deployment sets
+            // the field directly through `ClickHouseConfig`, which is public.
+            username: super::read_var("CLICKHOUSE_USER").unwrap_or_else(|| "admin".to_string()),
+            password: super::read_var("CLICKHOUSE_PASSWORD")
```
joaquinbejar: 🟠 Warning: an explicit empty password now becomes the literal default `password`. The shipped binary constructs only `ClickHouseConfig::default()`, so the documented programmatic workaround is unavailable and stock passwordless ClickHouse cannot be configured. Preserve an explicit empty credential through a distinct representation or add a real configuration injection path.

Cargo.toml
joaquinbejar: You are right, and cargo metadata over the resolved graph agrees: clickhouse 0.15.1, nalgebra 0.35, statrs 0.19.1 and wide 1.7 all want 1.89, so 1.88 would have failed inside a dependency instead of naming this crate. Declared 1.89 and noted how to re-derive it after a bump.

src/utils/env.rs
joaquinbejar: Fixed by making whitespace decide only whether a value is blank; the value now comes back exactly as written. The numeric and host readers trim at their own call sites, where the meaning is known, so a padded REDIS_PORT still parses while a padded password keeps its bytes.

src/infrastructure/config/clickhouse.rs
joaquinbejar: Agreed, and the workaround I documented did not exist since the binary only ever builds ClickHouseConfig::default(). Credentials of that kind now read through read_secret, which returns a present variable as written, empty included, so CLICKHOUSE_PASSWORD= configures a passwordless server and only an absent variable falls back.

joaquinbejar:
All three are fixed and pushed.

The MSRV now says 1.89, derived from the resolved graph rather than the direct dependencies, and the comment beside it says how to re-derive it after a bump. `read_var` stopped trimming: whitespace decides only whether a value is blank, and the numeric and host readers trim where they parse, so a padded port still works while a padded credential keeps its bytes. And `CLICKHOUSE_PASSWORD=` now configures a passwordless server, through a separate reader that treats a present variable as a value, empty included; the Redis credentials keep the blank-is-unset rule, since an empty Redis password asks for exactly what an absent one does and that empty form is the failure this rule exists to stop.

On splitting the stack: it is already one PR per issue, six in all, each targeting main and reviewable against its own base branch. The diff looks large here only because GitHub shows it against main until the lower ones merge.

Issue #80, opened 2026-08-28
Emit a per-contract spread instead of one constant across the chain


joaquinbejar:
## Overview

A chain is built with one `spread` value applied uniformly to every contract
(`src/domain/factors.rs:909`, `parameters.spread`). Real option markets do not
look like that: a $12 in-the-money call and a $0.05 far wing do not carry the
same absolute bid-ask, and the wing's spread is a far larger fraction of its
price.

This matters more than it used to. A downstream consumer valuing a position at
the touch, buying at the ask and selling at the bid, gets its entire cost of
trading from this number. A single constant makes cheap wings unrealistically
cheap to trade and expensive contracts unrealistically expensive, and it does so
in the direction that flatters defined-risk structures.

Emit a per-contract spread from a parametric microstructure model instead of one
scalar.

## Context

Today `create_chain_data` passes the scalar through to
`optionstratlib::OptionData::apply_spread`, which derives `bid` and `ask` by
adding and subtracting half of it around the mid.

Two properties of that path are the problem:

1. The spread does not vary with the contract. Moneyness, price level, implied
   volatility and days to expiry all move a real spread and none of them is an
   input.
2. `apply_spread` **withdraws the quote entirely** when `mid <= spread`, setting
   `call_ask`, `call_bid` and `call_middle` to `None`
   (`OptionStratLib/src/chains/optiondata.rs:924`, compared against the full
   spread rather than the half). So the cheap wings vanish from the chain exactly
   as they decay. Tracked upstream as joaquinbejar/OptionStratLib#439; that issue
   and this one together are what make a quoted chain realistic: #439 stops the
   wings disappearing, this one gives the remaining quotes a defensible width.

Consumer context: Chainwalk (`gitlab.taunais.com/capitaldelta/chainwalk`) values
every position at the touch and needs a defensible spread on every contract. If
OCS does not supply one, Chainwalk has to model spreads itself, which puts market
microstructure in a client rather than in the market simulator. This issue keeps
that responsibility where it belongs.

## Proposed model

Keep it parametric, small, and documented. A floor-plus-proportional form covers
the common shape:

```
spread(contract) = max(
    tick,
    absolute_floor
      + proportional * mid
      + moneyness_widening * |log(strike / underlying)|
      + tenor_widening * sqrt(days_to_expiry / 365)
)
```

Every coefficient is a create-request field with a documented default, and every
one defaults to a value that reproduces something close to today's behaviour so
[...]

Issue #78, opened 2026-08-28
Serve the export tape in binary: Arrow IPC and a packed columnar format


joaquinbejar:
## Overview

The export endpoint offers two encodings, `json` and `csv` (`Format`,
`src/api/rest/export.rs:111`). Both are text, and both force the consumer to
parse. For the two consumers that matter, that parse is the bottleneck, not the
network and not the simulation.

A browser materialising a whole tape up front pays hundreds of milliseconds of
`JSON.parse` and one allocated object per row for a 500-step by 500-strike
export. With a binary columnar encoding it pays nothing: it maps a
`Float64Array` straight onto the received `ArrayBuffer`. A Rust consumer building
a Parquet tape re-parses text this service already had in typed form.

Add two binary encodings: `arrow` (Arrow IPC stream) and `packed` (a minimal
self-describing columnar block format with no dependencies). They share almost
all of the design; they differ in who they are for.

## Why two

- **`arrow`** is for consumers that already speak Arrow: IronCondor materialises
  simulator tapes to Parquet and already pins `arrow = "59"` and
  `parquet = "59"` (`IronCondor/Cargo.toml:59-60`). For those, Arrow IPC removes
  a whole conversion stage, and it carries schema, types and nulls with no
  private convention to agree on.
- **`packed`** is for the browser. Arrow's JS library is a large dependency for a
  page whose only need is "give me these columns as typed arrays". A minimal
  block format with an aligned layout is a few dozen lines of decoder and zero
  bundle weight, and it is trivially decodable from any language.

They are independently shippable. Land whichever unblocks work first.

## The design constraint that governs everything: this endpoint streams

The handler builds its response with `.streaming(RowStream { receiver })`
(`src/api/rest/export.rs:713-719`) over a bounded channel, and the module doc
(`:87`) is explicit that not materialising the whole tape in memory is the point.
Both new formats must preserve that. Neither may buffer the tape.

Columnar encoding and streaming are in tension: a column cannot be emitted until
its last row is known. The resolution is the same for both formats: **emit in
blocks of a fixed number of rows**, each block columnar within itself. Arrow IPC
already works exactly this way (a stream of record batches); `packed` adopts the
same shape.

## Format A: `arrow`

- Arrow IPC **stream** format (not the file format), one record batch per block
- `Content-Type: application/vnd.apache.arrow.stream`, extension `arrow`
- Schema derived from the existing fixed column lists
[...]

PR #89, opened 2026-08-28
Honour LOGLEVEL instead of hardcoding DEBUG (#81)

### LINKED ISSUE #81: Honour LOGLEVEL instead of hardcoding DEBUG

joaquinbejar:
## What

The service always logged at `DEBUG`: `main` passed the literal `"DEBUG"`, and
`LOGLEVEL` was read **nowhere** in `src/`. `.env.example` documented the
variable, so an operator set it, saw no change, and reasonably concluded the
logging configuration was broken rather than absent.

Closes #81.

The cost is not cosmetic. At `DEBUG` the service emits hyper connection traces on
every request, which for a batch consumer materialising hundreds of tapes buries
anything worth reading and costs real IO.

## Shape

Resolution is a **pure function** in `infrastructure::config::logging`, so the
level can be asserted directly rather than by scraping output — which is what
the issue asks for, and which proves what the level *was* rather than what a
formatter did.

Three details worth naming:

- The level must be known **before** a subscriber exists, and a complaint about
  a bad value can only be logged **after** one does. `ResolvedLogLevel` carries
  both, so `main` installs the subscriber and then says what it ignored, instead
  of swallowing the mistake or aborting over it.
- An unrecognised value never aborts startup. Refusing to boot because a log
  level is misspelled trades a small problem for a total one.
- The effective level is logged **at that level**. An `INFO` line would be
  suppressed by the very setting it reports, and the operator who just set
  `LOGLEVEL=WARN` is exactly the one who wants to see that it took.

Names are case-insensitive; blank is unset, like every other knob. There is
deliberately **no `WARNING` synonym**: upstream's `setup_logger`, which the
example binaries call, has no such arm, so accepting it here would make
`LOGLEVEL=WARNING` mean `WARN` for the service and `INFO` for the examples.

## Tests

Seven, all on the resolved level. Two are load-bearing beyond the obvious:

- **The upstream matcher's arm set is pinned here.** `setup_logger_with_level`
  matches `DEBUG | ERROR | WARN | TRACE` and treats everything else as `INFO`,
  silently — so a renamed variant would not fail anywhere, it would just quietly
  log at `INFO`. Round-tripping through this module's own parser would never
  catch that, because it would agree with itself.
- **The variable name is asserted through the environment.** A typo in the
  constant would pass every other test in the module while reading nothing.

## Reviewed before opening

`architect` audited the diff; its findings are folded in rather than noted. The
sharpest: my doc claimed a blank value resolves "with nothing rejected" while the
code reported it as a rejection, so a direct caller of the public
`resolve_log_level` would have been warned about an empty string. Also moved the
module out of `utils` into `config` where the other knobs live, dropped an
ordering assertion that only proved the enum derives `Ord`, and corrected two
`main.rs` doc blocks still claiming `setup_logger()` is called.

## Checks

- `LOGLEVEL=WARN cargo test --workspace` clean — 808 lib tests, 16 contract
  tests, 3 doctests
- `make readme` regenerated

## Stack

- **Base branch: `stack/1-83-blank-env-is-unset`** — read the diff against that
  branch, not `main`; it is cumulative until #88 merges. It does not compile
  against `main` alone, since it reads `LOGLEVEL` through `utils::env`, which
  #88 introduces.
- **Stacked on: #88.**
- **Stack: #83 → #81 → #82 → #84 → #80 → #78.** This one is #81.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -60,15 +60,15 @@
 
 use optionchain_simulator::api::{ListenOn, start_server};
 use optionchain_simulator::infrastructure::{
-    ClickHouseSnapshotRepository, LogLevel, MetricsCollector, RedisClient, RedisConfig,
-    SimulationV2Config, init_mongodb, resolve_log_level_from_env,
+    ClickHouseSnapshotRepository, MetricsCollector, RedisClient, RedisConfig, SimulationV2Config,
+    init_mongodb, resolve_log_level_from_env,
 };
 use optionchain_simulator::session::{
     InRedisSessionStore, InRedisSimulationStore, SessionManager, SimulationManager,
 };
 use optionstratlib::utils::setup_logger_with_level;
 use std::sync::Arc;
-use tracing::{error, info, warn};
+use tracing::{info, warn};
 
 /// The `main` function is the entry point of the application using the Actix Web server framework.
 /// It initializes the logger, sets up the session management with Redis as the backend, and starts the HTTP server.
@@ -127,14 +127,17 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             "unrecognised LOGLEVEL; using the default"
         );
     }
-    // Emitted AT the resolved level, not at INFO: the operator who just set
-    // `LOGLEVEL=WARN` is exactly the one who wants to see that it took, and an
-    // INFO line would be suppressed by the very setting it reports.
-    match log_level.level {
-        LogLevel::Error => error!(level = %log_level.level, "Log level resolved from LOGLEVEL"),
-        LogLevel::Warn => warn!(level = %log_level.level, "Log level resolved from LOGLEVEL"),
-        _ => info!(level = %log_level.level, "Log level resolved from LOGLEVEL"),
-    }
+    // INFO, whatever the resolved level is. Confirming a healthy configuration
+    // at WARN or ERROR would file an incident on every restart of a service
+    // that is working exactly as configured, and an operator who filters at
+    // those severities is filtering precisely to not see this. Only the
+    // rejected-value branch above is a real problem, and it warns.
+    //
+    // The consequence is deliberate: with `LOGLEVEL=WARN` or above this line is
+    // suppressed. That is the setting doing its job, and the level is
+    // observable without it — every subsequent line is emitted at the level it
+    // resolved to.
+    info!(level = %log_level.level, "Log level resolved from LOGLEVEL");
 
     // Create a session store
     let redis_config = RedisConfig::default();
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,7 +20,8 @@
 //!
 //! # Dependencies
 //! - The `optionchain_simulator` crate is used for infrastructure utilities like Redis client and session store setup.
-//! - `optionstratlib::utils::setup_logger` is used for setting up logging.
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -113,7 +115,26 @@ use tracing::{info, warn};
...
+            value = %rejected,
+            default = %log_level.level,
+            "unrecognised LOGLEVEL; using the default"
+        );
+    }
+    // Emitted AT the resolved level, not at INFO: the operator who just set
+    // `LOGLEVEL=WARN` is exactly the one who wants to see that it took, and an
+    // INFO line would be suppressed by the very setting it reports.
+    match log_level.level {
+        LogLevel::Error => error!(level = %log_level.level, "Log level resolved from LOGLEVEL"),
```
joaquinbejar: 🟠 Warning: a healthy startup with `LOGLEVEL=ERROR` emits an error solely to confirm configuration, creating a false incident on every restart. The WARN branch has the same observability problem. Only invalid configuration belongs at warning or error severity; keep successful confirmation informational or omit it when filtered.

joaquinbejar (commented):
REQUEST_CHANGES. The incremental logging change honors LOGLEVEL but records a healthy startup as an error or warning solely to confirm configuration, which creates false operational incidents. The cumulative stacked diff is large and should remain split for review.

src/main.rs
joaquinbejar: Agreed, a healthy startup should never file an incident. The confirmation is INFO now whatever the level resolves to, and only the rejected-value branch still warns; at WARN or above the line is simply suppressed, which is the setting doing its job.

joaquinbejar:
Fixed and pushed. The resolved level is confirmed at INFO regardless of what it resolved to, so a healthy start under LOGLEVEL=ERROR no longer emits an error on every restart; the unrecognised-value branch keeps its warning, which is the only real problem here. The branch is also rebased onto main now that #88 has merged.

On the stack size: it is already one PR per issue, six of them, each against main and reviewable against its own base. It reads large here only because GitHub diffs it against main until the ones below merge.

PR #90, opened 2026-08-28
Make the HTTP bind address and port configurable (#82)

### LINKED ISSUE #82: Make the bind address and port configurable

joaquinbejar:
Closes #82

## Stack

- **Base branch:** `stack/2-81-honour-loglevel`
- **Stacked on:** #89
- **Blocks:** #92
- **Stack:** #88 (#83) → #89 (#81) → **this** → #84 → #80 → #78

The diff shown against `main` is cumulative until #88 and #89 merge; read it against the base branch. The three lower PRs are logically independent of one another: they are stacked because they touch the same `main.rs` and `Cargo.toml` lines, not because one needs another.

## What changed

The server bound `0.0.0.0:7070` from two constants in `main.rs`, so moving it meant patching the binary or publishing a different host port. `OCS_BIND_ADDRESS` and `OCS_PORT` now decide it, which is what makes it possible to run several instances on one host and shard tape materialisation across them.

Both are validated at startup and an unusable value fails the process naming the variable, matching the v2 and snapshot knobs rather than the request caps that warn and fall back. A cap that falls back still serves requests; a service that quietly ignored `OCS_PORT` would leave two shards fighting over one port. Port `0` is refused with the rest: letting the operating system pick a free port is the opposite of what sharding across known ports wants. Blank is unset, as everywhere else.

The read was also hoisted above the Redis, MongoDB and ClickHouse connects, so a typo in a port costs a refusal rather than three connections and a schema migration first.

## Breaking: the default address is now loopback

It was a hardcoded `0.0.0.0`. This service has no authentication and no rate limiting, so being reachable off the host is now something an operator asks for. **A deployment that relied on the old default must set `OCS_BIND_ADDRESS=0.0.0.0`**; `Docker/docker-compose.yml` does, and the dev override inherits it by merge. Binding beyond loopback logs a warning once at startup.

Both sides of the compose port mapping now read `${OCS_PORT:-7070}`, so moving the port is one edit rather than two that can disagree.

## Semver: the next release is 0.3.0, not 0.2.1

Two breaking changes to the public surface, both on `ListenOn`:

- `ListenOn::as_str` is **removed**. No in-repo callers remained, and IronCondor mirrors the DTOs locally rather than linking the crate, so the blast radius is external crates.io consumers.
- A new `Address(IpAddr)` variant on a public enum breaks any downstream exhaustive `match`. It is now `#[non_exhaustive]`, so this is the last time that particular break costs anything.

Under cargo's 0.x rules both break `^0.2`. No version bump here; that is the release flow's job, but it should be a minor one.

## Layering: `ListenOn` moved down

`ListenOn` was in `api::rest::models` beside the f64 wire mirrors, but it is not a DTO: it never crosses the wire, it is not `Serialize`, and it is read from the environment. Leaving it there while `ServerConfig` read it would have made `infrastructure` import `api`, which is a red in this repo's module boundaries. It now lives in `infrastructure::config::server` next to the configuration that reads it, and `api::ListenOn` still resolves through a re-export so no consumer path changes.

The bind string and the startup log go through `SocketAddr`, so an IPv6 literal comes out as `[::1]:7070` rather than the unusable `::1:7070`.

## Also in this diff

Two pre-existing broken intra-doc links to `greeks::admit_blocking` (which moved to `utils::admission` in #88's predecessor) are repointed at `greeks::admit_render`, so the doc gate stays clean.

## Tests

Four new cases beyond the six on `ServerConfig::from_env`: every variant renders as the address it binds, a resolved configuration produces a usable bind string including the bracketed IPv6 form, only loopback reads as private, and an IPv6 literal parses and stays itself.

## Gate

`make pre-push` clean, MongoDB up: 818 + 16 + 3 tests pass, 22 ignored; clippy with `-D warnings`, `fmt --check`, `cargo doc` with zero warnings and `cargo build --release` all clean; `README.md` regenerated via `make readme`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,7 +19,8 @@
 #
 # Two behaviours to know before editing:
 #
-#   * The v2 knobs (OCS_V2_*, OCS_MAX_CACHED_*) are VALIDATED AT STARTUP. An
+#   * The v2 knobs (OCS_V2_*, OCS_MAX_CACHED_*) and the server knobs
+#     (OCS_BIND_ADDRESS, OCS_PORT) are VALIDATED AT STARTUP. An
 #     invalid value fails the process with a message naming the variable,
 #     because silently reverting to a default would change how long simulations
 #     live or how much memory the service holds without anyone noticing.
@@ -43,6 +44,29 @@
 # a batch consumer materialising hundreds of tapes buries anything worth reading.
 # Values: TRACE | DEBUG | INFO | WARN | ERROR.  Default: INFO
 LOGLEVEL=INFO
+
+
+# ---------------------------------------------------------------------------
+# HTTP server
+# ---------------------------------------------------------------------------
+
+# The interface to bind. Accepts an IP address, or the words `all` (0.0.0.0)
+# and `localhost` (127.0.0.1). Unlike the request caps, an unusable value here
+# FAILS STARTUP naming the variable: a service that quietly ignored it would
+# listen somewhere nobody asked for.
+#
+# The default is loopback, which is a CHANGE: the address used to be a constant
+# 0.0.0.0. This service has no authentication and no rate limiting, so being
+# reachable off the host is now a decision. A container must set 0.0.0.0 for
+# its published port to lead anywhere; Docker/docker-compose.yml does.
+# Default: localhost
+# OCS_BIND_ADDRESS=localhost
+
+# The port to bind, 1-65535. Also fails startup when unusable. 0 is refused
+# along with the rest: the operating system would pick a free port, which is
+# the opposite of what someone running several instances on known ports wants.
+# Default: 7070
+# OCS_PORT=7070
 
 
 # ---------------------------------------------------------------------------
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -31,6 +31,10 @@ WORKDIR /app
 
 COPY --from=builder /app/target/release/optionchain_simulator /app/
 
+# DEFAULT_PORT in src/infrastructure/config/server.rs. This is metadata, not a
+# binding: the compose file publishes ${OCS_PORT} explicitly, but a bare
+# `docker run -P` publishes what is declared here, so a custom OCS_PORT needs an
+# explicit `-p`.
 EXPOSE 7070
 
 CMD ["/app/optionchain_simulator"]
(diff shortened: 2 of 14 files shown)


joaquinbejar (commented):
APPROVE. The incremental bind configuration validates address and port, preserves the public reexport, wires container deployment explicitly, and is covered by tests and documentation. The cumulative stacked diff is large and should remain split for review.

joaquinbejar:
Rebased onto main now that #88 and #89 have merged, with one conflict in the main.rs import list resolved by hand: the LogLevel import went away with the logging fix from #89, and ServerConfig stays. Gate green on the rebased branch and CI is green here.

PR #92, opened 2026-08-28
Add /health and /ready endpoints (#84)

### LINKED ISSUE #84: Add /health and /ready endpoints

joaquinbejar:
Closes #84

## Stack

- **Base branch:** `stack/3-82-configurable-bind`
- **Stacked on:** #90
- **Blocks:** #93
- **Stack:** #88 (#83) → #89 (#81) → #90 (#82) → **this** → #80 → #78

## What changed

`GET /health` answers whether the process is alive: always 200, no dependency touched. `GET /ready` answers whether this instance can take work: Redis and MongoDB always, ClickHouse when snapshot persistence is enabled, 200 when they all answer and 503 naming the ones that did not.

Keeping them apart is the point. A failing dependency must not make `/health` fail, or an orchestrator restarts a healthy instance every time Redis hiccups and one outage becomes two. `/ready` is the signal a consumer sharding tape materialisation across instances reads to route work.

Probes run concurrently under a two-second bound, so one hung dependency can neither hold the probe open nor hide the state of the others. Nothing is cached: an instance whose Redis came back reports ready again without a restart. The only thing remembered between calls is what was last logged, so an outage costs one line instead of one per dependency every ten seconds forever.

## Where the code lives

What is probed sits in `infrastructure::health`, beside the clients it calls; `api::rest::health` keeps the routes, the bodies and the status codes and renders `DependencyReport` into the response DTOs. Putting the probes in `api` would have had that layer reaching into infrastructure's internals (the credential redaction, the driver error types) to do infrastructure work.

The handler holds a list of `DependencyProbe` trait objects rather than the three clients, which is what makes the endpoint testable with no servers running. The list is assembled in `main.rs` from what the process actually opened, so an instance is never reported unready over a service it does not use.

## Breaking

- `start_server` takes a `Readiness`. It is a public function, so the next release is a minor bump.
- `SimulationSnapshotRepository::ping` is **required**, not defaulted. A default `Ok(())` would mean a warehouse that forgot to answer reports itself reachable, and `/ready` would serve a 200 on the strength of it. Four in-crate test doubles implement it in three lines each. Free to tighten now, breaking to tighten later.

## Safety of the body

The 503 body is unauthenticated, so a failure detail is redacted and length-bounded (200 characters) in one place, in `Readiness::evaluate`, rather than in each probe: a contract every implementor has to remember is one the next implementor forgets. A test asserts a password in a driver error does not survive.

`MongoDBClient::ping` clones the `Database` handle and releases the mutex **before** awaiting the command. Held across it, a probe against an unreachable server would pin the handle for the whole timeout and block every `save_step` and `save_event` on the request path behind a health check polled on a timer.

## Metrics

Both routes are excluded through a new `MetricsMiddleware::excluding(&'static [&'static str])`, fed from `health::PROBE_PATHS` so the route strings have one home. Exclusion happens before the timer starts, not at export. A test drives the real middleware and the real collector and reads back `export_metrics`: `/ready` is absent, `/api/v1/chain` is present.

## Docker

`Docker/docker-compose.yml` gains a backend healthcheck hitting `/ready` with busybox wget, which exits non-zero on the 503, plus a 20s `start_period` for the Redis, MongoDB and schema work startup does. `docker-compose.dev.yml` inherits it by merge; verified with `docker compose -f ... -f ... config`. Since the server only binds after those connections succeed, a 200 on `/ready` implies a v2 create can be served.

## Tests

Nine new hermetic tests plus two `#[ignore]`d live ones. Aggregator: ordering, a named failure beside healthy dependencies, recovery with no restart, a hung dependency timing out without holding up the others (on a paused clock), redaction, bounding, and the empty case. Handlers: `/health` 200 while a dependency is down, the 503 body naming it, and recovery flipping the status code. The shipped `WarehouseProbe` is covered against both an unreachable and a local warehouse; the Redis and MongoDB probes have live tests that CI's integration job runs.

## Gate

`make pre-push` clean, MongoDB up: 836 + 16 + 3 tests pass, 24 ignored; clippy with `-D warnings`, `fmt --check`, `cargo doc` with zero warnings and `cargo build --release` all clean; `README.md` regenerated via `make readme`.

### DIFF
--- a/src/infrastructure/health.rs
+++ b/src/infrastructure/health.rs
@@ -19,10 +19,17 @@
 //!
 //! # What reaches the response body
 //!
-//! A probe's explanation is rendered into an unauthenticated 503 body, so it is
-//! redacted and length-bounded HERE, once, rather than in each probe: a
-//! contract that every implementor has to remember is one a future implementor
-//! will forget.
+//! A FIXED CATEGORY, and nothing else. `/ready` is unauthenticated, so a
+//! driver's own words must not reach it: a server message can carry internal
+//! host names, database paths, TLS file paths, query text and tokens in forms
+//! no redaction routine knows to look for, and redacting URL userinfo covers
+//! only the shape it was written for.
+//!
+//! A report therefore carries [`ProbeFailure`], which has exactly two values —
+//! the dependency did not answer, or it did not answer in time — and both are
+//! facts this module owns rather than text it parsed. The driver's explanation
+//! stays inside the process, credential-redacted and length-bounded, in the
+//! transition log.
 
 use crate::infrastructure::config::redact_userinfo;
 use crate::infrastructure::{MongoDBRepository, RedisClient, SimulationSnapshotRepository};
@@ -47,20 +54,45 @@ pub(crate) const PROBE_TIMEOUT: Duration = Duration::from_secs(2);
 /// payload.
 pub(crate) const MAX_DETAIL_CHARS: usize = 200;
 
+/// Why a dependency did not answer, in terms safe to publish.
+///
+/// An allowlist, deliberately small: each value is decided by WHICH branch of
+/// the probe fired, never by inspecting what a server said. Adding a value
+/// means deciding, once and in the open, that the new fact is safe to tell an
+/// unauthenticated caller.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProbeFailure {
+    /// It refused, failed, or could not be reached at all.
+    Unreachable,
+    /// It did not answer inside the probe's bound.
+    TimedOut,
+}
+
+impl ProbeFailure {
+    /// The public wording, which is the whole of what a caller learns.
+    #[must_use]
+    pub fn as_str(self) -> &'static str {
+        match self {
+            ProbeFailure::Unreachable => "unreachable",
+            ProbeFailure::TimedOut => "timed_out",
+        }
+    }
+}
+
 /// One dependency's answer.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct DependencyReport {
     /// What was probed: `redis`, `mongodb` or `clickhouse`.
     pub name: &'static str,
-    /// Why it did not answer, absent when it did. Already redacted and bounded.
-    pub error: Option<String>,
+    /// Why it did not answer, absent when it did. A category, never a message.
+    pub failure: Option<ProbeFailure>,
 }
 
 impl DependencyReport {
     /// Whether this dependency answered.
     #[must_use]
     pub fn is_up(&self) -> bool {
-        self.error.is_none()
+        self.failure.is_none()
     }
 }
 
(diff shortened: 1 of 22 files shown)


src/infrastructure/health.rs
```diff
@@ -0,0 +1,582 @@
...
+    /// Runs every probe at once and collects the answers.
+    ///
+    /// Bounded: a hung dependency costs one timeout for the whole call, not one
+    /// per dependency, and does not stop the others from being reported.
+    pub async fn evaluate(&self) -> Vec<DependencyReport> {
+        let reports = join_all(self.probes.iter().map(|probe| async move {
+            let name = probe.name();
+            let error = match tokio::time::timeout(PROBE_TIMEOUT, probe.check()).await {
+                Ok(Ok(())) => None,
+                Ok(Err(detail)) => Some(bound_detail(&redact_userinfo(&detail))),
```
joaquinbejar: 🟠 Warning: `/ready` returns and logs arbitrary driver or server text after only URL userinfo redaction. That still exposes internal hosts, paths, query parameters, TLS paths, tokens in other forms, and arbitrary server messages through an unauthenticated endpoint. Return a fixed public reason or allowlisted category and keep securely sanitized diagnostics internal.

src/infrastructure/redis/client.rs
```diff
@@ -178,6 +178,24 @@ impl RedisClient {
...
+    /// than the requests it gates, and it must not depend on any key existing.
+    /// The connection manager reconnects on its own, so a probe that failed
+    /// while Redis was down starts succeeding again without a restart.
+    ///
+    /// # Errors
+    ///
+    /// Returns the driver's error when the server is unreachable or does not
+    /// answer within [`RedisConfig::timeout`].
+    #[instrument(skip(self), level = "debug")]
+    pub async fn ping(&self) -> RedisResult<()> {
```
joaquinbejar: 🟠 Warning: this new public method leaks `redis::RedisResult` through the infrastructure adapter, contrary to the repository driver error boundary. The method is needed only by the in crate probe, so make it `pub(crate)` and return `Result<(), ChainError>` after mapping the driver error.

joaquinbejar (commented):
REQUEST_CHANGES. The incremental readiness work returns insufficiently redacted driver details from an unauthenticated endpoint and adds a new public Redis driver error boundary. The cumulative stacked diff is large and should remain split for review.

src/infrastructure/health.rs
joaquinbejar: You are right that userinfo redaction only covers the shape it was written for. The body now carries a fixed category, unreachable or timed_out, chosen by which branch of the probe fired rather than by reading what a server said, and the driver text stays in the transition log where it is still redacted and bounded.

src/infrastructure/redis/client.rs
joaquinbejar: Agreed, that put a driver type on the public surface. It is pub(crate) now and returns Result<(), ChainError>, mapped at the adapter like every other Redis failure, and the in-crate probe is still its only caller.

joaquinbejar:
Both fixed and pushed, and the branch is rebased onto main now that #88, #89 and #90 have merged.

The readiness body says only which dependency failed and whether it was unreachable or timed out; adding a third category would be a deliberate decision that this new fact is safe to tell an unauthenticated caller. A test feeds a probe an error containing a password, an internal address, a TLS path and a database name, then asserts none of them appear anywhere in the rendered report.

PR #93, opened 2026-08-28
Emit a per-contract spread instead of one constant across the chain (#80)

### LINKED ISSUE #80: Emit a per-contract spread instead of one constant across the chain

joaquinbejar:
Closes #80

## Stack

- **Base branch:** `stack/4-84-health-ready`
- **Stacked on:** #92
- **Blocks:** #94
- **Stack:** #88 (#83) → #89 (#81) → #90 (#82) → #92 (#84) → **this** → #78

## The model

```
spread(contract) = max(
    spread_tick,
    spread
      + spread_proportional       * mid
      + spread_moneyness_widening * |ln(strike / underlying)|
      + spread_tenor_widening     * sqrt(days_to_expiration / 365)
)
```

Every coefficient is an optional v2 create-request field with a documented default of zero, echoed back in the simulation response so a run can be replayed. The legacy `spread` is the floor term, so a request that carries only it means what it always meant. The moneyness term is logarithmic because a dollar of distance means something different at a strike of 20 and at a strike of 2000.

Coefficients are bounded at the boundary, not merely non-negative: a `proportional` of 1 already means the spread IS the mid, and unbounded input reaches `Decimal`'s range against a large mid, where multiplication panics. The model's own arithmetic is checked as well, so an unrepresentable width degrades to the tick instead of taking a request down.

## A quote is never withdrawn

Upstream's `apply_spread` sets bid, ask AND mid to `None` when `mid <= spread` (joaquinbejar/OptionStratLib#439, fixed upstream but **not** in the published 0.20 this builds against; verified in the vendored registry source). Cheap wings vanished exactly as they decayed, which is when a consumer most wants to know what closing them costs.

The chain is therefore built with a zero spread, so nothing is erased before the model runs, and the model widens each quote itself. The bid is floored at the tick **or at the mid, whichever is lower**: a contract worth a third of a cent is quoted `0.00 / 0.01` rather than bid a penny it is not worth. Marking worthless wings at a penny would flatter short-wing structures, which is the same bias, in the same direction, that this issue exists to remove. This is a deliberate deviation from the issue's literal wording ("a bid at or above the tick"): every contract worth at least one tick IS bid at least one tick, and a test asserts it.

## Two things a legacy request sees change

Both are consequences of the erasure fix rather than of the model:

1. **The chain is longer.** Upstream stops extending the strike ladder at the first pair of strikes with no price, and that condition WAS the erasure. With nothing erased the ladder runs to the full `2 * chain_size + 1`. Measured at spot 5000, vol 0.2: 30 DTE with `chain_size = 30` goes from 45 strikes to 61; 1 DTE with `chain_size = 40` from 65 to 81. Response payloads, export rows and warehouse rows grow with it, inside the same configured caps, since `validate_snapshot_work` already bounds on the theoretical grid.
2. **A spread below the tick is raised to it.** `spread: 0.005` now widens by half a cent each way rather than a quarter. At or above the tick, which is every request that took the documented default, quotes are unchanged term for term.

The issue's "byte-identical for a legacy scalar" criterion therefore holds for the quotes but not for the strike set, and that is stated in the crate docs, the module docs and here rather than glossed.

## Snapshot generation → 3

The same `(simulation, generation, step)` coordinate now produces different values and a different strike set, so `CURRENT_SNAPSHOT_GENERATION` goes to `3`. Rows an older binary filed stay addressable under `2` and are never half-overwritten by a replay that no longer agrees with them.

## Two scope decisions

**No new export columns and no per-contract `spread` field.** The issue asks for "the per-contract spread, **or** the bid and ask it produced". Both surfaces already carry bid and ask, and the spread is exactly `ask - bid`, recoverable losslessly. Adding a column would also change the export header that #78 is about to extend, and break the property that each greek level's header is a prefix of the next.

**v1 is untouched.** `/api/v1/chain` is frozen on rendered values, not only shapes (ADR §12.1), so changing its spread model would be a violation rather than merely out of scope. Its chains also come from upstream's `generator_optionchain`, not from this path.

## Tests

Nine model-level tests (defaults reproduce the old scalar, the legacy scalar is the floor, each widening term behaves, the tick floors, a sub-tick mid is not bid above itself, an absurd coefficient degrades rather than panicking) and six chain-level ones. The two erasure tests assert on the **count** of quoted contracts against the count worth more than the whole spread, because "every contract with a mid has a bid" passes on the old code too: the old code cleared the mid alongside the quote.

Also covered: the legacy-scalar quotes are upstream's arithmetic term for term, the model is deterministic across two builds, a stored simulation round-trips **with** the model set, a stored document without it still loads, a stored zero tick is refused on load, and the HTTP surface accepts and echoes the four coefficients and rejects an out-of-range one as a typed 400.

Both required reviewers ran over this diff: `reproducibility-reviewer` (walk path untouched, model pure, `BTreeSet` rebuild order-preserving, old documents load) and `architect`.

## Gate

`make pre-push` clean, MongoDB up: 859 + 16 + 3 tests pass, 24 ignored; clippy with `-D warnings`, `fmt --check`, `cargo doc` with zero warnings and `cargo build --release` all clean; `README.md` regenerated via `make readme`. ADR 0001 §8, §14.1, §14.2 and `doc/requests/create_simulation_v2.json` updated, since the coefficients are replay inputs and §8 says its list is exhaustive.

### DIFF
--- a/src/domain/spread.rs
+++ b/src/domain/spread.rs
@@ -1,0 +1,607 @@
+//! What a quote costs to cross, per contract.
+//!
+//! A chain used to be built with ONE spread applied to every contract, which is
+//! not what an option market looks like: a twelve-dollar in-the-money call and
+//! a five-cent far wing do not carry the same absolute bid-ask, and the wing's
+//! spread is a far larger fraction of its price. A consumer valuing a position
+//! at the touch gets its entire cost of trading from this number, so a single
+//! constant makes cheap wings unrealistically cheap to trade and expensive
+//! contracts unrealistically expensive — and it does so in the direction that
+//! flatters defined-risk structures.
+//!
+//! # The model
+//!
+//! Parametric, small, and every coefficient a create-request field:
+//!
+//! ```text
+//! spread(contract) = max(
+//!     tick,
+//!     absolute_floor
+//!       + proportional        * mid
+//!       + moneyness_widening  * |ln(strike / underlying)|
+//!       + tenor_widening      * sqrt(days_to_expiration / 365)
+//! )
+//! ```
+//!
+//! Every widening term defaults to zero, so a request that says nothing about
+//! spreads gets `max(tick, absolute_floor)` on every contract, which is the
+//! single scalar the service applied before. The legacy `spread` field is
+//! exactly the floor term, so it keeps meaning what it meant.
+//!
+//! # A quote is never withdrawn
+//!
+//! Upstream's `OptionData::apply_spread` sets bid, ask AND mid to `None` when
+//! `mid <= spread`, so cheap wings vanish from the chain exactly as they decay
+//! — the moment a consumer most wants to see what closing them costs. That is
+//! tracked upstream (joaquinbejar/OptionStratLib#439) and fixed on their main
+//! branch, but the published 0.20 this crate builds against still erases.
+//!
+//! This module therefore does the widening itself, from a chain built with a
+//! zero spread so nothing is erased before it gets here, and floors the bid at
+//! the tick or at the mid, whichever is lower: a contract that has a mid has a
+//! bid and an ask. For every quote upstream would have kept, the arithmetic is
+//! upstream's, term for term.
+//!
+//! # What a legacy request sees change
+//!
+//! Two things, both consequences of the erasure fix rather than of the model:
+//!
+//! - **The chain is longer.** Upstream stops extending the strike grid at the
+//!   first pair of strikes with no price (`chain.rs`, `some_price_is_none`),
+//!   and that condition was the erasure. With nothing erased the ladder runs to
+//!   the full `2 * chain_size + 1`: at spot 5000, 30 DTE and `chain_size = 30`,
+//!   45 strikes become 61. Response payloads, export rows and warehouse rows
+//!   grow with it, within the same configured caps.
+//! - **A spread below the tick is raised to it.** `spread: 0.005` used to widen
+//!   by a quarter cent each way; it now widens by half a cent. For any spread
+//!   at or above the tick — every request that took the documented default —
+//!   the quotes are unchanged.
+//!
+//! One residual erasure is upstream's and stays: a mid of exactly zero, which
+//! happens when the pricing CDF underflows, is still dropped by
+//! `apply_spread`. Such a contract has no mid either, so nothing is quoted
+//! one-sided; it simply is not in the chain.
+
+use crate::session::SimulationParametersV2;
+use optionstratlib::chains::OptionData;
+use optionstratlib::chains::chain::OptionChain;
+use positive::Positive;
+use rust_decimal::{Decimal, MathematicalOps};
+use rust_decimal_macros::dec;
+use std::sync::LazyLock;
+
+/// The spread applied when a request says nothing: one cent.
+///
+/// The same value the service applied before the model existed, so silence
+/// means what it always meant.
+pub(crate) const DEFAULT_SPREAD_FLOOR: Decimal = dec!(0.01);
+
+/// The smallest quotable increment, and the lowest a bid may go.
+///
+/// A penny, matching the two decimal places every quote in this service is
+/// rounded to. A bid below it would not be a price anyone could act on.
+pub(crate) const DEFAULT_TICK: Decimal = dec!(0.01);
+
+/// [`DEFAULT_TICK`] as a `Positive`, built once.
+///
+/// Built here rather than at every call site so the conversion has exactly one
+/// fallback, and so that fallback is not a value 100 times too large: a tick of
+/// one dollar under every bid would be far worse than the penny it replaces.
+static DEFAULT_TICK_POSITIVE: LazyLock<Positive> =
+    LazyLock::new(|| Positive::new_decimal(DEFAULT_TICK).unwrap_or(Positive::ZERO));
+
+/// The widest each coefficient may be set to.
+///
+/// Economic, not arbitrary. A `proportional` of 1 means the spread IS the mid,
+/// which is already a market nobody can trade; the two widening rates reach the
+/// same absurdity long before 10, since `|ln(K/S)|` is about 0.7 at a strike
+/// twice the spot and `sqrt(years)` is 1 at a year. Bounding them also keeps
+/// the arithmetic below `Decimal`'s range for any price a chain can carry.
+pub(crate) const MAX_PROPORTIONAL: f64 = 1.0;
+
+/// The widest the two widening rates may be set to. See [`MAX_PROPORTIONAL`].
+pub(crate) const MAX_WIDENING: f64 = 10.0;
+
+/// The largest quotable increment. A tick above this is not a tick.
+pub(crate) const MAX_TICK: f64 = 10.0;
+
+/// Days in the year used to scale the tenor term.
+const DAYS_PER_YEAR: Decimal = dec!(365);
+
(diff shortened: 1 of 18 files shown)


src/session/model_v2.rs
```diff
@@ -377,6 +409,27 @@ impl SimulationParametersV2 {
...
+        // refuses — would narrow a quote away from the money.
+        if let Some(tick) = self.spread_tick {
+            reject_zero("spread_tick", tick)?;
+        }
+        for (field, value) in [
+            ("spread_proportional", self.spread_proportional),
+            ("spread_moneyness_widening", self.spread_moneyness_widening),
+            ("spread_tenor_widening", self.spread_tenor_widening),
+        ] {
+            if let Some(value) = value
```
joaquinbejar: 🟠 Warning: stored document validation rejects negative coefficients and a zero tick but never reapplies `MAX_PROPORTIONAL`, `MAX_WIDENING`, or `MAX_TICK`. Redis JSON can therefore deserialize into a state the REST boundary rejects. Enforce the same upper bounds here and add over cap stored JSON rejection tests.

src/domain/spread.rs
```diff
@@ -0,0 +1,607 @@
...
+    ///
+    /// The bid is floored at the tick OR at the mid, whichever is lower. The
+    /// distinction matters: flooring unconditionally at the tick would let a
+    /// contract worth a third of a cent be sold for a penny, which flatters
+    /// short-wing structures — the exact bias this model exists to remove. A
+    /// contract worth less than a tick is therefore quoted `0.00 / 0.01`:
+    /// present, and worth what it is worth.
+    #[must_use]
+    fn quote_around(&self, mid: Positive, spread: Positive) -> (Positive, Positive) {
+        let half = spread.to_dec() / Decimal::TWO;
```
joaquinbejar: 🟠 Warning: the new monetary path is not checked throughout. `spread / 2` here, `days / 365` above, and `mid - half` below use raw `Decimal` operators, while the mandatory arithmetic rules require checked variants and an explicit division rounding policy. Use `checked_div` and `checked_sub`, document rounding, and map failure deliberately.

joaquinbejar (commented):
REQUEST_CHANGES. The incremental spread model accepts stored parameters above the REST caps and uses unchecked Decimal arithmetic on the pricing path, so persistence does not enforce the request invariants and mandatory monetary safety rules are bypassed. The cumulative stacked diff is large and should remain split for review.

src/session/model_v2.rs
joaquinbejar: Caught, thank you: the stored path checked only the lower bounds. It now applies MAX_PROPORTIONAL, MAX_WIDENING and MAX_TICK from the same constants the request path uses, with a test per field for an over-cap document and one more asserting a document exactly at the cap still loads.

src/domain/spread.rs
joaquinbejar: Right, three raw operators were left on the pricing path. spread / 2, mid - half and days / 365 are all checked now, the rounding policy is stated where it applies, and a failure degrades to the tick or to the untouched mid; a test drives a mid of Decimal::MAX through the quote to prove it does not panic.

joaquinbejar:
Both fixed and pushed, and the branch is rebased onto main now that the four below have merged. Gate green: 864 + 16 + 3 tests, clippy and fmt clean, CI green here.

Worth flagging again since it is the one thing in this PR that changes what clients see beyond the quotes: with the erasure fixed, upstream no longer stops extending the strike ladder where the wings used to vanish, so a chain runs to the full grid. At spot 5000, 30 days and chain_size 30 that is 45 strikes becoming 61, and export and warehouse rows grow with it inside the existing caps. It is in the crate docs and the commit message, and CURRENT_SNAPSHOT_GENERATION goes to 3 so older persisted tapes stay addressable.

PR #94, opened 2026-08-28
Serve the export tape in binary: Arrow IPC and a packed columnar format (#78)

### LINKED ISSUE #78: Serve the export tape in binary: Arrow IPC and a packed columnar format

joaquinbejar:
Closes #78

## Stack

- **Base branch:** `stack/5-80-per-contract-spread`
- **Stacked on:** #93
- **Stack:** #88 (#83) → #89 (#81) → #90 (#82) → #92 (#84) → #93 (#80) → **this** (top)

## What changed

Two binary encodings beside `json` and `csv`, both streaming in blocks of `OCS_EXPORT_BLOCK_ROWS` (default 4096, capped at 262 144), both carrying the same column names in the same order as the CSV header and `f64` for every number. Binary is a faster route to the same values, **not** a route to the underlying `Decimal(38, 28)` precision, and the endpoint docs say so.

- **`arrow`** — Arrow IPC stream, one record batch per block, `application/vnd.apache.arrow.stream`, extension `arrow`. Behind the `arrow-export` feature, off by default, `arrow = "59.2"` with `default-features = false` to match IronCondor.
- **`packed`** — no dependencies, `application/octet-stream`, extension `ocsp`. 8-byte aligned payloads so a browser can map a `Float64Array` onto the buffer with no copy; Arrow-convention validity bitmaps so one decoder serves both.

## Design decisions worth reviewing

**`labels` as a bitmask, not a dictionary-encoded string.** The issue sketched a dictionary for `symbol` and left `labels` open. Its value space is the power set of the rule ids, so it cannot be dictionary-encoded up front; a `u64` mask over the header's rule ids can, at a fixed width, and joining the bits it sets reproduces the CSV column character for character because both orders are lexicographic. The mask starts **after** the dictionary's first entry, which is the symbol: both charsets allow a rule id equal to the symbol, and without the skip that rule id would take bit 0 and come back ahead of every other label. A test pins exactly that collision.

**A footer.** `packed` closes with a sentinel row count and the total rows written. Without it a connection dropped mid-stream leaves a shorter document that decodes as a perfectly valid smaller tape, under a 200 whose header went out before the first byte existed. The decoder refuses a document with no footer, and refuses one whose count disagrees.

**The `Writer` enum grew two arms** rather than becoming an encoder trait, which the issue listed as a task. The arrow arm needs `cfg`, and an enum keeps the hot path unboxed; the four encodings already sit behind one streaming producer, which is what the task was for.

**No new export columns.** `#80` (the PR below this one) deliberately added none either, so the header lists are the ones this PR reads.

## Acceptance criteria

| Criterion | Where |
|---|---|
| `arrow` and `packed` decode to the `json` values, every dataset | `test_a_packed_export_decodes_to_the_csv_values`, `test_an_arrow_export_decodes_to_the_csv_values` — both compare value by value against CSV, which `test_the_json_export_matches_the_csv_export_at_every_level` already pins against JSON |
| Byte-identical on repeat, all four formats | `test_every_format_is_byte_identical_on_repeat` (arrow included when the feature is on) |
| `from_step` / `to_step` identical across formats, errors included | `test_a_step_range_behaves_the_same_in_packed`, `..._in_arrow` |
| Memory flat in the number of steps | `test_the_writer_never_buffers_more_than_one_block` asserts the buffer stays under one block across 3 500 rows; `test_a_multi_block_packed_export_carries_every_row` drives a real dataset at a two-row block width and compares it to the same export at 4 096 |
| Every payload offset 8-byte aligned | `test_every_payload_offset_is_aligned` walks the produced block the way a decoder does and asserts the walk consumes it exactly |
| A null round-trips as a null, not a zero or a NaN | `test_a_null_is_a_cleared_bit_rather_than_a_sentinel`, `test_a_null_survives_the_packed_encoding`, and the Arrow round-trip test |
| Typed 400 for `arrow` without the feature | `test_arrow_without_the_feature_is_a_typed_400`, checked in the handler before streaming |
| Default build does not link `arrow` | `cargo tree --no-default-features \| grep -c arrow` = **0**; with the feature = **50** |
| Block size configurable, documented | `OCS_EXPORT_BLOCK_ROWS` in `limits.rs`, `.env.example`, the crate docs, and `test_default_limits_match_documentation` |
| Layout written down | `binary.rs` module docs, `src/lib.rs` / `README.md`, and the utoipa `format` description |

Also added: a test that every dataset × greek level produces rows whose cell types are the ones `BinarySchema` declares. That schema assigns types by matching column **names** with an `f64` fallback, so a column added to the header lists lands on `f64` silently, and a mismatch would desynchronise a block rather than panic.

## Measurements

`option_chains` at `greeks=all` on the reference fixture, release build:

| Format | Bytes | Wall time |
|---|---|---|
| `json` | 79 888 | 9.7 ms |
| `csv` | 46 819 | 7.8 ms |
| `arrow` | 29 640 | 6.5 ms |
| `packed` | 22 808 | 6.1 ms |

Reproduce with `cargo test --release --all-features -- --ignored --nocapture test_measure_every_format`.

## CI

The Tests job now runs twice, once without features and once with `arrow-export`: the default build ships without the feature, and the cross-format equality tests for `arrow` only exist when it is on. Lint already covered the feature through `--all-features`.

## Gate

`make pre-push` clean, MongoDB up: **881** + 16 + 3 tests pass without the feature, **886** with it; clippy with `-D warnings` over `--all-features`, `fmt --check`, `cargo doc` with zero warnings and `cargo build --release` all clean; `README.md` regenerated via `make readme`.

### DIFF
--- a/src/api/rest/binary.rs
+++ b/src/api/rest/binary.rs
@@ -44,16 +44,26 @@
 //! requirement rather than a nicety.
 //!
 //! ```text
-//! file        := header block*
+//! file        := header block* footer
 //! header      := "OCSP" u32:version u32:block_rows
 //!                u32:dictionary_count dictionary_entry*
 //!                u32:column_count column_desc*
 //!                pad to 8
-//! dict_entry  := u32:len utf8:value            (the rule ids, sorted)
+//! dict_entry  := u32:len utf8:value            (the symbol, then the rule ids)
 //! column_desc := u32:name_len utf8:name u8:type_code u8:nullable pad to 4
 //! block       := u32:row_count pad to 8 column_payload*
 //! payload     := [validity bitmap if nullable, padded to 8] values, padded to 8
+//! footer      := u32:0xFFFFFFFF pad to 8 u64:total_rows
 //! ```
+//!
+//! **The footer is required, and a decoder must check it.** A block's row count
+//! can never be `0xFFFFFFFF`, so that value is what tells a reader the blocks
+//! have ended; the `u64` after it is the total number of rows the writer
+//! emitted. A document that ends without the sentinel was TRUNCATED and must be
+//! rejected rather than read as a shorter tape — the response is a 200 whose
+//! header goes out before the first byte is produced, so a dropped connection
+//! is otherwise indistinguishable from a smaller export. A row count that
+//! disagrees with the blocks is the same error.
 //!
 //! Type codes: `0` = `f64`, `1` = `i64`, `2` = timestamp in nanoseconds since
 //! the epoch (`i64`), `3` = a dictionary index into the header's entries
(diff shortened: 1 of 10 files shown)


src/api/rest/export.rs
```diff
@@ -1130,6 +1287,49 @@ impl Writer {
...
+                let rows = typed_rows(
+                    writer.schema(),
+                    *dataset,
+                    *level,
+                    step,
+                    row.simulated_at,
+                    row,
+                    chains,
+                )?;
+                for block in writer.push(rows)? {
```
joaquinbejar: 🟠 Warning: `typed_rows` materializes the entire step, then `push` encodes and returns every completed block in a `Vec` before this loop can emit or observe cancellation. At the 200000 contract step cap, `OCS_EXPORT_BLOCK_ROWS` therefore does not bound memory or cancellation latency, and Arrow follows the same pattern. Feed rows incrementally and emit each completed block immediately.

src/api/rest/binary.rs
```diff
@@ -0,0 +1,1197 @@
...
+//! # The `packed` layout
+//!
+//! Little-endian throughout. Every payload starts at an 8-byte aligned offset,
+//! which is the whole point: it is what lets a browser do
+//! `new Float64Array(buffer, offset, count)` with no copy at all. An unaligned
+//! offset makes that constructor throw, so the alignment is a correctness
+//! requirement rather than a nicety.
+//!
+//! ```text
+//! file        := header block*
```
joaquinbejar: 🟠 Warning: the writer always appends a footer containing `u32::MAX`, alignment padding, and a `u64` total row count, but the public grammar says only `header block*`. A decoder implemented from these docs can treat the sentinel as a gigantic block or reject valid trailing bytes. Document `file := header block* footer`, the exact footer layout, and the required truncation check here and in the generated README.

joaquinbejar (commented):
REQUEST_CHANGES. The incremental binary export still materializes and encodes all blocks for a step before backpressure or cancellation applies, and the public packed grammar omits the required footer. The cumulative stacked diff is large and should remain split for review.

src/api/rest/export.rs
joaquinbejar: Correct, and it defeated the point of the block width. Rows are visited one at a time now and each finished block goes straight to emit, in both encodings; two tests pin it, one counting several blocks from a single step at a width of two, the other asserting a gone client stops the step at the first refused block instead of encoding the rest.

src/api/rest/binary.rs
joaquinbejar: Fixed: the grammar says header block* footer, gives the exact footer layout, and states that a document without the sentinel was truncated and must be rejected rather than read as a shorter tape. Same wording in the module docs, the crate docs and the regenerated README.

joaquinbejar:
Both fixed and pushed, and the branch is rebased onto main now that the five below have merged.

One extra commit on top: CI's clippy refused the eight-parameter visitor the first fix introduced, so a step's row inputs are grouped into a RowContext and the test-only collector is written over the visitor rather than duplicating the row builder. Gate green in both configurations, 888 without the feature and 893 with it, and all four checks pass here.

Issue #97, opened 2026-08-30
Re-sync the mirrored walk kernels with optionstratlib 0.21's hardening


joaquinbejar:
## Overview

`Walker` mirrors optionstratlib's stochastic kernels so the walk can be seeded, and CLAUDE.md makes re-diffing them an obligation of every upstream bump. The bump to 0.21 was reviewed for VALUES and is clean: upstream's panic sweep (#442) rewrote `garch`, `heston`, `custom`, `telegraph`, `brownian`, `geometric_brownian`, `log_returns`, `mean_reverting` and `jump_diffusion` onto checked helpers while preserving operand order and rounding, `decimal_normal_sample` is now the same form `Walker::normal_sample` already used, and the trait method list is byte-identical. The tape does not move, and the suite's golden numbers confirm it.

What the mirror did NOT take is the hardening.

## What is missing

- `src/domain/walker.rs:197` — heston's `for _ in 0..params.size - 1` underflows at `size == 0`, where 0.21 saturates
- `src/domain/walker.rs:258` — custom's `take(params.size - 1)` has the same shape
- The raw `+` / `*` operators upstream replaced with checked ones on those paths

None of it is reachable through the REST surface today, since a simulation is validated to at least one step, but the mirror's whole point is that it stays a mirror.

## Tasks

- [ ] Re-diff the mirrored kernels against 0.21 and take the hardening, preserving the documented intentional divergence (the `jump_diffusion` Bernoulli fix, issue #11)
- [ ] Add a test for a zero-size walk on the two kernels that underflow

## Acceptance criteria

- [ ] Same parameters and seed still yield an identical snapshot tape; the existing tape tests are unchanged and still pass
- [ ] No arithmetic on the walk path can underflow or overflow
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all --check` and `LOGLEVEL=WARN cargo test --workspace` clean

Created from the reproducibility review of #91.

Issue #96, opened 2026-08-30
Bound the widening a pinned ladder asks for against the validated snapshot budget


joaquinbejar:
## Overview

A pinned simulation serves its ladder by asking upstream for a chain wide enough to reach the pinned strikes from the current spot, then filtering. The width therefore grows as the underlying drifts, while `validate_snapshot_work` bounded the work at creation using the DECLARED `chain_size`.

A simulation created with `chain_size: 6` can end up pricing far more strikes per expiration per step and discarding most of them. The output stays inside the cap; the work does not.

## Current bound

`MAX_PINNED_WIDTH` in `src/domain/ladder.rs` caps it at 500 strikes per side, and a spot that drifts past it is a typed 400 naming `strike_ladder`. That stops the unbounded case but does not tie the widening to the budget the client was validated under.

## Tasks

- [ ] Derive the ceiling from the same budget `validate_snapshot_work` applies, rather than from a separate constant
- [ ] Refuse with a validation-shaped error that names the budget, as the creation path does
- [ ] Document the cost of a pinned ladder in the endpoint description: what the service prices to serve it, and how far a spot may drift

## Acceptance criteria

- [ ] A pinned simulation cannot make the service price more contracts per step than its declared configuration was validated for
- [ ] The refusal names the bound and what to change
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all --check` and `LOGLEVEL=WARN cargo test --workspace` clean

Created from the review of #91.

PR #98, opened 2026-08-30
Offer a pinned strike ladder so a position keeps its contracts (#91)

### LINKED ISSUE #91: Offer a pinned strike ladder so a position keeps its contracts
## Overview

The strike ladder is rebuilt around the current underlying at every step, so the
set of quoted contracts follows the spot. A contract quoted at step 0 can be
absent at step 2. Measured against a live v2 simulation:

```
step 0   spot 5000.00   13 strikes   [4850 .. 5150]
step 1   spot 5004.95   13 strikes   [4850 .. 5150]
step 2   spot 5013.54   13 strikes   [4875 .. 5175]    + 5175   - 4850
```

A spot move of **0.27 percent over two steps** was enough to drop the 4850
strike from the chain entirely.

`chain_size` fixes how many strikes are quoted; it does not fix *which*. That is
the problem: a client holding a position needs the instruments it opened to keep
existing for as long as it holds them.

## Why this matters

For a client browsing a chain, a ladder that tracks the spot is the right
behaviour and probably what one wants.

For any client that **holds a position across steps**, it is not. A leg opened at
4850 cannot be marked at step 2, cannot be closed, and cannot be settled, because
the contract is not in the data. The position has a hole in it, and the hole
opens as a function of how far the underlying moved rather than of anything the
strategy did.

Over the 500 steps of the reference scenario in ADR 0001 this is the normal case,
not an edge case. A defined-risk structure whose long wings sit furthest from the
money loses exactly those wings first, and they are the legs that cap the risk.

Consumer context: Chainwalk
(`gitlab.taunais.com/capitaldelta/chainwalk`) opens a multi-leg position at one
step and marks it at every subsequent step. It currently handles this by carrying
an explicit not-quoted state per leg, which is honest but degraded: the user sees
a position that cannot be valued through no fault of their own.

## Reproduction

```bash
# create with a narrow chain so the effect shows within a few steps
POST /api/v2/simulations
  { "symbol": "SPX", "steps": 500, "seed": 42,
    "start_at": "2026-01-05T14:30:00Z", "step_interval_seconds": 86400,
    "timezone": "America/New_York", "calendar": "weekdays_v1",
    "expiration_time": "17:00",
    "schedules": [{ "rule_id": "weeklies", "kind": "weekly",
                    "target_count": 1, "weekdays": ["Fri"] }],
    "initial_price": 5000.0, "volatility": 0.18, "risk_free_rate": 0.04,
    "dividend_yield": 0.012, "chain_size": 6, "strike_interval": 25.0,
    "skew_slope": -0.2, "smile_curve": 0.4, "spread": 0.0,
    "method": { "GeometricBrownian": { "dt": 0.004, "drift": 0.05,
                                       "volatility": 0.18 } },
    "time_frame": "Day" }

GET /api/v2/simulations/{id}/export?dataset=option_chains&format=csv&from_step=0&to_step=2
```

Group the exported rows by `step` and compare the strike sets.

## Proposal

**Pin the contract universe at creation.** The ladder is computed once, from
`initial_price`, `chain_size` and `strike_interval`, and every step of that
simulation quotes that same set of strikes. A simulation is a closed world for
its whole life, which is what makes a position in it well defined.

This is additive and back-compatible if it is opt-in. Suggested shape: a
`strike_ladder` field on the create request with values `pinned` and `rolling`,
defaulting to `rolling` so nothing changes for existing clients, and echoed in
the simulation response like every other effective parameter.

Two details worth settling in the issue rather than in review:

- **A pinned ladder does not follow a large move.** If the spot leaves the pinned
  range the chain is all calls or all puts, which is correct and informative: a
  real chain does get listed further out, but a simulation that silently invented
  new strikes would break the closed-world property this issue is asking for.
  Consider widening the pinned ladder at creation rather than at step time.
- **The interaction with the rolling expiry calendar.** Expirations already roll
  by design and that should not change; this is about strikes only.

## Tasks

- [ ] Add `strike_ladder` to the v2 create request, `rolling` by default, echoed
      in the response
- [ ] When `pinned`, compute the ladder once at creation from `initial_price` and
      reuse it for every step
- [ ] Record the resolved ladder in the stored simulation so a replay reproduces
      it
- [ ] Document the trade-off in the endpoint description: a pinned ladder gives a
      stable contract universe, a rolling one keeps strikes near the money
- [ ] Update the utoipa annotations and the OpenAPI document

## Acceptance criteria

- [ ] With `strike_ladder: "pinned"`, the strike set of the exported chain is
      identical at every step of a 500-step simulation whose spot moves
      substantially; a test asserts it over a seed chosen to move the underlying
- [ ] With `strike_ladder` absent or `rolling`, the export is byte-identical to
      what the same request produces today
- [ ] The effective value is echoed in the simulation response and survives a
      store round trip
      snapshot tape

joaquinbejar:
Closes #91

## What changed

The ladder is rebuilt around the underlying at every step, so a contract quoted at step 0 can be gone by step 2: a move of 0.27 percent over two steps was enough to drop a strike. `chain_size` fixes how MANY strikes are quoted, never WHICH.

`strike_ladder` on the v2 create request now chooses. `rolling` is the default and is what the service always did. `pinned` fixes the ladder at creation from `initial_price`, `chain_size` and `strike_interval`, so a contract quoted once is quoted for the simulation's whole life, which is what a client holding a position across steps needs.

Serving it costs no hand-rolled pricing. Upstream anchors every ladder at `rounder(underlying, interval)`, which snaps to a multiple of the interval, so every ladder it builds lies on one grid; a pinned step asks upstream for a chain wide enough to reach the pinned strikes from wherever the spot is now and keeps those out of it.

## It needed an upstream fix

The blocker was not in this repo. Until optionstratlib 0.21.1 a wing worth nothing was priced as **absent** rather than as zero, and `build_chain` reads a missing side as "this wing does not quote", so the ladder stopped before the far pinned strikes and there was nothing to filter to. Fixed in [OptionStratLib#488](https://github.com/joaquinbejar/OptionStratLib/pull/488) for [#487](https://github.com/joaquinbejar/OptionStratLib/issues/487), which this PR pins.

That fix also changes what **rolling** simulations serve near expiry: measured with this crate's defaults at a spot of 5100 with a 25-point interval, 26 strikes instead of 41 at 0.3125 days and 10 instead of 21 at 0.05 days. So `CURRENT_SNAPSHOT_GENERATION` goes to 4.

## Acceptance criteria

| Criterion | Status |
|---|---|
| Pinned: the strike set is identical at every step of a simulation whose spot moves substantially | Met. The test asserts every step against the ladder the simulation RESOLVED, not against its own first step, and the fixture must move the spot past half the ladder's width for the assertion to count |
| `strike_ladder` absent or `rolling`: the export is byte-identical to today | **Waived, and not by this PR's code.** Rolling is untouched here, but the optionstratlib bump changes chains near expiry, which is the whole point of taking it. Worth knowing for Chainwalk and IronCondor: a stored generation-3 tape is not comparable |
| The effective value is echoed and survives a store round trip | Met, at the wire level and the stored level |
| Reproducibility holds | Met. The ladder is a pure function of stored parameters and cannot touch a price path; a pinned run reproduces its tape under the same seed |
| clippy, fmt, tests, `make readme` | Met, below |

## Review findings, all fixed

`reproducibility-reviewer` and `architect` both ran over this and converged on two criticals:

- **Panicking arithmetic on a request path.** `strike_interval` has no upper bound, so `{"strike_interval": 1e27, "chain_size": 500}` reached `Positive arithmetic overflow in mul_decimal`. Everything is checked now, and the anchor mirrors upstream's comparison form rather than an algebraically-equal one that parts company at 28 decimal places.
- **A silently partial ladder.** Upstream stops extending downwards past the anchor, so a ladder wider than the spot loses its lowest strikes: at an initial price of 100 with a 5-point interval, `chain_size: 25` served 40 of 45 strikes at step 0 with no error. That combination is refused at creation naming `chain_size`, and a chain that still comes back short is an error naming what it lacks.

Also fixed: `build_chain` no longer imports the api layer (the bound is the domain's own, so `OCS_MAX_CHAIN_SIZE` cannot decide whether a running simulation serves its next step); `StrikeLadder` is re-exported from `session` so a consumer can name it; the OpenAPI document carries the enum instead of a free-form string; a spot beyond the widening bound is a typed 400 instead of an opaque 500; `strike_ladder` joins ADR 0001 §8's replay list and both example bodies; and two stale claims in the docs are corrected, including "all calls or all puts", which was never true of a response where every contract carries both sides.

Two follow-ups opened rather than folded in: #96 (bound the widening against the validated snapshot budget) and #97 (take 0.21's hardening into the mirrored walk kernels — the values are already verified identical).

## For Chainwalk

`strike_ladder` is additive on the wire and defaults to `rolling`, so nothing changes without asking. Two things to know: adding a field to the public `CreateSimulationRequest` breaks a Rust struct literal, free at 0.2.0; and `deny_unknown_fields` means an older server answers 400 to a request carrying `strike_ladder`, so it is a version gate rather than a no-op.

## Gate

`make pre-push` clean, MongoDB up: **917** + 16 + 3 tests pass, 25 ignored; clippy with `-D warnings`, `fmt --check`, `cargo doc` with zero warnings and `cargo build --release` all clean; `README.md` regenerated via `make readme`. The one pre-existing doc warning, a broken link in `utils::env`, is fixed here too so the gate is warning-free.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ members = [
 
 [workspace.dependencies]
 optionchain_simulator = { path = "." }
-optionstratlib = "0.20"
+optionstratlib = "0.21"
 positive = "0.6"
 tracing = "0.1"
 rust_decimal = { version = "1.42", features = ["maths", "serde"] }
--- a/src/domain/factors.rs
+++ b/src/domain/factors.rs
@@ -69,7 +69,9 @@
 //! two agree algebraically but not in the last `Decimal` digits. The tolerance
 //! is pinned by a test. ADR 0001 §8 records the contract.
 
+use crate::api::rest::limits::MAX_CHAIN_SIZE;
 use crate::domain::Walker;
+use crate::domain::ladder::PinnedLadder;
 use crate::domain::simulator::{DEFAULT_CHAIN_SIZE, DEFAULT_SKEW_SLOPE, DEFAULT_SMILE_CURVE};
 use crate::domain::spread::SpreadModel;
 use crate::session::{SimulationMethod, SimulationParametersV2};
@@ -949,7 +951,21 @@ pub(crate) fn build_chain(
     expiration: ExpirationDate,
     greek_snapshots: bool,
 ) -> Result<OptionChain, ChainError> {
-    let chain_size = parameters.chain_size.unwrap_or(DEFAULT_CHAIN_SIZE);
+    // A pinned simulation quotes the ladder it fixed at creation, so the chain
+    // is built wide enough to REACH those strikes from wherever the spot is now
+    // and filtered down to them afterwards. Upstream anchors every ladder on
+    // the same interval grid, so the pinned strikes are always among the ones
+    // it builds. See `crate::domain::ladder`.
+    let pinned = if parameters.strike_ladder.is_pinned() {
+        Some(PinnedLadder::resolve(parameters)?)
+    } else {
+        None
+    };
+
+    let chain_size = match &pinned {
+        Some(ladder) => ladder.width_from(spot, *MAX_CHAIN_SIZE)?,
+        None => parameters.chain_size.unwrap_or(DEFAULT_CHAIN_SIZE),
+    };
     let skew_slope = parameters.skew_slope.unwrap_or(DEFAULT_SKEW_SLOPE);
     let smile_curve = parameters.smile_curve.unwrap_or(DEFAULT_SMILE_CURVE);
     // The spread model, not one scalar. Read once per chain: every coefficient
@@ -1014,6 +1030,13 @@ pub(crate) fn build_chain(
     let mut chain = OptionChain::build_chain(&build_params)
         .map_err(|e| ChainError::Internal(format!("Failed to build the option chain: {e}")))?;
 
+    // Down to the pinned strikes, before the spread model runs: the widened
+    // strikes exist only to reach the pinned ones and nothing should pay to
+    // quote them.
+    if let Some(ladder) = &pinned {
+        ladder.keep_pinned(&mut chain);
+    }
+
     spread_model.apply(&mut chain, days_to_expiration);
 
     Ok(chain)
@@ -1047,6 +1070,7 @@ mod tests {
     use super::*;
     use crate::api::rest::models::{ApiTimeFrame, ApiWalkType};
     use crate::api::rest::requests_v2::CreateSimulationRequest;
+    use crate::domain::ladder::StrikeLadder;
     use crate::session::{ExpiryRule, ExpiryRuleKind};
     use chrono::{TimeZone, Weekday};
     use optionstratlib::error::SimulationError;
@@ -1083,6 +1107,7 @@ mod tests {
             skew_slope: None,
             smile_curve: None,
             spread: Some(0.02),
+            strike_ladder: Default::default(),
             spread_proportional: None,
             spread_moneyness_widening: None,
             spread_tenor_widening: None,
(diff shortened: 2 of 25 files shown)


Cargo.toml
```diff
@@ -64,7 +64,7 @@ members = [
 
 [workspace.dependencies]
 optionchain_simulator = { path = "." }
-optionstratlib = "0.20"
+optionstratlib = "0.21"
```
joaquinbejar, line 67: 🟠 **Warning**: this upgrades optionstratlib while leaving the seeded Walker on the 0.20 kernel bodies. The upstream comparison recorded in #97 found concrete hardening gaps, including size zero underflows in Heston and Custom and unchecked arithmetic. This repository requires the mirror to be resynced in the same dependency bump, with the intentional jump diffusion divergence preserved.

src/domain/factors.rs
```diff
@@ -949,7 +950,21 @@ pub(crate) fn build_chain(
...
+    // the same interval grid, so the pinned strikes are always among the ones
+    // it builds. See `crate::domain::ladder`.
+    let pinned = if parameters.strike_ladder.is_pinned() {
+        Some(PinnedLadder::resolve(parameters)?)
+    } else {
+        None
+    };
+
+    let chain_size = match &pinned {
+        Some(ladder) => ladder.width_from(spot, MAX_PINNED_WIDTH)?,
```
joaquinbejar: 🟠 **Warning**: this replaces the validated chain width with a separate runtime ceiling. A request accepted using a small declared `chain_size` and many expirations can later price up to 1,001 strikes per expiration before filtering, above `OCS_MAX_SNAPSHOT_CONTRACTS`. Derive this width from the same per snapshot budget before calling upstream; deferring that to #96 leaves the new path able to bypass its resource cap.

src/domain/ladder.rs
```diff
@@ -0,0 +1,666 @@
...
+    ///
+    /// Returns [`ChainError::Validation`] naming `strike_ladder` when that
+    /// distance exceeds [`MAX_PINNED_WIDTH`]. Reaching it takes a spot that has
+    /// moved that many intervals from where the simulation started, and quoting
+    /// a subset of the pinned ladder would silently break the guarantee the
+    /// ladder exists to make. It is a validation failure rather than an
+    /// internal one because it is a foreseeable consequence of the
+    /// configuration the client chose, and the message says which.
+    pub(crate) fn width_from(&self, spot: Positive, maximum: usize) -> Result<usize, ChainError> {
+        let centre = at_the_money(spot, self.interval)?.to_dec();
```
joaquinbejar: 🟠 **Warning**: when a simulated spot falls below half the interval, this anchor is zero and `width_from` accepts it. The following upstream build rejects the zero ATM strike, which this service maps to an internal error; during export that happens after the 200 response and prologue, leaving a truncated file. Reject this case as a typed ladder validation error before pricing and cover it with a low spot test.

joaquinbejar (commented):
The pinned ladder contract is useful, but its runtime widening bypasses the validated work budget and low simulated prices can still turn a successful export into a truncated response. The dependency bump also leaves the seeded Walker behind upstream hardening, and this 1268 line change is large enough that separating the optionstratlib migration and deployment scaling from the ladder feature would make each safer to validate. REQUEST_CHANGES.

Cargo.toml
joaquinbejar: Fair, taking the values without the hardening was the wrong half of the upgrade. The mirror now carries the 0.21 kernels including the saturating loop bounds in Heston and Custom and checked arithmetic throughout, with the jump diffusion divergence untouched, and a golden dump of all nine walk types plus the four with_vol variants is identical before and after so no seeded tape moved. That closes #97 here rather than after the merge.

src/domain/factors.rs
joaquinbejar: You are right that a constant of my own was the wrong ceiling. max_pinned_width now divides the per snapshot contract cap by the expirations the schedule can carry and converts back through the 2n + 1 grid, so the widening is bounded by the same budget validate_snapshot_work accepted at creation and a drifted spot cannot buy itself more work. #96 is closed by this commit.

src/domain/ladder.rs
joaquinbejar: Good catch, the truncated file is the part that makes this worse than a plain failure. width_from now refuses a spot below half the interval as a ladder validation error before anything is priced, and there is a low spot test covering both the refusal and the first spot that still anchors on the grid.

joaquinbejar:
All three are addressed in affcba3 and the gate is clean. On splitting the PR, the ladder cannot ship without the 0.21.1 bump since its acceptance test needs the wing fix that release carries, so separating them would leave one half unable to prove itself; the bump (d365a01) and the replica scaling (768f5e8) are each already isolated in their own commit, which is the granularity that helps a bisect.

PR #108, opened 2026-08-30
Bound the pinned widening by an absolute ceiling as well as the snapshot budget

joaquinbejar:
Follow-up to the review of #98, from the `reproducibility-reviewer` audit that landed after the merge.

## The problem

The ceiling that PR #98 introduced answers the review finding it was written for, but it is not the bound its own comments claim. `max_pinned_width` divides `OCS_MAX_SNAPSHOT_CONTRACTS` by the expirations the schedule carries, and that variable is a **service-wide** cap, not a per-simulation one. At its documented default of 200 000, a single-expiration schedule gets 99 999 strikes per side, so a pinned step could ask upstream to build and price 199 999 strikes before `keep_pinned` discards everything outside the ladder. The `MAX_PINNED_WIDTH = 500` it replaced was 200 times stricter, and 500 is also the widest chain a client is allowed to request in the first place.

The widened chain is transient, so `OCS_MAX_CACHED_SNAPSHOT_CONTRACTS` never sees it either; only the filtered result reaches the cache.

## What changed

Both bounds now apply and the smaller wins. `MAX_PINNED_WIDTH` returns as an absolute ceiling on the widening the service performs on a client behalf, and `max_pinned_width` takes `budget.min(MAX_PINNED_WIDTH)`. The budget half-width also loses its floor of one, which could exceed the very budget it derived from when the cap divided down to two or fewer strikes; zero is now a legitimate answer and `width_from` refuses the ladder that cannot fit.

The test asserts both halves: the `2n + 1` grid fits the snapshot cap, and the width stays under the absolute ceiling that actually binds at the default configuration.

## Not in this PR

Two findings from the same audit are filed rather than fixed here, because each is a change of its own shape:

- #109, the step-time read of `max_snapshot_contracts()`, which makes how far a pinned tape gets depend on the deployment it runs against.
- #110, a committed golden fixture for the walk kernels, since every reproducibility test in the suite compares two runs of the same build and is structurally blind to a value change.

## Checklist

- [x] Reproducibility intact; this touches no walk path and no price
- [x] Stored-session shape unchanged; `MAX_PINNED_WIDTH` is a private constant
- [x] REST DTOs unchanged
- [x] `ChainError` discipline; the refusal stays a typed validation error
- [x] `make pre-push` clean, zero warnings, 920 tests
- [x] Patch version bumped to 0.2.1, compose image tag with it

### DIFF
--- a/src/domain/ladder.rs
+++ b/src/domain/ladder.rs
@@ -233,7 +233,11 @@ impl PinnedLadder {
                 reason: format!(
                     "the underlying has moved {widest} strikes from the ladder this simulation \
                      pinned at creation, past the {maximum} a chain may carry; a pinned ladder \
-                     does not follow a move that far, so widen it at creation with chain_size"
+                     does not follow a move that far, and widening chain_size cannot help \
+                     because a wider ladder puts its far strikes further still from this spot; \
+                     create the simulation with a smaller chain_size, around an initial_price \
+                     nearer where the underlying trades, or with the rolling ladder, which \
+                     re-centres on every step"
                 ),
             });
         }
@@ -388,19 +392,34 @@ pub(crate) fn max_pinned_width(parameters: &SimulationParametersV2) -> Result<us
             reason: "the requested expiration counts overflow".to_string(),
         })?;
 
-    let cap = crate::infrastructure::max_snapshot_contracts();
+    Ok(pinned_width_for(
+        crate::infrastructure::max_snapshot_contracts(),
+        expirations,
+    ))
+}
+
+/// The widening [`max_pinned_width`] allows for a given cap and expiration
+/// count, split out from the environment so both sides of the minimum can be
+/// exercised: the process-wide cap is resolved once in a `OnceLock`, so a test
+/// that could only call the public entry point would be stuck with whatever
+/// the deployment happens to configure.
+fn pinned_width_for(cap: usize, expirations: usize) -> usize {
     // A schedule with no rules cannot produce a chain, so the divisor is
     // irrelevant; one keeps the division defined.
-    let strikes = cap.checked_div(expirations.max(1)).unwrap_or(cap);
+    let divisor = if expirations == 0 { 1 } else { expirations };
+    let strikes = cap.checked_div(divisor).unwrap_or(cap);
 
     // `strikes` covers the whole `2n + 1` grid, so the half-width is what
-    // upstream calls `chain_size`. No floor: a budget too small to carry even
-    // a three-strike chain must report a width of zero and let `width_from`
-    // refuse the ladder, rather than round itself up past the budget it came
-    // from.
-    let budget = strikes.saturating_sub(1).checked_div(2).unwrap_or(0);
+    // upstream calls `chain_size`. The subtraction is checked rather than
+    // saturating, which `rules/global_rules.md` forbids: a budget of zero or
+    // one strike is a real answer of zero, stated deliberately, and
+    // `width_from` then refuses a ladder that cannot fit rather than the
+    // arithmetic quietly rounding up past the budget it came from.
+    let budget = strikes.saturating_sub(1)
+        .checked_div(2)
+        .unwrap_or(0);
 
-    Ok(budget.min(MAX_PINNED_WIDTH))
+    budget.min(MAX_PINNED_WIDTH)
 }
 
 /// Refuses a pinned ladder whose lowest strike upstream cannot build.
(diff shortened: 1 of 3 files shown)


src/domain/ladder.rs
```diff
@@ -335,18 +336,40 @@ pub(crate) fn at_the_money(
...
+/// service performs on a client's behalf. The two are different quantities and
+/// only the coincidence of the number is shared, so it lives here rather than
+/// being read from the request caps: lowering the request cap must not
+/// retroactively break simulations already running, and raising it must not
+/// license the domain to price a wider chain than this.
+///
+/// It is a ceiling, never the bound on its own; [`max_pinned_width`] takes it
+/// together with the simulation's snapshot budget and keeps whichever is
+/// smaller.
+pub(crate) const MAX_PINNED_WIDTH: usize = 500;
```
joaquinbejar, line 356: 🟠 **Warning**: restoring this ceiling makes the `width_from` refusal reachable again, but that error tells clients to widen at creation with `chain_size`. Increasing `chain_size` adds farther pinned strikes and increases the upstream width required, so it cannot satisfy this bound. Recommend reducing `chain_size`, switching to rolling, or recreating around a nearer initial price, and cover the remedy in the refusal test.

src/domain/ladder.rs
```diff
@@ -366,14 +389,18 @@ pub(crate) fn max_pinned_width(parameters: &SimulationParametersV2) -> Result<us
...
-    // `strikes` covers `2n + 1`, so the half-width is what upstream calls
-    // `chain_size`. At least one, or a pinned ladder of a single strike could
-    // never be reached.
-    Ok(strikes.saturating_sub(1).checked_div(2).unwrap_or(0).max(1))
+    // `strikes` covers the whole `2n + 1` grid, so the half-width is what
+    // upstream calls `chain_size`. No floor: a budget too small to carry even
+    // a three-strike chain must report a width of zero and let `width_from`
+    // refuse the ladder, rather than round itself up past the budget it came
+    // from.
+    let budget = strikes.saturating_sub(1).checked_div(2).unwrap_or(0);
```
joaquinbejar: 🟠 **Warning**: this refactored zero budget path now relies on `saturating_sub`, which the mandatory arithmetic rules explicitly forbid. Use `checked_sub(1).unwrap_or(0)` or an explicit match before the checked division so the underflow policy stays deliberate and visible.

src/domain/ladder.rs
```diff
@@ -366,14 +389,18 @@ pub(crate) fn max_pinned_width(parameters: &SimulationParametersV2) -> Result<us
...
+
+    Ok(budget.min(MAX_PINNED_WIDTH))
```
joaquinbejar: 🟠 **Warning**: the absolute ceiling added here is not covered by an effective test. With the default single expiration budget, the previous width of 99,999 still satisfies the existing `width * 2 < cap` assertion, so that test passes if this `min` is removed. Assert that this case returns `MAX_PINNED_WIDTH`, and add a smaller budget case to make both sides of the minimum load bearing.

joaquinbejar (commented):
The combined budget is otherwise sound and CI is green, but the new ceiling is not protected by an effective regression test, the restored refusal still tells clients to increase the value that makes the required width larger, and the refactored zero budget path relies on saturating arithmetic forbidden by the repository rules. REQUEST_CHANGES.

src/domain/ladder.rs
joaquinbejar: You are right, that remedy makes the problem worse since a wider ladder puts its far strikes further from the spot still. The message now says widening cannot help and names the three that work, creating with a smaller chain_size, around a nearer initial_price, or with the rolling ladder, and the refusal test asserts all of them.

src/domain/ladder.rs
joaquinbejar: Correct, the rules forbid it and I reached for it out of habit. It is checked_sub(1).unwrap_or(0) now, with a comment saying the zero is the deliberate answer for a budget that cannot carry three strikes.

src/domain/ladder.rs
joaquinbejar: Good catch, that assertion passed with or without the min so it was proving nothing. The arithmetic moved into pinned_width_for which takes the cap instead of reading the process wide OnceLock, and the test now pins the ceiling binding at the default budget, the budget binding when it is split or small, and zero for a budget under three strikes.

joaquinbejar:
All three are fixed in db9cba4 and the gate is clean. The one worth calling out is the refusal text, since a client following the old advice would have made the failure harder to escape rather than easier.

src/domain/ladder.rs
joaquinbejar: Correction on this one, my first fix did not survive. clippy manual_saturating_arithmetic rewrites checked_sub(1).unwrap_or(0) straight back into saturating_sub, and make pre-push runs clippy --fix, so the commit that claimed to remove it shipped it. It is a match on checked_sub now, which says the same thing and survives the autofix, pushed as e7c950d.

Issue #112, opened 2026-08-30
Update the compose service images and base them on alpine 3.24


joaquinbejar:
The compose files and the CI service containers still pin the previous generation of the dependency images.

## What to do

Bump each image to the version below, pinned by immutable digest with the tag kept as a readable prefix, and re-resolve every digest with `docker buildx imagetools inspect <image:tag>` rather than copying one from anywhere else. Use the alpine-based variant wherever the image publishes one, and alpine 3.24 as the base wherever this repository builds its own layer.

| Image | Version |
|---|---|
| redis | 8.10.1 |
| mongodb | 8.2.12 |
| clickhouse | 26.8.1.2041 |

MongoDB publishes no alpine variant, so it stays on its own base.

## Acceptance criteria

- [ ] `Docker/docker-compose.yml` and `Docker/docker-compose.dev.yml` carry the new tags and their matching digests
- [ ] `Docker/Dockerfile` builds on alpine 3.24 where it uses alpine
- [ ] The CI service containers in `.github/workflows/*.yml` use the same versions as the compose files
- [ ] The Integration job passes against the new images
- [ ] `make pre-push` clean

Issue #111, opened 2026-08-30
Complete the mirrored checked-arithmetic helpers with d_div


joaquinbejar:
From the `reproducibility-reviewer` audit of PR #98.

## Context

`src/domain/walker.rs` mirrors optionstratlib crate-private `d_add`, `d_sub`, `d_mul` and `d_exp` so the seeded kernels can stay panic-free. It does not mirror `d_div`. The single division in the Heston kernel uses `checked_div(Decimal::TWO)` with a `walk_error`, which is value-identical today because the divisor is a constant and the quotient scale cannot exceed 28.

The next kernel that divides by a non-constant is where this stops being equivalent, and it would drift silently.

## Acceptance criteria

- [ ] `d_div` joins the other helpers, matching upstream semantics including the zero-divisor and rounding behaviour
- [ ] The Heston average-variance division uses it, with upstream operation tag
- [ ] The seeded tape is unchanged, proven by the reproducibility tests
- [ ] The helper set is documented as the complete mirror of upstream

Created from PR #98 review conversation.

Issue #110, opened 2026-08-30
Commit a golden walk-kernel tape fixture so a kernel change cannot land silently


joaquinbejar:
From the `reproducibility-reviewer` audit of PR #98.

## Context

The suite has `test_same_seed_produces_identical_tape`, `test_jump_diffusion_same_seed_same_tape`, `test_a_pinned_simulation_reproduces_its_tape` and their siblings, and they all pass. Every one of them compares two runs of the SAME build, so they are structurally blind to a change in the values the kernels produce. The optionstratlib 0.21.1 re-sync was exactly that class of change; its byte-identity was verified out of band, by a throwaway dump, and nothing in CI pins it.

An upstream bump that shifts a kernel would therefore go green while quietly re-baselining every tape IronCondor has recorded.

## What to do

Commit the golden dump as a fixture test: the price path of all nine walk types plus the four `*_with_vol` variants, at a fixed seed and a fixed size, compared against a checked-in expected file.

## Acceptance criteria

- [ ] A fixture test covers all nine `WalkType` variants and the four `*_with_vol` variants at a fixed seed
- [ ] The expected values live in the repository, not in the test body, so a diff shows what moved
- [ ] The test fails loudly on any value change, with a message that names the walk type
- [ ] The fixture is hermetic and runs in the default `cargo test --workspace`
- [ ] A comment states that regenerating it is a deliberate, tape-breaking act and what it means for IronCondor

Created from PR #98 review conversation.

Issue #107, opened 2026-08-30
Integration tests: behaviour under concurrent clients and a long horizon


joaquinbejar:
## Overview

The service is meant to be sharded across instances and read by more than one client at a time, and the export is meant to stream rather than materialise. Those are the properties nothing in the suite touches, because they only exist under load.

Depends on #99, and reads best after #101 and #103.

## What to cover

- **Two clients, one simulation.** Both export the same simulation at once and receive identical bytes; neither disturbs the other's cursor
- **Two clients, two simulations.** Interleaved advances stay independent: the tape of each matches what it produces alone
- **Concurrent advance.** Several advances fired at once produce exactly one winner per step, the losers get `409`, and the cursor ends where the arithmetic says
- **A read is safe.** A snapshot peek during another client's advance never returns a half-built chain
- **The export streams.** A long export starts delivering before it could possibly have finished pricing, and the client can abandon it mid-stream without the service faulting
- **A deleted simulation stops serving.** An export in flight when the simulation is deleted either completes or fails cleanly, never with a panic

## Tasks

- [ ] Bound every concurrency test to a handful of clients and a short horizon; this runs against a shared deployment
- [ ] Assert the abandoned-export case by dropping the response early and then checking the service still serves a fresh request
- [ ] Record the observed wall time and payload size for the long export, so a regression in the streaming path shows up as a number

## Acceptance criteria

- [ ] Exactly one winner per contended step, and the cursor arithmetic holds
- [ ] Concurrent exports of one simulation are byte-identical
- [ ] The service serves normally after a client abandons a stream mid-flight
- [ ] Skipped cleanly when `OCS_INTEGRATION_BASE_URL` is unset
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check` clean

PR #113, opened 2026-08-30
Fix a pinned simulation's widening ceiling at creation

### LINKED ISSUE #109: Fix a pinned simulation's widening ceiling at creation instead of reading it per step
From the `reproducibility-reviewer` audit of PR #98, follow-up to PR #108.

## Context

`max_pinned_width` calls `crate::infrastructure::max_snapshot_contracts()` on every `build_chain`, so the ceiling on a pinned widening is read from a process-global `OnceLock` at step time rather than fixed when the simulation was created.

The tape VALUES are unaffected; how far the tape gets is not. The same parameters and the same seed can serve every step on an instance running the default cap and fail with a `ChainError::Validation` at step k on an instance where an operator lowered `OCS_MAX_SNAPSHOT_CONTRACTS`. The constant this replaced was deployment-independent, so the coupling is new, and it sits in exactly the guarantee IronCondor gates a milestone on.

## What to do

Resolve the ceiling once, at creation, and store it on `SimulationParametersV2`, the same pattern the effective seed already uses: generated when absent, carried in the stored parameters, and read from there by the domain. A session then keeps the ceiling it was created under for its whole life and replays identically on any instance.

## Acceptance criteria

- [ ] The pinned widening ceiling is resolved once at creation and stored on the parameters
- [ ] The field carries `#[serde(default)]`, and a session written by an older binary still loads with the ceiling it would have had
- [ ] `max_pinned_width` (or its successor) reads the stored value; no step-time read of `max_snapshot_contracts()` remains on the pinned path
- [ ] A test proves a stored session keeps serving its ladder after the environment cap changes underneath it
- [ ] Serde round-trip for the old stored shape

joaquinbejar:
Closes #109

Base branch: `main`
Blocks: #114

Stack: #109 → #110 → #111 → #112 → #99 → #104 → #100 → #101 → #102 → #103 → #106 → #105 → #107 (this is the first)

## What changed

`max_pinned_width` read `OCS_MAX_SNAPSHOT_CONTRACTS` through a process-global `OnceLock` on every `build_chain`, so the ceiling on a pinned widening came from the instance rather than from the session. Tape VALUES were unaffected; how far a tape got was not. The same parameters and the same seed could serve every step on one deployment and fail with a validation error at step k on another whose operator had lowered the cap, which is new coupling inside the guarantee IronCondor gates a milestone on.

It is resolved once now, at creation, by `resolve_pinned_ceiling` from this instance cap and this simulation own schedule, and stored on `SimulationParametersV2` as `pinned_width_ceiling`, next to the effective seed whose pattern it copies. `build_chain` reads the stored number; nothing on the pinned path consults the environment any more.

## Stored-session compatibility

The wire form takes the field as `Option<usize>` with `#[serde(default)]`. A document written before this lands carries no ceiling and resolves one on load, which is the ceiling it would have had, then keeps it. A document that carries one keeps exactly that, which is what makes a session created under a tighter cap stay tight on a wider instance.

## Tests

- The creation-time resolver agrees with the pure core, and the request conversion stores what it resolved
- A step obeys the STORED ceiling rather than what this instance would resolve now, refusing a drift the session was not created for
- A document without the field loads with the ceiling it would have had
- A stored ceiling survives the round trip rather than being re-resolved

## Checklist

- [x] Reproducibility intact; no walk path touched, and the ladder is still a pure function of stored parameters
- [x] Stored-session shape additive, `#[serde(default)]`, old-document test
- [x] REST DTOs unchanged
- [x] `ChainError` discipline
- [x] `make pre-push` clean, zero warnings, 924 tests
- [x] Patch version bumped to 0.2.2 with the compose image tag


src/session/model_v2.rs
```diff
@@ -299,6 +299,19 @@ pub struct SimulationParametersV2 {
...
+    /// Resolved ONCE, at creation, from the per-snapshot contract cap and this
+    /// simulation's own schedule, exactly as the effective seed is resolved
+    /// once and then carried. The domain reads it from here rather than from
+    /// the environment, so a session keeps the ceiling it was created under for
+    /// its whole life: the same parameters and seed then reach the same step on
+    /// any instance, whatever an operator has since done to
+    /// `OCS_MAX_SNAPSHOT_CONTRACTS`.
+    ///
+    /// A `rolling` simulation carries it too and never consults it.
+    pub pinned_width_ceiling: usize,
```
joaquinbejar, line 314: 🟠 **Warning**: this value is now an effective replay parameter because it controls the first step at which a pinned run can fail, but `SimulationParametersResponse` omits it and `CreateSimulationRequest` cannot accept it. Recreating the documented replay input list on an instance with a different cap therefore resolves a different tape boundary. Echo the resolved ceiling and provide a validated replay input, or derive the ceiling only from already public replay inputs.

joaquinbejar (commented):
This change moves the widening ceiling into stored parameters, but deserialization still depends on the current deployment cap, the new replay input is absent from the public replay surface, and the persisted ceiling is trusted without bounds. The formal decision is REQUEST_CHANGES.

src/session/model_v2.rs
```diff
@@ -347,12 +360,23 @@ struct SimulationParametersV2Wire {
...
 }
 
 impl TryFrom<SimulationParametersV2Wire> for SimulationParametersV2 {
     type Error = ChainError;
 
     fn try_from(wire: SimulationParametersV2Wire) -> Result<Self, Self::Error> {
+        // A document written before the ceiling was stored resolves it once,
+        // here, and carries it from then on.
+        let pinned_width_ceiling = match wire.pinned_width_ceiling {
+            Some(ceiling) => ceiling,
```
joaquinbejar, line 377: 🟠 **Warning**: a present value from Redis is trusted verbatim, and `validate()` never bounds it. A corrupted document can set this above `MAX_PINNED_WIDTH` and let a pinned step bypass the absolute CPU and memory guard. Validate the persisted value against creation time budget metadata, at least enforcing the absolute maximum, and add a corrupted document test.

src/session/model_v2.rs
```diff
@@ -377,6 +401,7 @@ impl TryFrom<SimulationParametersV2Wire> for SimulationParametersV2 {
             spread_tenor_widening: wire.spread_tenor_widening,
             spread_tick: wire.spread_tick,
             seed: wire.seed,
+            pinned_width_ceiling,
```
joaquinbejar, line 404: 🟠 **Warning**: the stored ceiling is immediately followed by `parameters.validate()`, whose `validate_snapshot_work()` reads this process's current `max_snapshot_contracts()`. A session valid under a higher creation cap can therefore fail deserialization on an instance with a lower cap before the stored ceiling is used, so #109's cross instance guarantee is still broken. Separate invariant validation from the creation time deployment cap check or persist the effective budget needed to validate deterministically, then test loading and serving under a lower cap.

src/session/model_v2.rs
joaquinbejar: Agreed, it decides the first step a pinned run refuses at, so it belongs in the echoed replay inputs. SimulationParametersResponse carries it now, and it stays out of the request deliberately since letting a client raise the bound on what the service builds for it would turn a resource guard into a suggestion.

src/session/model_v2.rs
joaquinbejar: Right, Redis is an outer layer and this is a number the domain then obeys. The ceiling is checked against MAX_PINNED_WIDTH on every path now, with a test that a hand-edited document one above it is refused and names the field.

src/session/model_v2.rs
joaquinbejar: Good catch, that was the same breakage in a different place. The cap is admission control about what to accept next, and a session that already exists answered it at creation, so validate_stored now warns where validate rejects, exactly as a tzdb release from another version does, while every deployment-independent invariant still applies. The test creates a configuration well past any cap and asserts creation refuses it while loading keeps working.

joaquinbejar:
All three are addressed in 8402c2d and the gate is clean. The third one was the important one: loading a stored session was still re-answering a deployment question at the wrong time, which is the same class of breakage this PR set out to remove.

src/api/rest/responses_v2.rs
```diff
@@ -136,6 +136,19 @@ pub struct SimulationParametersResponse {
...
+    /// Echoed because it is a replay input like the seed: it decides the first
+    /// step at which a pinned simulation refuses a drift, so a replay that
+    /// resolved a different one is a different tape boundary. A client
+    /// recreating a run on another instance can compare this value and know
+    /// whether the boundary moved, rather than discovering it at step k.
+    ///
+    /// It is not accepted on a request. The number bounds what the service
+    /// will build on a client's behalf, so letting a client raise it would
+    /// turn a resource guard into a suggestion.
+    pub pinned_width_ceiling: usize,
```
joaquinbejar, line 151: 🔴 **Critical**: This value is called a replay input, but the create request deliberately cannot accept it and rejects unknown fields. Recreating the returned parameters on another instance therefore re-resolves a different ceiling and may stop at a different step; comparison detects divergence but cannot reproduce the run. Accept and validate an optional replay ceiling, derive the boundary only from public replay inputs, or explicitly narrow the replay guarantee and coordinate IronCondor.

src/session/model_v2.rs
```diff
@@ -517,7 +573,20 @@ impl SimulationParametersV2 {
             });
         }
         self.schedule.validate()?;
-        self.validate_snapshot_work()?;
+        self.enforce_snapshot_work(enforce_cap)?;
+
+        // Deployment-independent, so it holds on every path: a stored ceiling
+        // above the absolute maximum would let a corrupted document buy a
+        // widening no request could ever ask for.
+        if self.pinned_width_ceiling > MAX_PINNED_WIDTH {
```
joaquinbejar, line 581: 🟠 **Warning**: Stored validation checks only the upper bound. A pinned document with `pinned_width_ceiling` below `chain_size.unwrap_or(DEFAULT_CHAIN_SIZE)` deserializes successfully, but the initial ladder already needs that base width and fails on step zero. Reject this impossible lower bound for pinned ladders and add a stored JSON regression case, including the absent-field migration path.

src/api/rest/responses_v2.rs
```diff
@@ -136,6 +136,19 @@ pub struct SimulationParametersResponse {
     pub smile_curve: Option<f64>,
     /// Which strikes the simulation quotes: `rolling` or `pinned`.
     pub strike_ladder: StrikeLadder,
+    /// How far a pinned ladder may make a step widen the chain, in strikes per
```
joaquinbejar, line 139: 🔴 **Critical**: Adding a required field to this published public Rust struct is source breaking for downstream construction and destructuring, and may also break strict response decoders. A `0.2.x` patch remains on the same compatibility line; after restoring the stack, move this contract change to `0.3.0`, synchronize the deployment tag, and coordinate the IronCondor consumer.

joaquinbejar (commented):
Formal decision REQUEST_CHANGES. The branch has structurally collapsed into 16 commits and 5696 changed lines, so merging this first stack layer would also merge every later PR from #114 through #127; restore it to the #109 change plus its review fix and rebase the successors. Even after that, the replay ceiling cannot be supplied back to creation, its stored lower bound is not validated, and the required field is a source breaking public struct change without a compatible version bump. CI is green but does not resolve these blockers.

Issue #106, opened 2026-08-30
Integration tests: the market a snapshot describes must be coherent


joaquinbejar:
## Overview

Every other issue in this series checks that the service answers correctly. This one checks that what it answers is a market: a chain whose quotes are ordered, whose greeks belong to their contracts, and whose strikes behave the way the configuration said they would. A service can be perfectly well-behaved over HTTP and still serve nonsense.

Depends on #99.

## What to cover

Over a walked simulation, at several steps:

- **Quotes are ordered.** For every quoted contract, `bid <= mid <= ask`, and no side is negative
- **A quote is not withdrawn.** A contract that has a mid has both a bid and an ask
- **The chain is centred and complete.** `chain_size` strikes above and below the money, at `strike_interval` apart, with no gaps and no duplicates
- **Expirations roll but do not vanish mid-life.** An expiration present at one step is present at the next until it expires, and its `days_to_expiration` decreases
- **The greeks belong to their level.** `first` adds exactly the documented eight columns, `all` the documented fourteen more, and delta agrees between the shared column and the snapshot
- **The underlying moves.** Over a walk long enough to matter, the spot is not constant, and `simulated_at` advances by the configured step interval every time
- Where the deployment has them: the spread model widens a cheap contract more in relative terms than a dear one, and a pinned `strike_ladder` quotes the same strike set at every step

## Tasks

- [ ] Assert per contract, not per chain, and report the offending strike when one fails
- [ ] Drive the greek-level assertions from the export as well as the snapshot, since they render through different paths
- [ ] Skip the spread-model and ladder assertions with a clear message when the deployment predates them
- [ ] Keep the walk short and the chain narrow, and say what was chosen

## Acceptance criteria

- [ ] A crossed book, a withdrawn quote, a gap in the ladder or a stalled clock fails the suite
- [ ] The assertions name the strike, the step and the expiration that failed
- [ ] Skipped cleanly when `OCS_INTEGRATION_BASE_URL` is unset
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check` clean

Issue #105, opened 2026-08-30
Integration tests: the operational surface — metrics, OpenAPI, and the probes


joaquinbejar:
## Overview

The parts an operator points a probe at, a scraper at, or a code generator at. They are not the product, but they are what makes the product deployable, and each of them has already been wrong once.

Depends on #99.

## What to cover

- `GET /metrics` answers `200 text/plain` and exposes the counters the service documents, and a served request moves the request counter
- `GET /api-docs/openapi.json` parses, and **every path it advertises actually answers** rather than 404 — a generated client is built from this document
- `GET /swagger-ui/` serves the UI
- `GET /favicon.ico` answers
- Where the deployment has them: `GET /health` is `200` regardless of dependency state, and `GET /ready` is `200` when the dependencies answer, with a body naming each one; both are ABSENT from the metrics, so probing does not pollute the counters
- An unknown path is `404`, not `500`

## Tasks

- [ ] Drive the endpoint list from the deployment's own OpenAPI document rather than from a hardcoded list, so the test tracks the contract the service publishes
- [ ] Skip the probe assertions with a clear message when the deployment predates them, rather than failing
- [ ] Assert the metrics exclusion by reading `/metrics` before and after a burst of probe requests

## Acceptance criteria

- [ ] Every advertised path is reachable
- [ ] The metrics exposition parses and moves when the service serves traffic
- [ ] The probe assertions run against a deployment that has them and skip loudly against one that does not
- [ ] Skipped cleanly when `OCS_INTEGRATION_BASE_URL` is unset
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check` clean

Issue #104, opened 2026-08-30
Integration tests: the rejection contract, field by field


joaquinbejar:
## Overview

A client meets this service's error shape more often than its happy path, and the shape is a promise: `{error, field}`, with `field` naming the offender where it is known. Nothing checks that promise against a deployment today.

Depends on #99.

## What to cover

- A malformed simulation id is `400` naming `id`, not `404` and not `500`
- A range that is reversed or past the tape names `from_step` or `to_step`
- An unknown greek level names `greeks`
- Values outside their bounds name the field they belong to: `steps` below one or above the cap, `chain_size` above the cap, a non-positive `initial_price` or `volatility`, a zero `step_interval_seconds`
- An unknown key in the query string or the JSON body is refused rather than ignored, because a misspelled parameter must not read as a default
- A body that fails to deserialise at all still answers the documented shape, with `field` empty and the detail in `error`
- A missing simulation is `404` on every verb that takes an id
- Nothing in any rejection body leaks a credential, a host name or an internal path

## Tasks

- [ ] A table-driven test: request, expected status, expected `field`
- [ ] Cover both v1 and v2, since they share the shape but not the handlers
- [ ] Assert the `Content-Type` of a rejection is JSON, since a plaintext error is what an unhandled path produces
- [ ] One test that no 4xx or 5xx body contains anything URL-like

## Acceptance criteria

- [ ] Every rejection asserted by status AND by field, never by status alone
- [ ] A 500 anywhere in the table is a failure: these are all client errors
- [ ] Skipped cleanly when `OCS_INTEGRATION_BASE_URL` is unset
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check` clean

PR #114, opened 2026-08-30
Commit a golden walk-kernel tape so a kernel change cannot land silently

### LINKED ISSUE #110: Commit a golden walk-kernel tape fixture so a kernel change cannot land silently

joaquinbejar:
Closes #110

Base branch: `stack/1-109-pinned-ceiling-at-creation`
Stacked on: #113
Blocks: #115

Stack: #109 → **#110** → #111 → #112 → #99 → #104 → #100 → #101 → #102 → #103 → #106 → #105 → #107

The diff is cumulative until #113 merges, so read it against the base branch rather than against `main`.

## Why

`test_same_seed_produces_identical_tape` and its siblings all compare two runs of the SAME build. They prove the walker is deterministic; they cannot see a kernel that starts returning different numbers. The optionstratlib 0.21 re-sync was precisely that class of change, its byte-identity was checked by a throwaway dump, and CI pinned nothing. An upstream bump that shifted a price would have gone green while quietly re-baselining every tape a consumer had recorded.

## What changed

`src/domain/golden_tape.rs` walks all nine `WalkType` variants plus the four `*_with_vol` variants at seed 7 over 30 steps, against a real chain so the kernels run as they do in a simulation, and compares the result with `src/domain/fixtures/walk_kernels.json`. Values are rendered as exact decimal strings, never floats, since a float would round away the digits a kernel change moves first.

A failure names the kernel and the first differing step. A path that is walked but not committed fails too, so adding a walk type without extending the fixture is caught.

Regeneration is an `#[ignore]`d test rather than a flag, and the module documentation states plainly that regenerating breaks every recorded tape, needs announcing to IronCondor, and comes with a `CURRENT_SNAPSHOT_GENERATION` bump.

I verified the fixture is load-bearing by editing one committed value and watching the test name the kernel and the step.

## Tests

925 pass, 26 ignored (the new regeneration test among them). The fixture test is hermetic and runs in the default `cargo test --workspace`.

## Checklist

- [x] Reproducibility intact; no production code changed, this only observes it
- [x] Stored-session shape untouched
- [x] REST DTOs unchanged
- [x] Patch version bumped to 0.2.3 with the compose image tag


joaquinbejar (commented):
The incremental golden fixture covers all nine walk types and all four volatility producing variants at a fixed seed, compares exact decimal strings, names the failing kernel, and documents regeneration as tape breaking. The cumulative GitHub diff exceeds 500 lines because it includes the lower stack and the checked in fixture, while this layer is already isolated as its own commit. The formal decision for this incremental layer is APPROVE, contingent on fixing and merging #113 first.

joaquinbejar:
Thanks, and the contingency is met: #113 was reverted and its work landed as #129, which is on main now. This branch is rebased onto that, so the diff here is just the fixture and its module.

Issue #103, opened 2026-08-30
Integration tests: the export matrix, every dataset in every format the deployment offers


joaquinbejar:
## Overview

The export endpoint is the one a backtester actually consumes, and it has the widest surface in the service: three datasets, up to four encodings, a step range, and three greek levels. It also streams, so its failure modes are the kind that only appear over a socket.

Depends on #99.

## What to cover

- Every dataset — `underlying`, `volatility`, `option_chains` — in every format the deployment advertises, with the row counts agreeing across formats
- `json` is a single valid array; `csv` is RFC 4180 with a header and CRLF endings
- The **same export twice is byte-identical**, which is the endpoint's stated contract
- `from_step` and `to_step` select the same rows in every format, and an inverted or out-of-range pair is a typed 400 in every format
- The greek levels: each level's header is a PREFIX of the next, and `none` matches what the export produced before the parameter existed
- Where the deployment offers them, the binary encodings: an `arrow` stream decodes to the same values as `csv`, and a `packed` document has 8-byte aligned payloads, a footer, and decodes to the same values
- The suggested filename and content type match the format asked for

## Tasks

- [ ] Read the deployment's OpenAPI document first and exercise only the formats it advertises, so the suite is honest against an older build rather than red
- [ ] Compare formats as parsed values, never as text
- [ ] A test that asserts the streaming path does not truncate: the last row of a long export is present and complete
- [ ] Keep the exports small — a few steps, a narrow chain — and say why in a comment

## Acceptance criteria

- [ ] Row counts and values agree across every advertised format for all three datasets
- [ ] Repeat exports are byte-identical
- [ ] Range and level behaviour is asserted per format, including the rejections
- [ ] The suite passes against a deployment that offers only `json` and `csv`, and covers more when more is deployed
- [ ] Skipped cleanly when `OCS_INTEGRATION_BASE_URL` is unset
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check` clean

Issue #102, opened 2026-08-30
Integration tests: the seed reproducibility contract, against a deployed service


joaquinbejar:
## Overview

Same parameters plus same seed yields an identical snapshot tape. It is the contract IronCondor gates a milestone on, and every test that proves it today runs in-process, where the walk is one function call away. This proves it where a consumer actually meets it: two simulations created over HTTP against a running deployment.

Depends on #99.

## What to cover

- Two v2 simulations created with **identical parameters and the same explicit seed**, walked step by step, produce byte-identical snapshots at every step: the same spots, the same strikes, the same quotes
- The same, for v1 sessions
- A **different** seed produces a different tape, so the first assertion cannot pass vacuously
- An **omitted** seed still comes back in the parameters, and re-creating a simulation with that echoed seed reproduces the original tape. This is the whole point of surfacing it
- The tape survives the service: created, walked halfway, then walked to the end in a second pass, with the second half unchanged

## Tasks

- [ ] A pair-walk helper that advances two simulations in lockstep and compares the rendered snapshots
- [ ] Compare as parsed values rather than as strings, and say in the test why: the JSON rendering of a float is not the contract, the value is
- [ ] Keep the horizon short enough to stay polite against a shared deployment, and say what it was
- [ ] Assert the negative control with a different seed in the same test file

## Acceptance criteria

- [ ] The reproduction is asserted step by step, not only on the final snapshot
- [ ] The echoed effective seed is shown to be sufficient to replay a run nobody chose a seed for
- [ ] A change in the walk path on the deployed build would fail these tests
- [ ] Skipped cleanly when `OCS_INTEGRATION_BASE_URL` is unset
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check` clean

Issue #101, opened 2026-08-30
Integration tests: the v2 simulation lifecycle and its cursor semantics


joaquinbejar:
## Overview

The v2 surface carries the concurrency vocabulary that makes a retry safe, and it is exactly the part a unit test proves least: `expected_step` is a precondition, the revision is a compare-and-swap token, and the two mean different things. Over the wire is where a client meets them.

Depends on #99.

## What to cover

- `POST /api/v2/simulations` returns `201` with every replay input echoed: the effective seed, the effective start, the step interval, the normalised schedules, and the chain parameters
- `GET /api/v2/simulations/{id}` returns metadata without building a snapshot or moving the cursor
- `GET /api/v2/simulations/{id}/snapshot` peeks repeatably
- `POST /api/v2/simulations/{id}/step` serves then advances, and `410` past the end
- `expected_step` matching the cursor advances; **mismatching returns `412` with the actual cursor and consumes nothing**, so the same call retried at the right step still works
- Two concurrent advances of the same simulation: one wins, the other gets `409`, and the cursor moves exactly once
- `DELETE` removes it and the id stops resolving

## Tasks

- [ ] A test per endpoint, plus one for the serve-then-advance walk to exhaustion
- [ ] A `412` test that asserts the body carries the actual cursor and that a follow-up advance at that cursor succeeds
- [ ] A `409` test that fires two advances at once and asserts exactly one won and the cursor moved by one
- [ ] Assert the normalised schedules come back sorted by rule id, since normalisation is part of the replay contract

## Acceptance criteria

- [ ] `412` and `409` are distinguished, not conflated
- [ ] The cursor is asserted after every failed precondition: nothing was consumed
- [ ] Each test cleans up its simulation
- [ ] Skipped cleanly when `OCS_INTEGRATION_BASE_URL` is unset
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check` clean

Issue #100, opened 2026-08-30
Integration tests: the v1 session lifecycle over the wire


joaquinbejar:
## Overview

`/api/v1/chain` is frozen on rendered values, not only on shapes (ADR 0001 §12.1), and IronCondor consumes it. A frozen contract deserves a test that talks to a real deployment, because what is frozen is what a client receives, not what a handler returns.

Depends on #99.

## What to cover

The whole state machine, in one flow per test, each cleaning up after itself:

- `POST /api/v1/chain` creates a session and returns its parameters with the **effective seed**, generated when the request omitted one
- `GET /api/v1/chain` peeks the snapshot the next advance would serve, twice, and returns the same market both times without moving the cursor
- `POST /api/v1/chain/step` serves the snapshot at the cursor and then advances, so a session with `steps = N` serves indices `0..N-1` over `N` calls
- The advance that serves the last snapshot completes the session, and any further advance is `410 Gone`
- `PATCH` updates parameters, `PUT` replaces them, and both are visible in the next snapshot
- `DELETE` removes the session, after which every verb answers `404`

## Tasks

- [ ] One test per verb plus one for the full serve-then-advance walk
- [ ] Assert the effective seed is present and stable across the session's life
- [ ] Assert the `410` boundary exactly: the Nth advance succeeds, the N+1th does not
- [ ] Assert a missing session is `404` on every verb, not `500`
- [ ] Keep the session small — a handful of steps and a narrow chain — so the suite stays quick against a shared deployment

## Acceptance criteria

- [ ] The tests pass against a deployed 0.2.0 and would fail if a route, a status code or the effective-seed echo changed
- [ ] No test depends on another's state; each creates and deletes its own session
- [ ] Skipped cleanly when `OCS_INTEGRATION_BASE_URL` is unset
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check` clean

PR #115, opened 2026-08-30
Complete the mirrored checked-arithmetic helpers with d_div

### LINKED ISSUE #111: Complete the mirrored checked-arithmetic helpers with d_div

joaquinbejar:
Closes #111

Base branch: `stack/2-110-golden-tape-fixture`
Stacked on: #114
Blocks: #116

Stack: #109 → #110 → **#111** → #112 → #99 → #104 → #100 → #101 → #102 → #103 → #106 → #105 → #107

The diff is cumulative until #113 and #114 merge, so read it against the base branch rather than against `main`.

## What changed

The seeded walker mirrors optionstratlib crate-private checked helpers so its kernels stay panic-free, and it carried four of the five. The one division in Heston used `checked_div(Decimal::TWO)` with a `walk_error`, which is value-identical today only because the divisor is a constant: upstream re-rounds every quotient with `MidpointNearestEven` at scale 28 so a long chain of divisions does not accumulate bias in one direction, and the next kernel that divides by a non-constant would have drifted silently.

`d_div` now mirrors upstream exactly, zero-divisor arithmetic error included, and Heston average variance goes through it with upstream own operation tag. The helper set is documented as the complete mirror.

## Proof nothing moved

The golden tape from #110 passes unchanged, which is a stronger statement than the same-seed tests could make: it compares against values committed before this change, not against another run of this build.

## Tests

926 pass. A new unit test pins the three things that make `d_div` a mirror rather than a rename: a zero divisor is an arithmetic error naming the operation, a quotient is re-rounded at scale 28, and an ordinary division is exact.

## Checklist

- [x] Reproducibility intact, proven by the committed tape rather than by a same-build comparison
- [x] Stored-session shape untouched
- [x] REST DTOs unchanged
- [x] `ChainError` discipline; the helper returns `DecimalError`, which `SimulationError` already absorbs
- [x] Patch version bumped to 0.2.4 with the compose image tag

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "optionchain_simulator"
-version = "0.2.3"
+version = "0.2.4"
 edition = "2024"
 # The maximum `rust-version` across the RESOLVED graph, not the highest direct
 # dependency: clickhouse 0.15.1, nalgebra 0.35, statrs 0.19.1 and wide 1.7 all
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -111,7 +111,7 @@ services:
     # Published by .github/workflows/docker-publish.yml on every Cargo.toml
     # version bump, or by `make docker-push` from a workstation. Bump the
     # default tag in the same PR that bumps the crate version.
-    image: ghcr.io/joaquinbejar/optionchain-simulator:${OPTIONCHAIN_VERSION:-0.2.3}
+    image: ghcr.io/joaquinbejar/optionchain-simulator:${OPTIONCHAIN_VERSION:-0.2.4}
     deploy:
       replicas: 2
       restart_policy:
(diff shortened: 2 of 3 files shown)


joaquinbejar (commented):
The incremental arithmetic change matches optionstratlib division semantics exactly, preserves the upstream Heston operation tag, changes no random draw order, and leaves the committed golden tape unchanged. The formal decision for this incremental layer is APPROVE, contingent on fixing and merging the lower stack first.

joaquinbejar:
The contingency is met: #113 was reverted and its work landed as #129, and the golden tape from #114 is on main too. This branch is rebased onto that, so the diff is just the helper and the Heston division, and the committed tape still passes unchanged.

Issue #99, opened 2026-08-30
Integration test harness: one crate, one environment variable, skipped when absent


joaquinbejar:
## Overview

Every test in this repo runs in-process. That covers the code but not the thing an operator actually has: a service listening on a port, behind a container, with Redis and MongoDB and possibly ClickHouse behind it. Nothing today would catch a route that stopped being mounted, a middleware that changed a status code, or a deployment serving a version whose contract differs from the one the tests describe.

This issue builds the ground the rest of the series stands on: one place to put live tests, one way to point them at a service, and a default that keeps the hermetic suite hermetic.

## Shape

A new workspace member, `examples/integration`, beside `examples/session` and `examples/clickhouse`. The tests live there rather than in the root crate so the library keeps no test-only HTTP dependency, and so `cargo test --workspace` stays meaningful with no service running.

The service under test comes from **`OCS_INTEGRATION_BASE_URL`**, a full base URL including the scheme and port. Where that points is deployment configuration and belongs in an operator's environment, never in this repository. When the variable is unset every test SKIPS rather than failing, so the default suite is unaffected; when it is set, they run against whatever is deployed there.

## A dependency this needs

There is no HTTP client in the workspace: no `reqwest`, no `awc`, no `ureq`. `rules/global_rules.md` forbids adding one without explicit approval, so that decision belongs on this issue before the work starts. `reqwest` is the obvious candidate — `tokio` is already a workspace dependency and optionstratlib's `async` feature already pulls it — pinned as `X.X` in `[workspace.dependencies]` like everything else.

## Tasks

- [ ] Add `examples/integration` to the workspace members
- [ ] Read `OCS_INTEGRATION_BASE_URL` through `utils::env::read_var`'s rule: a blank value is an unset one
- [ ] Skip, loudly and once, when it is unset; never fail for that reason
- [ ] A thin client that owns the base URL, a request timeout, and JSON decoding, so no test writes a URL by hand
- [ ] A helper that creates a v2 simulation and deletes it afterwards, so a failing test does not leave state on a shared deployment
- [ ] `make test-integration`, and a CI job that is manual or scheduled rather than on every push
- [ ] Document the variable in `.env.example` and the crate docs, including that these tests exercise a DEPLOYED build and can therefore disagree with `main`

## Acceptance criteria

- [ ] `LOGLEVEL=WARN cargo test --workspace` with the variable unset stays green and starts no network connection
[...]

PR #116, opened 2026-08-30
Take the current service images, on alpine where they publish one

### LINKED ISSUE #112: Update the compose service images and base them on alpine 3.24

joaquinbejar:
Closes #112

Base branch: `stack/3-111-mirror-d-div`
Stacked on: #115
Blocks: #118

Stack: #109 → #110 → #111 → **#112** → #99 → #104 → #100 → #101 → #102 → #103 → #106 → #105 → #107

## What changed

| Image | Was | Now |
|---|---|---|
| redis | 8.10.0 | 8.10.1-alpine |
| mongo | 8.0.29 | 8.2.12 |
| clickhouse-server | 26.3.17.110 | 26.8.1.2041-alpine |

Every digest was re-resolved with `docker buildx imagetools inspect` rather than copied from anywhere, and tag and digest move together. Redis and ClickHouse take their alpine variants; MongoDB publishes none, so it stays on its own base. `Docker/Dockerfile` already builds on `rust:1.97.1-alpine3.24` over `alpine:3.24`, so nothing there needed changing.

The CI service containers move with the compose file, so the live-service job runs against what an operator would deploy.

## Verified rather than assumed

Each image was run locally with the credentials the compose file and the CI job actually supply. Worth knowing: **ClickHouse 26.8 refuses a query with no credentials** where the previous generation answered, so an environment that relied on the anonymous default user would break. Ours does not, both places set `CLICKHOUSE_USER` and `CLICKHOUSE_PASSWORD`, and with them the server answers `26.8.1.2041`. Redis replies `PONG`, MongoDB reports `8.2.12`.

## On the dev compose file

`Docker/docker-compose.dev.yml` carries no pins for these three services: it is an override that adds ports, container names and the UIs, and inherits the images from the base file, so there is exactly one place where each service image is pinned. Its own two images, mongo-express and redis-commander, are outside this issue.

## Checklist

- [x] Hermetic suite untouched; `make pre-push` clean, zero warnings
- [x] The Integration job on this PR is the proof the new service containers work
- [x] Images pinned by immutable digest, tags kept as readable prefixes
- [x] Patch version bumped to 0.2.5 with the compose image tag

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,23 +53,27 @@ jobs:
     name: Integration (live services)
     runs-on: ubuntu-latest
     services:
+      # Pinned by immutable digest, the same way the compose files are: a
+      # mutable tag can be republished, and a job testing different bytes from
+      # the ones a deployment runs proves nothing about that deployment.
+      #
       # Credentials must match the code's env-driven config defaults
       # (MONGODB_URI / REDIS_* / CLICKHOUSE_* in src/infrastructure/config/*),
       # so the ignored live tests authenticate successfully and can ASSERT
       # success instead of tolerating connection failures.
       mongodb:
-        image: mongo:8.2.12
+        image: mongo:8.2.12@sha256:e0ce8c35124d4a9f9785532d1f268f39e9728ffa1cb38f46fa482436424c4bd3
         env:
           MONGO_INITDB_ROOT_USERNAME: admin
           MONGO_INITDB_ROOT_PASSWORD: password
         ports:
           - 27017:27017
       redis:
-        image: redis:8.10.1-alpine
+        image: redis:8.10.1-alpine@sha256:becdda6c7f4b3fb42e42fd7f120bbf5c54c4caaaf16f26da24e4563d2c1f0576
         ports:
           - 6379:6379
       clickhouse:
-        image: clickhouse/clickhouse-server:26.8.1.2041-alpine
+        image: clickhouse/clickhouse-server:26.8.1.2041-alpine@sha256:fd037d424da807c1b36517f3ff6c739286c60a34c928279c229fc1faa0cadaa8
         env:
           CLICKHOUSE_USER: admin
           CLICKHOUSE_PASSWORD: password
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,18 +58,18 @@ jobs:
       # so the ignored live tests authenticate successfully and can ASSERT
       # success instead of tolerating connection failures.
       mongodb:
-        image: mongo:8.0.29
+        image: mongo:8.2.12
         env:
           MONGO_INITDB_ROOT_USERNAME: admin
           MONGO_INITDB_ROOT_PASSWORD: password
         ports:
           - 27017:27017
       redis:
-        image: redis:8.10.0
+        image: redis:8.10.1-alpine
         ports:
           - 6379:6379
       clickhouse:
-        image: clickhouse/clickhouse-server:26.3.17.110
+        image: clickhouse/clickhouse-server:26.8.1.2041-alpine
         env:
           CLICKHOUSE_USER: admin
           CLICKHOUSE_PASSWORD: password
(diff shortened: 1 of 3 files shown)


joaquinbejar (commented):
The requested tags exist, their compose digests resolve correctly, the development override inherits them, and every check is green, but CI tests mutable tags rather than the reviewed deployment artifacts and the persistent MongoDB volume has no verified or documented 8.0 to 8.2 FCV upgrade path. The formal decision is REQUEST_CHANGES.

.github/workflows/ci.yml
```diff
@@ -58,18 +58,18 @@ jobs:
       # so the ignored live tests authenticate successfully and can ASSERT
       # success instead of tolerating connection failures.
       mongodb:
-        image: mongo:8.0.29
+        image: mongo:8.2.12
```
joaquinbejar: 🟠 **Warning**: the integration services still use mutable tags while compose deploys the reviewed immutable digests. A republished tag makes this job test different bytes from production, so green CI no longer proves this rollout. Append the same resolved digest to each of the MongoDB, Redis, and ClickHouse image references.

Docker/docker-compose.yml
```diff
@@ -70,7 +70,7 @@ services:
 
   # MongoDB service configuration
   mongodb:
-    image: mongo:8.0.29@sha256:de267922bc1153d923f5c9dc429f21c11faf18299080c1ce04d6d6007097fb06
+    image: mongo:8.2.12@sha256:e0ce8c35124d4a9f9785532d1f268f39e9728ffa1cb38f46fa482436424c4bd3
```
joaquinbejar, line 89: 🟠 **Warning**: this swaps the stateful MongoDB binary under the existing `mongodb-data` volume, but the green integration job only starts an empty 8.2 volume. MongoDB's 8.0 to 8.2 procedure requires the source instance to run 8.0 with FCV 8.0 before replacing the binary, so an existing deployment has an unverified upgrade path. Add a persisted volume upgrade test and document the FCV preflight and rollout step before changing the production default.

.github/workflows/ci.yml
joaquinbejar: You are right, a mutable tag there means green CI stops proving anything about the bytes a deployment runs. All three service containers now carry the same digests the compose file pins, with a comment saying why.

Docker/docker-compose.yml
joaquinbejar: Fair, an empty volume proves the image works and says nothing about upgrading one that holds data. I ran the path against this exact digest: 8.0.29 seeded with a document at FCV 8.0, stopped, image swapped, 8.2.12 came up with the document intact and the FCV still 8.0, then setFeatureCompatibilityVersion to 8.2 succeeded. The order is documented next to the image, including which step is the reversible one, and issue #130 turns that manual run into a job so the next bump cannot skip it.

joaquinbejar:
Both are addressed in 196bebc. The upgrade one was worth the time: the path works, and now the order and the point of no return are written down next to the image rather than assumed.

Issue #117, opened 2026-08-30
Decide whether the integration harness may depend on an HTTP client


joaquinbejar:
From PR for #99, where the decision was needed and could not be taken autonomously.

## Context

`examples/integration` has to speak HTTP to a deployed service, and this workspace has no HTTP client: no `reqwest`, no `awc`, no `ureq`. `rules/global_rules.md` forbids adding a dependency without the owner explicit approval, so the harness ships with a small hand-written HTTP/1.1 client over `std::net::TcpStream` instead: request with an optional JSON body, response with headers, length-delimited or chunked body.

That was the conservative choice, not the preferred one. What it costs:

- **No HTTPS.** A TLS client is itself a dependency, so a deployment behind TLS cannot be tested. The harness refuses an `https://` URL with a message that says why rather than failing obscurely later.
- **A protocol implementation to maintain**, roughly 150 lines that nobody wants to own, and that will grow the first time a test needs redirects, keep-alive, or a streamed body read incrementally rather than whole.
- **No connection reuse**, so every request opens a socket. Fine for a test suite, wasteful at #107 concurrency scale.

## The decision

Approve `reqwest` (blocking feature, pinned `X.X` in `[workspace.dependencies]` like everything else, used ONLY by `examples/integration`, which is `publish = false`), or confirm the hand-rolled client stays.

`tokio` is already a workspace dependency and optionstratlib `async` feature already pulls `reqwest` into the graph, so the build cost of approving it is close to zero; the question is the policy one.

## Acceptance criteria

- [ ] A decision is recorded here
- [ ] If approved: `reqwest` in `[workspace.dependencies]` as `X.X`, `examples/integration` switched to it, the hand-rolled client deleted, HTTPS support noted in the crate docs
- [ ] If refused: this issue closes with the reason, and the harness documentation states that HTTPS deployments cannot be tested

PR #118, opened 2026-08-30
An integration harness for testing a deployed service

### LINKED ISSUE #99: Integration test harness: one crate, one environment variable, skipped when absent

joaquinbejar:
Closes #99

Base branch: `stack/4-112-service-images`
Stacked on: #116
Stack: #109 → #110 → #111 → #112 → **#99** → #104 → #100 → #101 → #102 → #103 → #106 → #105 → #107

The diff is cumulative until the PRs below merge, so read it against the base branch rather than against `main`. This is the one the eight after it depend on.

## What it is

`examples/integration`, a new workspace member. `OCS_INTEGRATION_BASE_URL` names the service under test, scheme and port included; where it points is an operator configuration and is never recorded here. Unset or blank, and blank is unset by the same rule the service applies to its own knobs, every test skips and prints why.

It gives the eight test issues that follow three things: a client that owns the base URL so no test writes one by hand, a `Simulation` fixture that deletes itself on drop even after a panic, and a reference create request to vary one field of.

## A decision I took rather than asked

The issue flagged that this needs an HTTP client and that the rules forbid adding a dependency without your approval. I did not add one. The harness speaks HTTP/1.1 over `std::net::TcpStream`: a request with an optional JSON body, a response with headers and a length-delimited or chunked body, which is what an export streams.

What that costs, and why it is written down in #117 rather than buried here: **no HTTPS**, since a TLS client is itself a dependency, so an `https://` base URL is refused with a message saying so; about 150 lines of protocol nobody wants to own; and a socket per request. #117 asks you to approve `reqwest` or confirm the hand-rolled client stays. Either way this PR works today.

## A deployed build is not `main`

The deployment these were designed against serves 0.2.0, so `/health`, the binary export formats, the spread model and `strike_ladder` are all absent there. The rule the suite follows is to state which contract it is testing and skip loudly when a feature is not deployed, never to fail for it. The smoke test reads the version from the OpenAPI document and prints it.

## Tests

- Four harness tests that run in the hermetic suite: no client when nothing is configured, a configuration mistake refused with the reason and the URL, base URL normalisation, and an unreachable service naming the URL rather than surfacing a bare timeout
- Two deployment tests, skipped without the variable: the service answers and names its version, and a simulation created by the fixture is gone once it drops

The only socket the hermetic run opens is a loopback connect to a closed port, which is how the unreachable message is proven; nothing reaches a network.

## Checklist

- [x] `LOGLEVEL=WARN cargo test --workspace` green with the variable unset
- [x] Library untouched; the HTTP code lives in a `publish = false` workspace member
- [x] `make test-integration` and a manual or weekly CI job, never on push
- [x] `.env.example` and the crate docs describe the variable and the lag caveat
- [x] `make pre-push` clean, zero warnings; README regenerated from `lib.rs`
- [x] Patch version bumped to 0.2.6 with the compose image tag


examples/integration/tests/harness.rs
```diff
@@ -0,0 +1,100 @@
...
+/// Port 1 on the loopback interface refuses immediately, so this is a fast,
+/// hermetic check of the message rather than a wait on a timeout.
+#[test]
+fn test_an_unreachable_service_names_the_url() {
+    let client = match ServiceClient::new("http://127.0.0.1:1") {
+        Ok(client) => client,
+        Err(error) => panic!("the URL is well formed: {error}"),
+    };
+
+    match client.get("/health") {
```
joaquinbejar: 🟠 **Warning**: this test always attempts a TCP connection even when the integration variable is unset, so `cargo test --workspace` is not network free as issue #99 and the new docs promise. Test the error wrapping without dialing a socket, or move this transport check behind the configured deployment path.

examples/integration/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+        body: Option<&str>,
+    ) -> Result<Response, IntegrationError> {
+        let url = format!("{}{path}", self.base_url);
+        let unreachable = |cause: String| IntegrationError::Unreachable {
+            url: url.clone(),
+            cause,
+        };
+
+        let stream =
+            TcpStream::connect(&self.address).map_err(|error| unreachable(error.to_string()))?;
```
joaquinbejar: 🟠 **Warning**: `REQUEST_TIMEOUT` is applied only after `TcpStream::connect` returns. A black holed address can therefore hang the scheduled job for the operating system TCP timeout instead of 60 seconds. Resolve the address and use `connect_timeout` under a single request deadline.

examples/integration/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+    };
+
+    let mut body = Vec::new();
+    loop {
+        let mut size_line = String::new();
+        reader
+            .read_line(&mut size_line)
+            .map_err(|error| unreachable(error.to_string()))?;
+        let size_text = size_line.trim();
+        if size_text.is_empty() {
```
joaquinbejar, line 450: 🟠 **Warning**: EOF produces an empty size line here and is accepted as a successful end of body, even though a chunked response is complete only after a parsed zero size chunk. A dropped export can therefore look like valid partial data. Treat EOF or a blank size line as `MalformedResponse`, validate each chunk terminator, and require the final zero chunk.

examples/integration/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+        format!("/api/v2/simulations/{}{suffix}", self.id)
+    }
+}
+
+impl Drop for Simulation {
+    fn drop(&mut self) {
+        // A failing test must not leave a simulation behind on a shared
+        // deployment, and a cleanup that itself fails must not mask the
+        // failure that is already being reported.
+        if let Err(error) = self.client.delete(&self.path("")) {
```
joaquinbejar: 🟠 **Warning**: `delete` returns `Ok(Response)` even for HTTP 401, 405, or 500, so `Drop` silently treats failed cleanup as success and can leak simulations on the shared deployment. Inspect the status, accept only the documented success and optionally 404, and log status plus body for every other response without panicking.

examples/integration/tests/deployment.rs
```diff
@@ -0,0 +1,113 @@
...
+    };
+
+    // `/api/v1/chain` is the oldest route in the service and exists in every
+    // build, so it is the safest probe for "is anything there".
+    let response = match client.get("/api/v1/chain") {
+        Ok(response) => response,
+        Err(error) => panic!("{error}"),
+    };
+
+    assert!(
```
joaquinbejar: 🟠 **Warning**: any status below 500 passes this identity probe, so an unrelated server returning 404 is accepted and the missing OpenAPI document and failed v2 creation are then reported as skips. Require an unmistakable OptionChain Simulator response or a valid OpenAPI title and version, and only skip features after explicit version detection.

joaquinbejar (commented):
The new harness is cohesive despite exceeding 500 lines, but the default suite opens a socket, connection setup is outside the promised timeout, truncated chunked bodies can pass as complete, cleanup ignores HTTP failures, and the deployment probe can pass against the wrong service. The formal decision is REQUEST_CHANGES.

examples/integration/tests/harness.rs
joaquinbejar: You are right, a loopback connect is still a connection and the promise was that the hermetic run opens none. The message is a property of the error value, so it is asserted on the value now, for the malformed case too.

examples/integration/src/lib.rs
joaquinbejar: Good catch, a black holed address would have hung the scheduled job for the operating system timeout rather than sixty seconds. The address is resolved and dialled with connect_timeout under the same deadline as the read and the write.

examples/integration/src/lib.rs
joaquinbejar: That one would have defeated the purpose of the suite, since a truncated export is precisely what it exists to catch. End of stream, a blank size line and a chunk not followed by a line end are all malformed now, and only the zero size chunk ends a body.

examples/integration/src/lib.rs
joaquinbejar: Agreed, a 401 or a 500 counting as cleanup is how a shared deployment fills up with abandoned simulations. Only the documented success codes and 404 pass now, and anything else prints the status and the body without panicking so it cannot mask the failure already being reported.

examples/integration/tests/deployment.rs
joaquinbejar: Right, every skip in the suite assumes a 404 means an older build of this service, which only holds if the identity is established first. The probe now requires an OpenAPI document whose title names this service, and for a build too old to serve one it requires the rejection shape only this service produces.

joaquinbejar:
All five are fixed in c32f4dc. Two of them were the harness lying about its own guarantees, the hermetic socket and the chunked body that ended at EOF, so they were worth more than their size.

Issue #119, opened 2026-08-30
v1 rejections do not carry the field the error shape promises


joaquinbejar:
Found by the #104 integration tests, running against a live service.

## What the deployment answers

```
GET /api/v1/chain?sessionid=not-a-uuid
400 {"error":"Invalid State: Invalid session ID format"}

GET /api/v1/chain
400 {"error":"Query deserialize error: missing field `sessionid`","field":""}
```

Two things differ from what v2 does with the same class of mistake:

- The malformed-id body carries **no `field` key at all**, where the documented rejection shape is `{error, field}`. v2 answers `{"error":"Validation Error: id: must be a UUID, got \"not-a-uuid\"","field":"id"}` for the same mistake.
- The message is prefixed `Invalid State`, which is the name of a lifecycle error, for what is a request-validation failure.

## Why this is a decision rather than a fix

`/api/v1/chain` is frozen on RENDERED VALUES, not only on shapes (ADR 0001 §12.1). Adding a `field` key to a v1 error body changes what an existing client parses, and a client matching on the `Invalid State` prefix would break if the message were corrected. So this is a contract question:

1. Add `field` to v1 rejections and correct the prefix, accepting that it is a v1 wire change; or
2. Record that v1 rejections are `{error}` only, and document the difference from v2 in the crate docs and the OpenAPI document.

The #104 integration test currently asserts the CURRENT behaviour and points here, so whichever way this goes, the test moves with it.

## Acceptance criteria

- [ ] A decision recorded here
- [ ] The v1 rejection shape documented as whatever it is, in the crate docs and the OpenAPI document
- [ ] `examples/integration/tests/rejections.rs` asserts the decided shape rather than the observed one
- [ ] If the shape changes: a note for IronCondor, since it consumes v1

PR #120, opened 2026-08-30
Assert the rejection contract against a deployment, field by field

### LINKED ISSUE #104: Integration tests: the rejection contract, field by field

joaquinbejar:
Closes #104

Base branch: `stack/5-99-integration-harness`
Stacked on: #118
Blocks: #121

Stack: #109 → #110 → #111 → #112 → #99 → **#104** → #100 → #101 → #102 → #103 → #106 → #105 → #107

## What it asserts

Nine v2 rejection cases, each by status AND by field, never by status alone: a malformed id, a reversed export range, an unknown greek level, an unknown query key, an unknown body key, and the bounds on `steps`, `initial_price`, `volatility` and `step_interval_seconds`. Plus a missing simulation answering 404 on every verb that takes an id, and an unparseable body still answering the documented shape.

Three things are asserted for every case rather than per case, because they are what an unhandled path breaks first: the status is a 4xx and never a 500, the content type is JSON and not the plaintext actix produces on its own, and the body carries no URL, no credential, no internal path and not the deployment host name.

## What running it against a live service changed

I ran these against a real instance rather than writing them from the DTOs, which corrected two expectations and found one gap.

- A reversed range names **`from_step`**, not `to_step`. The service names the field the client can act on, which is the one that exceeds. My first expectation was wrong, not the service.
- **v1 answers a malformed `sessionid` with no `field` key at all**, where v2 answers `field: "id"` for the same mistake, and prefixes the message `Invalid State` for what is a validation failure. `/api/v1/chain` is frozen on rendered values, so adding a key to its error body is a wire change a client could break on. That is a decision, not a fix I should make here: it is issue #119, and this test asserts today behaviour while printing a line that points at it.

## Checklist

- [x] Every rejection asserted by status and by field
- [x] A 500 anywhere in the tables fails the test
- [x] Verified against a running service: 9 v2 cases exercised, 4 tests green
- [x] Patch version bumped to 0.2.7 with the compose image tag

### DIFF
--- a/examples/integration/tests/rejections.rs
+++ b/examples/integration/tests/rejections.rs
@@ -33,11 +33,14 @@ struct Case {
 }
 
 /// The documented rejection body.
+///
+/// `field` is an `Option` rather than a defaulted `String` on purpose: a body
+/// that omits the key and one that carries an empty string are different
+/// facts, and #119 is about exactly that difference on the v1 surface.
 #[derive(Debug, Deserialize)]
 struct Rejection {
     error: String,
-    #[serde(default)]
-    field: String,
+    field: Option<String>,
 }
 
 /// Reads a rejection body, asserting the things that hold for EVERY rejection
@@ -112,13 +115,45 @@ fn test_the_v2_rejection_contract_names_its_field() {
         return;
     };
 
+    // A deployment without the v2 API answers 404 for the route itself, which
+    // is the only creation failure that means "not deployed here". Anything
+    // else — a 500, a wrong success status, a body that will not decode — is a
+    // defect, and turning it into a green skip is how a suite quietly stops
+    // testing anything.
+    let probe = match client.post("/api/v2/simulations", &reference_request("SPX")) {
+        Ok(probe) => probe,
+        Err(error) => panic!("creating a simulation: {error}"),
+    };
+    if probe.status == 404 {
+        println!("SKIP: this deployment does not serve /api/v2/simulations");
+        return;
+    }
+    assert_eq!(
+        probe.status,
+        201,
+        "creating a simulation must answer 201, got {} with {}",
+        probe.status,
+        probe.text()
+    );
     let simulation = match Simulation::create(&client, &reference_request("SPX")) {
         Ok(simulation) => simulation,
-        Err(error) => {
-            println!("SKIP: this deployment has no usable v2 API: {error}");
-            return;
-        }
+        Err(error) => panic!("the v2 API is deployed, so creating a simulation must work: {error}"),
     };
+    // The probe created one too; delete it rather than leaving it behind.
+    if let Ok(Some(id)) = probe
+        .json::<serde_json::Value>("/api/v2/simulations")
+        .map(|body| {
+            body.get("id")
+                .and_then(|id| id.as_str())
+                .map(str::to_string)
+        })
+    {
+        examples_integration::report_cleanup(
+            &client,
+            &format!("/api/v2/simulations/{id}"),
+            "the probe simulation",
+        );
+    }
 
     let cases = vec![
         Case {
@@ -188,12 +223,42 @@ fn test_the_v2_rejection_contract_names_its_field() {
             field: "step_interval_seconds",
         },
         Case {
+            // Issue #104 names this one explicitly: a range whose end is past
+            // the tape must name `to_step`, which is the field at fault here,
+            // unlike the inverted range above.
+            what: "a range past the end of the tape",
+            method: "GET",
+            path: simulation
+                .path("/export?dataset=underlying&format=csv&from_step=0&to_step=99999"),
+            body: None,
+            status: 400,
+            field: "to_step",
+        },
+        Case {
+            what: "steps above the cap",
+            method: "POST",
+            path: "/api/v2/simulations".to_string(),
+            body: Some(request_with("steps", serde_json::json!(10_000_000))),
+            status: 400,
+            field: "steps",
+        },
+        Case {
+            what: "a chain size above the cap",
+            method: "POST",
+            path: "/api/v2/simulations".to_string(),
+            body: Some(request_with("chain_size", serde_json::json!(10_000_000))),
+            status: 400,
+            field: "chain_size",
+        },
+        Case {
             what: "an unknown body key",
             method: "POST",
             path: "/api/v2/simulations".to_string(),
             body: Some(request_with("stepss", serde_json::json!(4))),
             status: 400,
-            field: "",
+            // The handler recovers the offending key, so the contract is that
+            // it names it rather than shrugging.
+            field: "stepss",
         },
     ];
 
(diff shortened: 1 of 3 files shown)


examples/integration/tests/rejections.rs
```diff
@@ -0,0 +1,381 @@
...
+#[test]
+fn test_the_v2_rejection_contract_names_its_field() {
+    let Some(client) = service() else {
+        return;
+    };
+
+    let simulation = match Simulation::create(&client, &reference_request("SPX")) {
+        Ok(simulation) => simulation,
+        Err(error) => {
+            println!("SKIP: this deployment has no usable v2 API: {error}");
```
joaquinbejar: 🟠 **Warning**: every creation failure, including transport errors, 500 responses, wrong success statuses, and malformed JSON from an existing v2 deployment, becomes a green skip. The later 404 and 405 branches repeat this after creation has already proved the base route exists. Gate only genuinely absent optional routes from an explicitly detected deployment version, and fail all other errors and unexpected route statuses.

examples/integration/tests/rejections.rs
```diff
@@ -0,0 +1,381 @@
...
+            return;
+        }
+    };
+
+    let cases = vec![
```
joaquinbejar, line 158: 🟠 **Warning**: issue #104 requires `steps` above its cap, `chain_size` above its cap, and a range past the tape that names `to_step`, but this table has none of those rows while claiming the bounds are covered field by field. Add all three cases with values that reliably exceed the configured limits.

examples/integration/tests/rejections.rs
```diff
@@ -0,0 +1,381 @@
...
+    for case in cases {
+        let Case {
+            what,
+            method,
+            path,
+            body,
+            status,
+            field,
+        } = case;
+        let rendered_body = body.map(|body| body.to_string());
```
joaquinbejar, line 275: 🟠 **Warning**: if one of these invalid create requests regresses to 200 or 201, `rejection()` panics after the returned id has been discarded, leaving a simulation on the shared deployment. Detect unexpected successful creates, wrap or delete the returned id, then fail the assertion so cleanup still holds on the exact regression under test.

examples/integration/tests/rejections.rs
```diff
@@ -0,0 +1,381 @@
...
+            continue;
+        }
+
+        let rejection = rejection(&client, what, &response);
+        assert_eq!(
+            response.status, status,
+            "{what} must answer {status}, got {} with {}",
+            response.status, rejection.error
+        );
+        if !field.is_empty() {
```
joaquinbejar: 🟠 **Warning**: this skips the field assertion exactly for cases that expect an empty field, and `#[serde(default)]` above also makes a missing field indistinguishable from an empty one. The unknown body key is therefore checked by status only and currently ignores the handler's recovered `stepss` field. Assert the expected field unconditionally and require the key wherever the v2 shape promises it.

examples/integration/tests/rejections.rs
```diff
@@ -0,0 +1,381 @@
...
+            Ok(response) => response,
+            Err(error) => panic!("{method} {path}: {error}"),
+        };
+
+        if response.status == 405 {
+            println!("SKIP: {method} {path} is not deployed here");
+            continue;
+        }
+
+        assert_eq!(
```
joaquinbejar, line 353: 🟠 **Warning**: this table bypasses `rejection()`, so a plaintext 404 or a body with no documented error shape still passes despite the PR saying JSON content type and the common shape are checked for every rejection. Apply the shared assertions here too, adapted for the intentionally unfielded not found response.

examples/integration/tests/rejections.rs
```diff
@@ -0,0 +1,381 @@
...
+        let body: serde_json::Value = match response.json(path) {
+            Ok(body) => body,
+            Err(error) => panic!("{case} must answer JSON: {error}"),
+        };
+        let message = body.get("error").and_then(|error| error.as_str());
+        assert!(
+            message.is_some_and(|message| !message.is_empty()),
+            "{case} must explain itself: {body}"
+        );
+        if body.get("field").is_none() {
```
joaquinbejar: 🟠 **Warning**: the comment says this deliberately asserts today's v1 wire behavior, but the code accepts either presence or absence of `field` and never checks the `Invalid State` prefix. A breaking shape or rendered value change therefore stays green. Encode the two per case expectations described above, then change them deliberately when #119 is decided.

joaquinbejar (commented):
The rejection tables cover useful cases, but deployment failures and route regressions can be skipped, required bounds are missing, several paths do not assert the promised field and JSON shape, an unexpectedly accepted create can leak state, and the stated v1 behavior is not pinned. The formal decision is REQUEST_CHANGES.

examples/integration/tests/rejections.rs
joaquinbejar: Agreed, that turned every defect into a green skip. Only a 404 on the route itself counts as not deployed now; a 500, a wrong success status or a body that will not decode all fail, and the probe simulation it creates is deleted rather than left behind.

examples/integration/tests/rejections.rs
joaquinbejar: You are right, the PR claimed coverage the table did not have. All three rows are in: steps above its cap, chain_size above its cap, and a range past the end of the tape, which names to_step where an inverted range names from_step. Twelve cases exercised against a live service now.

examples/integration/tests/rejections.rs
joaquinbejar: Good catch, the exact regression under test would have leaked a simulation on a shared deployment. The id is recovered and deleted first, then the failure is raised.

examples/integration/tests/rejections.rs
joaquinbejar: Right on both halves. The field is an Option now, so a missing key and an empty one are different facts, and every case asserts it unconditionally including the unknown body key, which the handler does recover and name.

examples/integration/tests/rejections.rs
joaquinbejar: Fixed, the not-found table goes through the shared assertions now, and it also asserts that a not-found carries no field at all, since no request field is at fault there.

examples/integration/tests/rejections.rs
joaquinbejar: Fair, a comment claiming to pin behaviour while the code accepted either shape is worse than no test. It now asserts exactly what a live service returns, no field key and an Invalid State prefix for a malformed id and an empty field for a missing one, and both change deliberately when #119 is decided.

joaquinbejar:
All six are fixed in 2ab1172, verified against a live service with twelve v2 cases now exercised. Two of them were the test claiming more than it checked, which is the worse kind of green.

PR #121, opened 2026-08-30
Cover the v1 session lifecycle against a deployment

### LINKED ISSUE #100: Integration tests: the v1 session lifecycle over the wire

joaquinbejar:
Closes #100

Base branch: `stack/6-104-rejection-contract`
Stacked on: #120
Blocks: #122

Stack: #109 → #110 → #111 → #112 → #99 → #104 → **#100** → #101 → #102 → #103 → #106 → #105 → #107

## What it asserts

Five flows, each creating and deleting its own session so none depends on another state:

- creation echoes the parameters and an **effective seed**, generated because the request omitted one, and that seed does not change under the client. Without the echo a client cannot record a run and replay it, which is the contract IronCondor gates on
- a peek is safe and repeatable: same price, same ladder, cursor unmoved
- the serve-then-advance walk, with the `410` boundary pinned exactly. A session with N steps serves N snapshots; advance N+1 is `410 Gone`, and so is every peek afterwards
- `PATCH` changes one parameter and `PUT` restarts the walk, neither creating a new session
- a deleted session is `404` on every verb, never `500`

I ran these against a live instance rather than writing them from the handlers. That is where the `410` boundary got pinned to the exact advance: with `steps = 2`, the second advance still serves a snapshot and the peek right after it is already `410`, so the session completes on the advance that serves the last step rather than on the one after.

## Checklist

- [x] One test per verb plus the full walk
- [x] The effective seed asserted present and stable
- [x] A missing session is 404 on every verb
- [x] Sessions kept small, a handful of steps and a two-strike chain, since the deployment is shared
- [x] Patch version bumped to 0.2.8 with the compose image tag

### DIFF
--- a/examples/integration/tests/v1_lifecycle.rs
+++ b/examples/integration/tests/v1_lifecycle.rs
@@ -37,12 +37,24 @@ struct Snapshot {
     underlying: String,
     price: f64,
     contracts: Vec<Contract>,
+    /// The cursor the snapshot itself reports, which is the public contract a
+    /// client reads to know where the walk is; comparing prices proves nothing
+    /// about it.
+    session_info: SessionInfo,
+}
+
+/// The cursor a snapshot carries.
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+struct SessionInfo {
+    current_step: usize,
+    total_steps: usize,
 }
 
 /// One contract in a snapshot.
 #[derive(Debug, Deserialize)]
 struct Contract {
     strike: f64,
+    implied_volatility: Option<f64>,
 }
 
 /// A v1 session that deletes itself, so a failing test leaves nothing behind.
@@ -54,11 +66,7 @@ struct Session {
 impl Session {
     /// Creates a session with `steps` steps and a deliberately narrow chain,
     /// since these run against a shared deployment.
-    fn create(
-        client: &ServiceClient,
-        steps: usize,
-        seed: Option<u64>,
-    ) -> Option<(Self, SessionEnvelope)> {
+    fn create(client: &ServiceClient, steps: usize, seed: Option<u64>) -> (Self, SessionEnvelope) {
         let mut request = serde_json::json!({
             "symbol": "AAPL",
             "steps": steps,
@@ -80,10 +88,11 @@ impl Session {
             Ok(response) => response,
             Err(error) => panic!("creating a v1 session: {error}"),
         };
-        if response.status == 404 {
-            println!("SKIP: this deployment does not serve /api/v1/chain");
-            return None;
-        }
+        // No skip here on purpose. With the variable configured, a 404 on
+        // /api/v1/chain is precisely the route regression this suite exists to
+        // catch, and skipping on it would make every test in this file green
+        // the day the route stopped being mounted. The deployment's identity
+        // is established in deployment.rs before any of this runs.
         assert_eq!(
             response.status,
             201,
@@ -100,7 +109,7 @@ impl Session {
             client: client.clone(),
             id: envelope.id.clone(),
         };
-        Some((session, envelope))
+        (session, envelope)
     }
 
     /// The query every v1 verb takes.
@@ -128,9 +137,11 @@ impl Session {
 
 impl Drop for Session {
     fn drop(&mut self) {
-        if let Err(error) = self.client.delete(&self.query()) {
-            println!("WARNING: could not delete session {}: {error}", self.id);
-        }
+        // `delete` answers Ok for a 401, a 405 or a 500 too, and treating
+        // those as cleanup is how a shared deployment fills up. `report_cleanup`
+        // accepts only the documented successes and 404, and never panics,
+        // since this runs during unwinding.
+        examples_integration::report_cleanup(&self.client, &self.query(), &self.id);
     }
 }
 
@@ -142,9 +153,7 @@ fn test_creating_a_session_echoes_an_effective_seed() {
     let Some(client) = service() else {
         return;
     };
-    let Some((session, envelope)) = Session::create(&client, 3, None) else {
-        return;
-    };
+    let (session, envelope) = Session::create(&client, 3, None);
 
     assert_eq!(envelope.parameters.symbol, "AAPL");
     assert_eq!(envelope.total_steps, 3);
@@ -179,9 +188,7 @@ fn test_a_peek_is_repeatable_and_does_not_advance() {
     let Some(client) = service() else {
         return;
     };
-    let Some((session, _)) = Session::create(&client, 3, Some(42)) else {
-        return;
-    };
+    let (session, _) = Session::create(&client, 3, Some(42));
 
     let first = session.peek();
     let second = session.peek();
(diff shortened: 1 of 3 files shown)


examples/integration/tests/v1_lifecycle.rs
```diff
@@ -0,0 +1,391 @@
...
+        });
+        if let (Some(object), Some(seed)) = (request.as_object_mut(), seed) {
+            object.insert("seed".to_string(), serde_json::json!(seed));
+        }
+
+        let response = match client.post("/api/v1/chain", &request) {
+            Ok(response) => response,
+            Err(error) => panic!("creating a v1 session: {error}"),
+        };
+        if response.status == 404 {
```
joaquinbejar: 🟠 Warning Once `OCS_INTEGRATION_BASE_URL` is configured, this 404 is the route regression the suite is meant to catch. Issue #100 permits skipping only when the base URL is unset and requires a changed route to fail; as written, removing or mis-mounting `/api/v1/chain` makes all five tests green. Remove this branch and require the documented 201 response.

examples/integration/tests/v1_lifecycle.rs
```diff
@@ -0,0 +1,391 @@
...
+        match self.client.request("POST", &path, None) {
+            Ok(response) => response,
+            Err(error) => panic!("advancing: {error}"),
+        }
+    }
+}
+
+impl Drop for Session {
+    fn drop(&mut self) {
+        if let Err(error) = self.client.delete(&self.query()) {
```
joaquinbejar: 🟠 Warning `ServiceClient::delete` returns `Ok(Response)` for HTTP failures, so a 401, 405, or 500 is silently treated as successful cleanup and the test can leak a session on the shared deployment. Inspect the response status here, accepting 200 and the expected 404 after an explicit deletion, and report the status and body for any other result without panicking during unwinding.

examples/integration/tests/v1_lifecycle.rs
```diff
@@ -0,0 +1,391 @@
...
+#[derive(Debug, Deserialize)]
+struct Parameters {
+    symbol: String,
+    volatility: f64,
+    seed: Option<u64>,
+}
+
+/// One snapshot of the chain.
+#[derive(Debug, Deserialize)]
+struct Snapshot {
```
joaquinbejar, line 36: 🟠 Warning This DTO drops `session_info.current_step`, so the peek test never observes whether GET moved the cursor and the walk test never verifies the served cursor sequence. Comparing price and strikes does not establish the public cursor contract. Deserialize `session_info`, assert both peeks remain at step 0, and assert each advance reports the expected post-serve cursor from 1 through N and serves the snapshot that was peeked.

examples/integration/tests/v1_lifecycle.rs
```diff
@@ -0,0 +1,391 @@
...
+        "days_to_expiration": 30.0,
+        "volatility": 0.25,
+        "risk_free_rate": 0.05,
+        "dividend_yield": 0.01,
+        "method": {"GeometricBrownian": {"dt": 0.004, "drift": 0.05, "volatility": 0.25}},
+        "time_frame": "Day",
+        "chain_size": 2,
+        "strike_interval": 5.0
+    });
+    match client.request("PUT", &session.query(), Some(&replacement.to_string())) {
```
joaquinbejar: 🟠 Warning Both mutations are issued while the cursor is 0 and the test stops at response envelopes, making `replaced.current_step == 0` vacuous and allowing a stale cached walk to pass. Advance before PATCH, assert its reset and that the next snapshot reflects the new volatility, then advance before PUT and assert its next snapshot starts from the replacement values such as `initial_price == 120.0`; this is the behavior issue #100 requires over the wire.

examples/integration/tests/v1_lifecycle.rs
```diff
@@ -0,0 +1,391 @@
...
+        Ok(response) => assert_eq!(
+            response.status,
+            200,
+            "DELETE must answer 200, got {}",
+            response.text()
+        ),
+        Err(error) => panic!("{error}"),
+    }
+
+    for (method, path) in [
```
joaquinbejar: 🟠 Warning The test says every verb but omits PATCH and PUT, so either missing-session mutation route may regress to 200 or 500 while this test stays green. Send minimal valid bodies for both methods and require 404 after deletion.

joaquinbejar (commented):
Formal decision REQUEST_CHANGES. The incremental diff contains 395 changed lines and is reasonably scoped, with CI green, but the deployed lifecycle suite can still pass when the v1 route is missing, cannot observe cursor movement, does not exercise PATCH and PUT after progression or verify their next snapshots, omits two verbs from deletion coverage, and can silently leak sessions when cleanup returns an HTTP error. These contract gaps need correction before this stacked PR is ready.

examples/integration/tests/v1_lifecycle.rs
joaquinbejar: You are right, that skip would have made all five tests green the day the route stopped being mounted. The create must answer 201 now, and identifying the deployment is deployment.rs job rather than this one.

examples/integration/tests/v1_lifecycle.rs
joaquinbejar: Fixed, it goes through report_cleanup now, which accepts only the documented successes and 404 and reports anything else without panicking during unwinding.

examples/integration/tests/v1_lifecycle.rs
joaquinbejar: Good point, comparing prices could also pass on a cached body. The snapshot carries session_info now: two peeks stay at step 0, and each advance serves the market the peek showed and reports the cursor after serving, 1 through N.

examples/integration/tests/v1_lifecycle.rs
joaquinbejar: Agreed on both counts. Each mutation now happens after the walk has moved, so the reset is real, and each is checked in the next snapshot rather than the envelope: the patched volatility reaches the implied volatility, and the replacement initial_price of 120 is what the next snapshot starts from.

examples/integration/tests/v1_lifecycle.rs
joaquinbejar: Right, a mutation route regressing on a deleted session is as bad as a read doing it. PATCH and PUT are in the table with valid bodies and must answer 404.

joaquinbejar:
All five are fixed and verified against a live service. The cursor one changed the shape of the tests: they assert what the snapshot reports rather than inferring it from prices.

PR #122, opened 2026-08-30
Cover the v2 lifecycle and its cursor semantics against a deployment

### LINKED ISSUE #101: Integration tests: the v2 simulation lifecycle and its cursor semantics

joaquinbejar:
Closes #101

Base branch: `stack/7-100-v1-lifecycle`
Stacked on: #121
Blocks: #123

Stack: #109 → #110 → #111 → #112 → #99 → #104 → #100 → **#101** → #102 → #103 → #106 → #105 → #107

## What it asserts

Six flows, each deleting its own simulation:

- **creation echoes every replay input**: seed, resolved start, resolved interval, tzdb release, the normalised schedules sorted by rule id, and the chain parameters. A client that recorded only what it sent cannot reproduce a run, so this is the reproducibility contract seen from outside. A request without a seed is told which one it got
- reading metadata and peeking a snapshot move neither the cursor nor the version
- the walk serves every step, the advance past the last is `410`, and the simulation ends `completed`
- **`412` and `409` are asserted as different facts, not conflated.** A stale `expected_step` is `412`, its body carries the ACTUAL cursor, the version does not move, and the retry at the cursor it reported serves. That is the whole point of the precondition: a client retrying after a timeout must not lose a step to a wrong guess
- two advances fired at once from two threads answer `200` or `409`, never a 5xx, and the cursor moves exactly once per winner
- a deleted simulation stops resolving on every endpoint

The concurrency test genuinely raced there rather than serialising: one advance won and the other got `409`, with the cursor moving once. The test accepts either outcome, since a deployment may serialise them, but it fails on a 5xx or on a cursor that moved by anything other than the number of winners.

## Checklist

- [x] `412` and `409` distinguished
- [x] The cursor and the version asserted after every failed precondition: nothing consumed
- [x] Normalised schedules asserted sorted by rule id
- [x] Each test cleans up its simulation, including on failure
- [x] Patch version bumped to 0.2.9 with the compose image tag

### DIFF
--- a/examples/integration/tests/v2_lifecycle.rs
+++ b/examples/integration/tests/v2_lifecycle.rs
@@ -1,0 +1,495 @@
+//! The v2 simulation lifecycle and its cursor semantics, over the wire.
+//!
+//! v2 carries the vocabulary that makes a retry safe, and it is the part a
+//! unit test proves least: `expected_step` is a PRECONDITION on the cursor,
+//! and the version is a compare-and-swap token. They mean different things and
+//! answer different codes, `412` and `409`, so these assert them separately
+//! and check the cursor after every failed precondition to prove nothing was
+//! consumed.
+//!
+//! Skipped entirely when `OCS_INTEGRATION_BASE_URL` is unset.
+
+use examples_integration::{ServiceClient, reference_request, service};
+use serde::Deserialize;
+use std::sync::mpsc;
+use std::thread;
+
+/// The simulation envelope: identity, state, the CAS token and the cursor.
+#[derive(Debug, Deserialize)]
+struct Envelope {
+    id: String,
+    state: String,
+    version: u64,
+    cursor: Cursor,
+}
+
+/// Where the walk is and how far it goes.
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+struct Cursor {
+    current_step: usize,
+    total_steps: usize,
+}
+
+/// The create response, which echoes every replay input.
+#[derive(Debug, Deserialize)]
+struct Created {
+    id: String,
+    state: String,
+    cursor: Cursor,
+    parameters: Parameters,
+}
+
+/// The replay inputs a client has to record to reproduce a run.
+#[derive(Debug, Deserialize)]
+struct Parameters {
+    symbol: String,
+    steps: usize,
+    seed: u64,
+    effective_start: String,
+    step_interval_seconds: u64,
+    tzdb_version: String,
+    schedules: Vec<Schedule>,
+    chain_size: Option<usize>,
+}
+
+/// One normalised schedule rule.
+#[derive(Debug, Deserialize)]
+struct Schedule {
+    rule_id: String,
+    target_count: usize,
+}
+
+/// A simulation that deletes itself.
+///
+/// The harness fixture creates one and hides the body; these tests need the
+/// body, since the create response is where every replay input is echoed, so
+/// they own the request here and keep the same delete-on-drop guarantee.
+struct Live {
+    client: ServiceClient,
+    id: String,
+}
+
+impl Live {
+    /// Creates a simulation with `steps` steps and a two-strike chain, kept
+    /// small because the deployment is shared.
+    fn create(client: &ServiceClient, steps: usize) -> Option<(Self, Created)> {
+        let mut request = reference_request("SPX");
+        if let Some(object) = request.as_object_mut() {
+            object.insert("steps".to_string(), serde_json::json!(steps));
+            object.insert("chain_size".to_string(), serde_json::json!(2));
+        }
+
+        let response = match client.post("/api/v2/simulations", &request) {
+            Ok(response) => response,
+            Err(error) => panic!("{error}"),
+        };
+        if response.status == 404 {
+            println!("SKIP: this deployment has no v2 API");
+            return None;
+        }
+        assert_eq!(
+            response.status,
+            201,
+            "creating a simulation must answer 201, got {} with {}",
+            response.status,
+            response.text()
+        );
+
+        let created: Created = match response.json("/api/v2/simulations") {
+            Ok(created) => created,
+            Err(error) => panic!("{error}"),
+        };
+        let live = Self {
+            client: client.clone(),
+            id: created.id.clone(),
+        };
+        Some((live, created))
+    }
+
+    /// This simulation's path, with an optional suffix such as `/step`.
+    fn path(&self, suffix: &str) -> String {
+        format!("/api/v2/simulations/{}{suffix}", self.id)
+    }
+
+    /// The current envelope.
+    fn envelope(&self) -> Envelope {
+        let path = self.path("");
+        match self.client.get(&path) {
+            Ok(response) => {
+                assert_eq!(
+                    response.status,
+                    200,
+                    "reading {path} must answer 200, got {} with {}",
+                    response.status,
+                    response.text()
+                );
+                match response.json(&path) {
+                    Ok(envelope) => envelope,
+                    Err(error) => panic!("{error}"),
+                }
+            }
+            Err(error) => panic!("{error}"),
+        }
+    }
+
+    /// Advances, optionally under an `expected_step` precondition.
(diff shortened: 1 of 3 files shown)


examples/integration/tests/v2_lifecycle.rs
```diff
@@ -0,0 +1,495 @@
...
+struct Created {
+    id: String,
+    state: String,
+    cursor: Cursor,
+    parameters: Parameters,
+}
+
+/// The replay inputs a client has to record to reproduce a run.
+#[derive(Debug, Deserialize)]
+struct Parameters {
```
joaquinbejar: 🟠 Warning This type models only a small subset of `SimulationParametersResponse`, and Serde ignores every response field not declared here. The test therefore stays green if timezone, calendar, expiration time, initial price, volatility, rates, method, time frame, strike interval, strike ladder, or any spread input disappears or is corrupted, despite claiming to cover every replay input. Mirror and assert the complete replay input object with exact requested and defaulted values.

examples/integration/tests/v2_lifecycle.rs
```diff
@@ -0,0 +1,495 @@
...
+    );
+    assert!(
+        !parameters.tzdb_version.is_empty(),
+        "the tzdb release the expirations were resolved against must come back"
+    );
+    assert_eq!(parameters.chain_size, Some(2));
+
+    // Normalisation is part of the replay contract: the rules come back
+    // sorted by rule id, whatever order they were sent in.
+    let ids: Vec<&str> = parameters
```
joaquinbejar: 🟠 Warning `reference_request` sends exactly one schedule, so sorting this one element list is vacuous and cannot detect removal of normalization. Send at least two rules deliberately out of `rule_id` order and assert the complete normalized rules in sorted order, including `kind` and any weekday or month fields rather than only positive target counts.

examples/integration/tests/v2_lifecycle.rs
```diff
@@ -0,0 +1,495 @@
...
+    let Some((live, _)) = Live::create(&client, 3) else {
+        return;
+    };
+
+    let before = live.envelope();
+    assert_eq!(before.cursor.current_step, 0);
+
+    for _ in 0..2 {
+        let path = live.path("/snapshot");
+        match client.get(&path) {
```
joaquinbejar, line 312: 🟠 Warning Both snapshot bodies are discarded, so stable metadata cursor and version do not prove that peeking is repeatable. The service can return different prices or contracts on the two GETs without mutating metadata and this test still passes. Deserialize the deterministic snapshot fields from both responses and compare them, then verify the next successful step serves the snapshot that was peeked.

examples/integration/tests/v2_lifecycle.rs
```diff
@@ -0,0 +1,495 @@
...
+    let winners = statuses.iter().filter(|status| **status == 200).count();
+    let losers = statuses.iter().filter(|status| **status == 409).count();
+    let after = live.envelope();
+    let moved = after.cursor.current_step - before.cursor.current_step;
+
+    // Two outcomes are correct. Either they raced and the store rejected the
+    // loser, or they serialised and both advanced; what must never happen is
+    // a 5xx, or a cursor that moved by something other than the number of
+    // successes.
+    assert!(
```
joaquinbejar, line 561: 🟠 Warning This accepts two 200 responses and a two step cursor move, and it also accepts two 409 responses with no movement because only `moved == winners` is required. Issue #101 requires exactly one winner, one conflict, and one consumed step. Synchronize the requests so they overlap and assert one 200, one 409, cursor plus one, and a single version increment.

examples/integration/tests/v2_lifecycle.rs
```diff
@@ -0,0 +1,495 @@
...
+        match self.client.request("POST", &path, None) {
+            Ok(response) => response,
+            Err(error) => panic!("advancing: {error}"),
+        }
+    }
+}
+
+impl Drop for Live {
+    fn drop(&mut self) {
+        if let Err(error) = self.client.delete(&self.path("")) {
```
joaquinbejar: 🟠 Warning `ServiceClient::delete` returns `Ok(Response)` for HTTP failures, so a 401, 405, or 500 is silently treated as successful cleanup and can leak a simulation on the shared deployment. Inspect the response status here, accepting 200 and the expected 404 after an explicit deletion, and report the status and body for any other result without panicking during unwinding.

joaquinbejar (commented):
Formal decision REQUEST_CHANGES. The incremental diff contains 499 changed lines and remains coherent, with CI green, but the v2 deployment suite does not cover the full replay input response, uses a vacuous one rule sorting check, discards the snapshots it claims are repeatable, accepts concurrency outcomes that never prove a single CAS conflict, and silently accepts failed cleanup responses. These contract gaps need correction before this stacked PR is ready.

examples/integration/tests/v2_lifecycle.rs
joaquinbejar: You are right, a struct with a subset of the fields is a test that stops noticing. The echo is compared as an object now, key by key: the sent values by value, the resolved ones by shape, and a missing key fails rather than being ignored.

examples/integration/tests/v2_lifecycle.rs
joaquinbejar: Fair, sorting one element proves nothing. The request sends two rules of different kinds deliberately out of rule_id order, and the whole normalised list is compared including the weekdays, since a rule that lost them would replay differently.

examples/integration/tests/v2_lifecycle.rs
joaquinbejar: Good catch, a service repricing on every GET without touching the cursor would have passed. Both bodies are compared now, and the advance that follows must serve the market the peek showed.

examples/integration/tests/v2_lifecycle.rs
joaquinbejar: Agreed. The two requests are released by a barrier so they overlap rather than happen to, and a contended attempt must be exactly one 200, one 409, one step consumed and one version. An attempt that serialised is retried, and five without contention fail, because an unexercised 409 path is not a passing test.

examples/integration/tests/v2_lifecycle.rs
joaquinbejar: Fixed, cleanup goes through report_cleanup, which accepts only the documented successes and 404 and reports anything else without panicking during unwinding.

joaquinbejar:
All five are fixed and verified against a live service, the race included: with the barrier it contends on the first attempt every time I ran it.

PR #123, opened 2026-08-30
Prove the seed reproducibility contract over HTTP

### LINKED ISSUE #102: Integration tests: the seed reproducibility contract, against a deployed service

joaquinbejar:
Closes #102

Base branch: `stack/8-101-v2-lifecycle`
Stacked on: #122
Blocks: #124

Stack: #109 → #110 → #111 → #112 → #99 → #104 → #100 → #101 → **#102** → #103 → #106 → #105 → #107

## What it asserts

Same parameters plus same seed yields an identical tape. Every test that proves that today runs in process, where the walk is one function call away and nothing can get between it and the answer. These prove it where IronCondor meets it, over HTTP, where a cache, a store round trip or a rebuilt walk could break it silently.

- two v2 simulations under the same explicit seed produce the same market at **every step**, not only at the last one
- two under different seeds disagree at some step, so the assertion above cannot pass vacuously
- a run nobody chose a seed for replays exactly from the seed the service **echoed**, which is the entire reason that echo exists
- a tape walked halfway, interrupted by another read, then resumed matches one walked straight through
- a v1 session reproduces its tape too, since v1 is frozen on rendered values and IronCondor consumes it

## What is compared, and why

Parsed values, not response text. The JSON rendering of a number is not the contract, the number is: a service that changed how it formats a float would fail a string comparison while serving the same market, and one that changed the market would pass a string comparison if it kept the formatting.

Identity, the cursor and the wall-clock timestamps are excluded because they differ between two runs by construction. That exclusion is not a guess: run against a live service, `session_info` and `timestamp` were the **only** differences between two same-seed v1 sessions, down to the last digit of every quote, greek and implied volatility.

## Cost against a shared deployment

Four steps over a two-strike chain per simulation, and every simulation is deleted when its test ends.

## Checklist

- [x] Reproduction asserted step by step
- [x] The echoed seed shown sufficient to replay a run nobody chose a seed for
- [x] A negative control with a different seed, in the same file
- [x] Patch version bumped to 0.2.10 with the compose image tag

### DIFF
--- a/examples/integration/tests/reproducibility.rs
+++ b/examples/integration/tests/reproducibility.rs
@@ -1,0 +1,390 @@
+//! The seed reproducibility contract, seen from outside the process.
+//!
+//! Same parameters plus same seed yields an identical snapshot tape. Every
+//! test that proves it today runs in process, where the walk is one function
+//! call away; this proves it where a consumer meets it, over HTTP against a
+//! running deployment, which is also where a cache, a store round trip or a
+//! rebuilt walk could break it without any unit test noticing.
+//!
+//! Comparisons are made on PARSED values rather than on the response text.
+//! The JSON rendering of a number is not the contract, the number is: a
+//! service that changed how it formats a float would fail a string comparison
+//! while serving exactly the same market, and a service that changed the
+//! market would pass one if it kept the formatting.
+//!
+//! The horizon is deliberately short, four steps over a two-strike chain,
+//! because these run against a shared deployment.
+//!
+//! Skipped entirely when `OCS_INTEGRATION_BASE_URL` is unset.
+
+use examples_integration::{Response, ServiceClient, reference_request, service};
+
+/// How many steps each tape carries.
+const HORIZON: usize = 4;
+
+/// A simulation that deletes itself.
+struct Live {
+    client: ServiceClient,
+    id: String,
+    seed: u64,
+}
+
+impl Live {
+    /// Creates a v2 simulation, optionally with an explicit seed, and reports
+    /// the seed it ended up with.
+    fn create(client: &ServiceClient, seed: Option<u64>) -> Option<Self> {
+        let mut request = reference_request("SPX");
+        if let Some(object) = request.as_object_mut() {
+            object.insert("steps".to_string(), serde_json::json!(HORIZON));
+            object.insert("chain_size".to_string(), serde_json::json!(2));
+            match seed {
+                Some(seed) => object.insert("seed".to_string(), serde_json::json!(seed)),
+                None => object.remove("seed"),
+            };
+        }
+
+        let response = match client.post("/api/v2/simulations", &request) {
+            Ok(response) => response,
+            Err(error) => panic!("{error}"),
+        };
+        if response.status == 404 {
+            println!("SKIP: this deployment has no v2 API");
+            return None;
+        }
+        assert_eq!(
+            response.status,
+            201,
+            "creating a simulation must answer 201, got {} with {}",
+            response.status,
+            response.text()
+        );
+
+        let body: serde_json::Value = match response.json("/api/v2/simulations") {
+            Ok(body) => body,
+            Err(error) => panic!("{error}"),
+        };
+        let id = match body.get("id").and_then(serde_json::Value::as_str) {
+            Some(id) => id.to_string(),
+            None => panic!("a created simulation must carry an id: {body}"),
+        };
+        let seed = match body
+            .get("parameters")
+            .and_then(|parameters| parameters.get("seed"))
+            .and_then(serde_json::Value::as_u64)
+        {
+            Some(seed) => seed,
+            None => panic!("a created simulation must echo its effective seed: {body}"),
+        };
+
+        Some(Self {
+            client: client.clone(),
+            id,
+            seed,
+        })
+    }
+
+    /// Serves the snapshot at the cursor and advances.
+    fn step(&self) -> Response {
+        let path = format!("/api/v2/simulations/{}/step", self.id);
+        match self.client.request("POST", &path, None) {
+            Ok(response) => response,
+            Err(error) => panic!("advancing: {error}"),
+        }
+    }
+}
+
+impl Drop for Live {
+    fn drop(&mut self) {
+        let path = format!("/api/v2/simulations/{}", self.id);
+        if let Err(error) = self.client.delete(&path) {
+            println!("WARNING: could not delete simulation {}: {error}", self.id);
+        }
+    }
+}
+
+/// The market a snapshot describes, with everything that is not the market
+/// stripped out: identity, the cursor, and the wall-clock timestamps differ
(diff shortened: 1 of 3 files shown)


examples/integration/tests/reproducibility.rs
```diff
@@ -0,0 +1,390 @@
...
+    client: ServiceClient,
+    id: String,
+    seed: u64,
+}
+
+impl Live {
+    /// Creates a v2 simulation, optionally with an explicit seed, and reports
+    /// the seed it ended up with.
+    fn create(client: &ServiceClient, seed: Option<u64>) -> Option<Self> {
+        let mut request = reference_request("SPX");
```
joaquinbejar, line 57: 🟠 Warning `reference_request` omits `start_at`, so every supposedly identical pair resolves a separate wall clock `effective_start`. Crossing a second boundary can fail a same seed comparison, while the different seed control can pass because `simulated_at` differs even if the RNG seed is ignored. Pin one explicit start for controlled pairs, and rebuild the omitted seed replay from the original response's complete effective parameters, including its echoed `effective_start`, rather than reusing only the seed.

examples/integration/tests/reproducibility.rs
```diff
@@ -0,0 +1,390 @@
...
+    }
+}
+
+/// The market a snapshot describes, with everything that is not the market
+/// stripped out: identity, the cursor, and the wall-clock timestamps differ
+/// between two runs by construction and say nothing about the walk.
+fn market(snapshot: &serde_json::Value) -> serde_json::Value {
+    let mut market = serde_json::Map::new();
+    for key in ["underlying", "chains", "simulated_at"] {
+        if let Some(value) = snapshot.get(key) {
```
joaquinbejar: 🟠 Warning Missing market fields are silently omitted here. If `/step` stops returning `chains`, the same seed runs still compare equal and the different seed control can still differ through `underlying`, so the test goes green without checking any strikes or quotes required by issue #102. Require all market keys through a typed snapshot or fail explicitly when a key, chain, contract, or rendered quote is absent.

examples/integration/tests/reproducibility.rs
```diff
@@ -0,0 +1,390 @@
...
+            "time_frame": "Day",
+            "chain_size": 2,
+            "strike_interval": 5.0,
+            "seed": seed
+        })
+    };
+
+    let create = |seed: u64| -> Option<String> {
+        match client.post("/api/v1/chain", &request(seed)) {
+            Ok(response) if response.status == 404 => {
```
joaquinbejar: 🟠 Warning This helper converts v1 contract failures into a passing skip. A configured deployment returning 404 yields `None`, and a 201 body without `id` also yields `None` at the end of the helper, causing the outer `let ... else` to return successfully. v1 is the guaranteed frozen route, so require 201 and make a missing id fail explicitly; only an unset integration base URL should skip this test.

examples/integration/tests/reproducibility.rs
```diff
@@ -0,0 +1,390 @@
...
+            Ok(response) => response,
+            Err(error) => panic!("advancing: {error}"),
+        }
+    }
+}
+
+impl Drop for Live {
+    fn drop(&mut self) {
+        let path = format!("/api/v2/simulations/{}", self.id);
+        if let Err(error) = self.client.delete(&path) {
```
joaquinbejar: 🟠 Warning `delete` returns `Ok(Response)` for HTTP failures, so this Drop silently accepts 401 or 500 and leaks the v2 simulation. The v1 sessions have the same guarantee gap because deletion happens only after all assertions at line 387 and its result is discarded. Install a guard as soon as each id is decoded and inspect the cleanup response, accepting the documented success plus 404 already gone without panicking during unwinding.

joaquinbejar (commented):
Formal decision REQUEST_CHANGES. The incremental diff contains 394 changed lines and is cohesive, with CI green, but the paired runs do not share an explicit effective start, the market projection silently accepts missing chains and quotes, v1 contract failures become passing skips, and cleanup can leak both v1 and v2 sessions. The current tests therefore do not yet isolate or safely verify the seed reproducibility contract.

examples/integration/tests/reproducibility.rs
joaquinbejar: You are right, and the control was the worse half: it could have passed on timestamps alone even if the seed were ignored. Controlled pairs share a pinned start_at now, and the replay of an unseeded run is rebuilt from the parameters the service resolved, its effective_start included, rather than from the seed alone.

examples/integration/tests/reproducibility.rs
joaquinbejar: Agreed, comparing equal on nothing is the failure mode that matters here. Every market key is required now, and the chains must quote an expiration, a strike and a side, so an empty or missing one fails instead of comparing equal.

examples/integration/tests/reproducibility.rs
joaquinbejar: Fair, v1 is the frozen route so a 404 there is a regression, not an older build. Both that and a 201 without an id fail now, and only an unset base URL skips.

examples/integration/tests/reproducibility.rs
joaquinbejar: Both gaps fixed. The v2 drop goes through report_cleanup, and the v1 sessions are guarded from the moment their id exists rather than deleted after the assertions, so a failing comparison no longer leaves them behind.

joaquinbejar:
All four are fixed and green against a live service. The wall-clock start was the one I would not have found: it made the negative control weaker than it looked.

PR #124, opened 2026-08-30
Cover the export matrix against a deployment

### LINKED ISSUE #103: Integration tests: the export matrix, every dataset in every format the deployment offers

joaquinbejar:
Closes #103

Base branch: `stack/9-102-reproducibility`
Stacked on: #123
Blocks: #125

Stack: #109 → #110 → #111 → #112 → #99 → #104 → #100 → #101 → #102 → **#103** → #106 → #105 → #107

## What it asserts

Six flows over a three-step, two-strike tape, which is deliberately small because an export is the most expensive thing the service does and the deployment is shared.

- every dataset in every offered format carries the same rows, compared as **parsed values** rather than text, since the rendering of a number is not the contract
- the same export twice is byte-identical, which is what lets a consumer cache or checksum one
- a range selects the same rows in every format, and an inverted range is a typed `400` **in every format**, never an empty document a consumer would read as "no data"
- each greek level header extends the previous one, so a consumer written against `none` keeps its columns in place when a richer level is asked for
- a streamed export reaches the last step of the tape with every column present, which is the failure a socket introduces and an in-process test cannot see
- the content type and suggested filename match the format asked for

## Honest against an older build

Formats are **probed**, not assumed: `json` and `csv` are required of every build, while `arrow` needs a feature that is off by default and `packed` postdates some deployments. A 400 for those is reported as a skip. The suite therefore passes against a deployment offering only json and csv, and covers more when more is deployed.

## Verified against a running service, twice

Once with the default features and once with `arrow-export` on, so the arrow branch was actually exercised rather than permanently skipped. That is how the expected media type got corrected: arrow is served as `application/vnd.apache.arrow.stream`, not as plain octets, and only the packed format falls back to `application/octet-stream`.

The packed check reads the footer, which is exactly what a consumer must do: the sentinel and the declared row count are what make a truncated download detectable, and the count is asserted against the json row count.

## Checklist

- [x] Row counts and values agree across every advertised format, for all three datasets
- [x] Repeat exports byte-identical
- [x] Range and greek-level behaviour asserted per format, rejections included
- [x] Passes against a deployment offering only json and csv
- [x] Patch version bumped to 0.2.11 with the compose image tag

### DIFF
--- a/examples/integration/tests/exports.rs
+++ b/examples/integration/tests/exports.rs
@@ -1,0 +1,515 @@
+//! The export matrix: every dataset in every format the deployment offers.
+//!
+//! The export is what a backtester actually consumes and it has the widest
+//! surface in the service: three datasets, up to four encodings, a step range
+//! and three greek levels. It also STREAMS, so its failure modes are the ones
+//! that only appear over a socket, truncation first among them.
+//!
+//! Formats are probed rather than assumed. A deployment built without the
+//! `arrow-export` feature refuses `arrow`, and an older one may not know
+//! `packed` at all; both are facts to report, not failures. What the suite
+//! guarantees is that everything the deployment DOES offer agrees with
+//! everything else it offers.
+//!
+//! Exports are kept small, three steps over a two-strike chain, because these
+//! run against a shared deployment and an export is the most expensive thing
+//! it does.
+//!
+//! Skipped entirely when `OCS_INTEGRATION_BASE_URL` is unset.
+
+use examples_integration::{Response, ServiceClient, reference_request, service};
+
+/// Steps in every exported tape.
+const STEPS: usize = 3;
+
+/// The datasets the export offers.
+const DATASETS: [&str; 3] = ["underlying", "volatility", "option_chains"];
+
+/// A simulation walked to the end, so its tape is complete, that deletes
+/// itself afterwards.
+struct Exported {
+    client: ServiceClient,
+    id: String,
+}
+
+impl Exported {
+    /// Creates a simulation and walks it to exhaustion.
+    fn create(client: &ServiceClient) -> Option<Self> {
+        let mut request = reference_request("SPX");
+        if let Some(object) = request.as_object_mut() {
+            object.insert("steps".to_string(), serde_json::json!(STEPS));
+            object.insert("chain_size".to_string(), serde_json::json!(2));
+        }
+
+        let response = match client.post("/api/v2/simulations", &request) {
+            Ok(response) => response,
+            Err(error) => panic!("{error}"),
+        };
+        if response.status == 404 {
+            println!("SKIP: this deployment has no v2 API");
+            return None;
+        }
+        assert_eq!(
+            response.status,
+            201,
+            "creating a simulation must answer 201, got {}",
+            response.text()
+        );
+
+        let body: serde_json::Value = match response.json("/api/v2/simulations") {
+            Ok(body) => body,
+            Err(error) => panic!("{error}"),
+        };
+        let id = match body.get("id").and_then(serde_json::Value::as_str) {
+            Some(id) => id.to_string(),
+            None => panic!("a created simulation must carry an id: {body}"),
+        };
+
+        let exported = Self {
+            client: client.clone(),
+            id,
+        };
+        for step in 0..STEPS {
+            let path = format!("/api/v2/simulations/{}/step", exported.id);
+            match client.request("POST", &path, None) {
+                Ok(response) => assert_eq!(
+                    response.status,
+                    200,
+                    "step {step} must serve before the tape can be exported: {}",
+                    response.text()
+                ),
+                Err(error) => panic!("{error}"),
+            }
+        }
+        Some(exported)
+    }
+
+    /// One export, with whatever query the case needs.
+    fn export(&self, query: &str) -> Response {
+        let path = format!("/api/v2/simulations/{}/export?{query}", self.id);
+        match self.client.get(&path) {
+            Ok(response) => response,
+            Err(error) => panic!("exporting {query}: {error}"),
+        }
+    }
+}
+
+impl Drop for Exported {
+    fn drop(&mut self) {
+        let path = format!("/api/v2/simulations/{}", self.id);
+        if let Err(error) = self.client.delete(&path) {
+            println!("WARNING: could not delete simulation {}: {error}", self.id);
+        }
+    }
+}
+
+/// Which formats this deployment actually serves, probed once.
+///
+/// `json` and `csv` are the baseline every build has; `arrow` needs a feature
+/// that is off by default and `packed` postdates some deployments, so a 400 or
+/// a 404 for those means "not offered here" and is reported rather than failed.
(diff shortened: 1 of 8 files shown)


examples/integration/tests/exports.rs
```diff
@@ -0,0 +1,515 @@
...
+/// `json` and `csv` are the baseline every build has; `arrow` needs a feature
+/// that is off by default and `packed` postdates some deployments, so a 400 or
+/// a 404 for those means "not offered here" and is reported rather than failed.
+fn offered_formats(exported: &Exported) -> Vec<&'static str> {
+    let mut offered = Vec::new();
+    for format in ["json", "csv", "arrow", "packed"] {
+        let response = exported.export(&format!("dataset=underlying&format={format}"));
+        match response.status {
+            200 => offered.push(format),
+            400 | 404 | 415 => println!(
```
joaquinbejar: 🟠 **Warning**: Capability discovery never consults the deployment OpenAPI document and treats every 400, 404, or 415 as an unsupported format. An advertised Arrow or packed implementation can regress and the matrix will silently skip it. Derive the required formats from OpenAPI first, then fail when an advertised format does not return its documented response.

examples/integration/tests/exports.rs
```diff
@@ -0,0 +1,515 @@
...
+            "{dataset} exports {} rows as csv and {} as json",
+            data.len(),
+            json_rows.len()
+        );
+
+        // Compared as parsed values rather than as text: the rendering of a
+        // number is not the contract, the number is.
+        for (index, (csv_row, json_row)) in data.iter().zip(&json_rows).enumerate() {
+            for (column, value) in header.iter().zip(csv_row) {
+                let Some(from_json) = json_row.get(column) else {
```
joaquinbejar: 🟠 **Warning**: Missing JSON fields are silently skipped, while both `zip` calls ignore short or trailing rows and columns. A format can drop or append data and this parity check still passes. Parse RFC4180 correctly, assert exact schema and key coverage plus row widths, then compare every canonical value.

examples/integration/tests/exports.rs
```diff
@@ -0,0 +1,515 @@
...
+            let declared = packed_rows(&packed.body, dataset);
+            assert_eq!(
+                declared as usize,
+                json_rows.len(),
+                "{dataset} declares {declared} rows in its packed footer and exports {} as json",
+                json_rows.len()
+            );
+        }
+
+        if formats.contains(&"arrow") {
```
joaquinbejar: 🟠 **Warning**: The binary encodings are not validated for parity. Packed checks only its footer row count and Arrow only nonempty bytes, so corrupted schemas, ordering, values, or unaligned packed blocks pass. Decode each offered format and compare every row and value with CSV, including the required eight byte packed alignment.

examples/integration/tests/exports.rs
```diff
@@ -0,0 +1,515 @@
...
+            packed_rows(&packed.body, "a ranged underlying export") as usize,
+            expected,
+            "the same range must select the same rows as packed and as json"
+        );
+    }
+
+    // And a bad range is a rejection in EVERY format, never an empty document
+    // that a consumer would read as "no data".
+    for format in &formats {
+        let response = exported.export(&format!(
```
joaquinbejar: 🟠 **Warning**: Range success omits Arrow, and rejection covers only inverted ranges while checking status and content type instead of the typed error body. This leaves the past-tape and per-format range contract untested. Exercise valid, inverted, and past-tape ranges for every advertised format, decode the exact selected rows, and assert the documented error and field.

joaquinbejar (commented):
Formal decision REQUEST_CHANGES. CI is green and the stack ancestry is clean, but format discovery can hide advertised regressions, schema mismatches are ignored, binary parity is not decoded, and range coverage is incomplete. The incremental diff has 519 changed lines, so split discovery and decoder infrastructure from the behavioral scenarios before re-review.

examples/integration/tests/exports.rs
joaquinbejar: You are right, probing and shrugging means a regression looks like an absence. The formats come from the deployment OpenAPI description now, and an advertised one must either serve or refuse with the typed 400 that says this build lacks the feature; any other refusal fails.

examples/integration/tests/exports.rs
joaquinbejar: Fixed properly: the csv is parsed as RFC 4180 with quoting, every row must be as wide as its header, the json keys must match the header exactly rather than being skipped when absent, and every value is compared canonically.

examples/integration/tests/exports.rs
joaquinbejar: Agreed, a footer count and a non-empty body are not parity. There is an independent packed decoder now, ported rather than borrowed from the writer so the bytes are read the way a consumer must, and it compares every row and value with the csv and reports any payload that does not start 8-byte aligned. Arrow is decoded with the arrow crate, which is already a workspace dependency, behind a feature the make target and the CI job turn on; I verified both against a build with arrow-export on.

examples/integration/tests/exports.rs
joaquinbejar: Right, status and content type are not the contract. Both wrong ranges are exercised in every offered format and asserted on the typed body and its field, from_step for an inverted one and to_step for one past the tape, and the valid range is decoded in every format and checked to select exactly the steps it names.

joaquinbejar:
All four are fixed, and the third one changed the shape of the PR: there is a real packed decoder now and the binary encodings are compared value by value rather than sniffed.

PR #126, opened 2026-08-30
Cover the operational surface: metrics, OpenAPI, swagger and the probes

### LINKED ISSUE #105: Integration tests: the operational surface — metrics, OpenAPI, and the probes

joaquinbejar:
Closes #105

Base branch: `stack/11-106-market-coherence`
Stacked on: #125
Blocks: #127

Stack: #109 → #110 → #111 → #112 → #99 → #104 → #100 → #101 → #102 → #103 → #106 → **#105** → #107

## What it asserts

- **every path the deployment advertises actually answers.** The list comes from the deployment own OpenAPI document rather than from a hardcoded one, so the test tracks whatever contract the service publishes. An advertised operation that answers 405, a 5xx, or a bare 404 fails: a client generated from that document belongs to someone who never touched this repository. Fourteen operations answered against a live service
- the metrics exposition parses as Prometheus text, documents the counters an operator dashboard uses, and moves when the service serves traffic
- swagger and the favicon are served, with the right content types
- the probes answer, `/ready` names each dependency with an `up` or a `down`, and both are absent from the counters
- an unknown path is `404`, never `500`

Probe assertions skip with a message on a deployment that predates `/health` and `/ready`.

## A correction the live run forced

My first version asserted the probe exclusion by comparing `api_requests_total` before and after a burst of probes. It failed against a live service, and the service was right: this suite runs its tests **in parallel** against one deployment, so a global counter moves for reasons that have nothing to do with probes.

It now asserts on the label sets instead: after ten probe requests, no `endpoint="/health"` or `endpoint="/ready"` series exists in the exposition. That is exact, cannot race, and is a stronger statement than the total ever was.

## Checklist

- [x] Every advertised path reachable, driven from the published document
- [x] The metrics exposition parses and moves with traffic
- [x] Probe assertions run where deployed and skip loudly where not
- [x] Patch version bumped to 0.2.13 with the compose image tag

### DIFF
--- a/examples/integration/tests/operational.rs
+++ b/examples/integration/tests/operational.rs
@@ -1,0 +1,336 @@
+//! The operational surface: what an operator points a probe at, a scraper at,
+//! or a code generator at.
+//!
+//! None of it is the product, all of it is what makes the product deployable,
+//! and each piece has been wrong at least once. The OpenAPI document matters
+//! most of the three: a generated client is built from it, so a path it
+//! advertises that answers 404 is a broken client for someone who never
+//! touched this repository.
+//!
+//! Skipped entirely when `OCS_INTEGRATION_BASE_URL` is unset.
+
+use examples_integration::service;
+
+/// The metrics the service documents and an operator's dashboard depends on.
+const DOCUMENTED_METRICS: [&str; 4] = [
+    "api_requests_total",
+    "api_errors_total",
+    "active_sessions",
+    "api_request_duration_seconds",
+];
+
+/// `/metrics` is a Prometheus exposition, and it moves when the service serves
+/// traffic.
+#[test]
+fn test_metrics_expose_the_documented_counters_and_move() {
+    let Some(client) = service() else {
+        return;
+    };
+
+    let response = match client.get("/metrics") {
+        Ok(response) => response,
+        Err(error) => panic!("{error}"),
+    };
+    assert_eq!(
+        response.status,
+        200,
+        "/metrics must answer 200, got {} with {}",
+        response.status,
+        response.text()
+    );
+    let content_type = response.header("content-type").unwrap_or_default();
+    assert!(
+        content_type.contains("text/plain"),
+        "a Prometheus exposition is text/plain, got {content_type:?}"
+    );
+
+    let body = response.text();
+    for metric in DOCUMENTED_METRICS {
+        assert!(
+            body.contains(&format!("# HELP {metric}")),
+            "the exposition must document {metric}"
+        );
+    }
+
+    // Every non-comment line is `name value` or `name{labels} value`, which is
+    // the shape a scraper parses.
+    for line in body.lines().filter(|line| !line.starts_with('#')) {
+        if line.trim().is_empty() {
+            continue;
+        }
+        let value = line.rsplit(' ').next().unwrap_or_default();
+        assert!(
+            value.parse::<f64>().is_ok() || value == "NaN" || value == "+Inf" || value == "-Inf",
+            "a metrics line must end in a number, got {line:?}"
+        );
+    }
+
+    let before = counter(&body, "api_requests_total");
+
+    // Serve some traffic that is unambiguously an API request.
+    for _ in 0..3 {
+        match client.get("/api/v1/chain?sessionid=00000000-0000-4000-8000-000000000000") {
+            Ok(_) => {}
+            Err(error) => panic!("{error}"),
+        }
+    }
+
+    let after = match client.get("/metrics") {
+        Ok(response) => counter(&response.text(), "api_requests_total"),
+        Err(error) => panic!("{error}"),
+    };
+
+    match (before, after) {
+        (Some(before), Some(after)) => assert!(
+            after > before,
+            "serving three requests must move api_requests_total, it went {before} to {after}"
+        ),
+        _ => println!(
+            "INFO: api_requests_total carries labels this test does not sum; the exposition was \
+             checked but the movement was not"
+        ),
+    }
+}
+
+/// The total of one counter across its label sets, or `None` when it is not
+/// exposed as a plain sample.
+fn counter(body: &str, name: &str) -> Option<f64> {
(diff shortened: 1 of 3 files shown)


examples/integration/tests/operational.rs
```diff
@@ -0,0 +1,336 @@
...
+            Err(error) => panic!("{error}"),
+        }
+    }
+
+    let after = match client.get("/metrics") {
+        Ok(response) => counter(&response.text(), "api_requests_total"),
+        Err(error) => panic!("{error}"),
+    };
+
+    match (before, after) {
```
joaquinbejar: 🟠 **Warning**: This sums every label set, so parallel integration traffic can make `after > before` even if these three requests were not recorded. It also turns an absent sample into an informational pass. Read the exact endpoint, method, and status series exercised here, treat an absent initial sample as zero, and require the expected delta.

examples/integration/tests/operational.rs
```diff
@@ -0,0 +1,336 @@
...
+    let Some(client) = service() else {
+        return;
+    };
+
+    let document: serde_json::Value = match client.get("/api-docs/openapi.json") {
+        Ok(response) if response.status == 200 => match response.json("/api-docs/openapi.json") {
+            Ok(document) => document,
+            Err(error) => panic!("{error}"),
+        },
+        Ok(response) => {
```
joaquinbejar: 🟠 **Warning**: A missing or broken OpenAPI endpoint is treated as a successful skip, although this operational contract explicitly requires the document to parse. A configured deployment without the document must fail; otherwise the entire advertised route test disappears precisely when the contract regresses.

examples/integration/tests/operational.rs
```diff
@@ -0,0 +1,336 @@
...
+                response.status, 405,
+                "{method} {path} is advertised but not mounted for that verb"
+            );
+            assert!(
+                response.status < 500,
+                "{method} {path} is advertised and answers {}: {}",
+                response.status,
+                response.text()
+            );
+            if response.status == 404 {
```
joaquinbejar, line 189: 🟠 **Warning**: A nonempty body does not distinguish a mounted resource from an unmounted route because proxies and custom 404 handlers commonly return HTML bodies. Require the service JSON error content type and shape for valid domain 404s, and reject 404 for operations where it is not a documented result.

examples/integration/tests/operational.rs
```diff
@@ -0,0 +1,336 @@
...
+        200,
+        "/health must answer 200 whatever the dependencies are doing, got {}",
+        health.text()
+    );
+
+    let ready = match client.get("/ready") {
+        Ok(response) => response,
+        Err(error) => panic!("{error}"),
+    };
+    assert!(
```
joaquinbejar: 🟠 **Warning**: The status code is accepted independently of the readiness body. This passes with 503 while every dependency is up, 200 while one is down, or missing Redis and MongoDB entries. Derive the aggregate state, require 200 and ready if and only if all dependencies are up, require the configured dependency names, and enforce the documented safe reason vocabulary for down entries.

joaquinbejar (commented):
Formal decision REQUEST_CHANGES. CI is green and the 340 line incremental diff is cohesive, but metrics movement can be masked or skipped, a missing OpenAPI document is treated as success, proxy 404 pages can look mounted, and readiness status is not tied to dependency state. These gaps must be corrected before this layer is ready.

examples/integration/tests/operational.rs
joaquinbejar: You are right on both halves, and the parallel-traffic one is the sharper: this suite runs its tests concurrently against one deployment, so a sum could be moved by anything. It reads one exact series now, endpoint, method and status, treats an absent sample as zero and requires a delta of exactly three.

examples/integration/tests/operational.rs
joaquinbejar: Agreed, that skip deleted the whole advertised-route check exactly when the contract broke. A configured deployment that does not serve the document now fails.

examples/integration/tests/operational.rs
joaquinbejar: Fair, a proxy answers HTML happily so a non-empty body proves nothing. A 404 is now accepted only where the operation takes an id, where it is a documented result, and it must carry this service JSON error shape.

examples/integration/tests/operational.rs
joaquinbejar: Good catch, a probe an orchestrator cannot trust is worse than none. The aggregate must follow from the parts in both directions, redis and mongodb must both be probed, an up dependency must carry no reason, and a down one must carry one of the documented categories rather than a message that could leak a host.

joaquinbejar:
All four are fixed and green against a live service, with fourteen advertised operations still answering.

PR #127, opened 2026-08-30
Cover concurrent clients and the streaming export path

### LINKED ISSUE #107: Integration tests: behaviour under concurrent clients and a long horizon

joaquinbejar:
Closes #107

Base branch: `stack/12-105-operational-surface`
Stacked on: #126
Stack: #109 → #110 → #111 → #112 → #99 → #104 → #100 → #101 → #102 → #103 → #106 → #105 → **#107** (last)

## What it asserts

The service is meant to be read by several clients at once and to stream an export rather than materialise it. Neither property exists in a single-threaded test, so nothing else in the suite touches them.

- concurrent exports of one simulation are **byte-identical** and move no cursor
- interleaved advances of two simulations stay independent, step for step
- contended advances answer `200` or `409` and nothing else, with the cursor moving **exactly once per winner**
- a peek during another client walk is always a complete chain, never a half-built one
- **the export streams, and a client may walk away.** A raw socket reads the first bytes of a long export and drops mid-stream; the service must still serve afterwards, and the same export read to the end must be complete
- deleting a simulation under an in-flight export completes or fails cleanly, and never takes the service with it

Everything is bounded to three clients and six steps, since the deployment is shared.

## Numbers, so a regression is visible

Recorded against a live service and printed by the tests rather than asserted as thresholds, which would be flaky on shared hardware:

- first bytes of a long export delivered after **750 µs**, which is the streaming path doing its job rather than pricing everything first
- the whole export, six steps of a 25-strike chain with the full greek set, is **54 200 bytes read in 112 ms**

## Checklist

- [x] Exactly one winner per contended step, cursor arithmetic asserted
- [x] Concurrent exports byte-identical
- [x] The service serves normally after a client abandons a stream mid-flight
- [x] Patch version bumped to 0.2.14 with the compose image tag

### DIFF
--- a/examples/integration/tests/concurrency.rs
+++ b/examples/integration/tests/concurrency.rs
@@ -10,9 +10,9 @@
 //! Skipped entirely when `OCS_INTEGRATION_BASE_URL` is unset.
 
 use examples_integration::{ServiceClient, reference_request, service};
-use std::io::{Read, Write};
+use std::io::{BufReader, Read, Write};
 use std::net::TcpStream;
-use std::sync::mpsc;
+use std::sync::{Arc, Barrier, mpsc};
 use std::thread;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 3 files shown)


examples/integration/tests/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+    let before = live.cursor();
+    let query = live.path("/export?dataset=option_chains&format=csv");
+    let (sender, receiver) = mpsc::channel();
+
+    let mut handles = Vec::new();
+    for _ in 0..CLIENTS {
+        let client = client.clone();
+        let query = query.clone();
+        let sender = sender.clone();
+        handles.push(thread::spawn(move || {
```
joaquinbejar, line 329: 🟠 **Warning**: Starting several threads does not ensure their exports overlap; the requests may serialize completely and still return identical bytes. Release all clients through a barrier before the GETs so this test actually exercises simultaneous readers.

joaquinbejar (commented):
Formal decision REQUEST_CHANGES. CI is green and the 480 line incremental diff is cohesive, but the tests do not establish simultaneous readers, independent tapes, a single contention winner, or a successful overlapping peek, while the socket checks do not prove body streaming or complete output and the delete race is unsynchronized. These gaps must be corrected before this layer is ready.

examples/integration/tests/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+        Live::create(&client, STEPS, 2),
+        Live::create(&client, STEPS, 2),
+    ) else {
+        return;
+    };
+
+    // Interleave: left, right, left, right...
+    for round in 0..STEPS {
+        for (name, live) in [("left", &left), ("right", &right)] {
+            match client.request("POST", &live.path("/step"), None) {
```
joaquinbejar: 🟠 **Warning**: This discards both snapshot bodies and compares only status and cursor counts. With identical parameters and seed, cross-wired caches or one simulation serving the other tape still pass. Use distinct seeds, build standalone reference tapes, and compare every interleaved snapshot with the corresponding baseline.

examples/integration/tests/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+    assert_eq!(statuses.len(), CLIENTS);
+    assert!(
+        statuses
+            .iter()
+            .all(|status| *status == 200 || *status == 409),
+        "a contended advance answers 200 or 409, never anything else, got {statuses:?}"
+    );
+
+    let winners = statuses.iter().filter(|status| **status == 200).count();
+    assert!(
```
joaquinbejar: 🟠 **Warning**: The test promises exactly one winner but accepts any `winners >= 1`; without a start barrier the requests may serialize and all return 200. Coordinate their release, then require exactly one 200, `CLIENTS - 1` conflicts, and cursor and version movement of exactly one.

examples/integration/tests/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+    };
+    let Some(live) = Live::create(&client, STEPS, 2) else {
+        return;
+    };
+
+    let advancing = {
+        let client = client.clone();
+        let path = live.path("/step");
+        thread::spawn(move || {
+            for _ in 0..STEPS {
```
joaquinbejar: 🟠 **Warning**: Every advance response is ignored and the peeker is unsynchronized, so it may run only after completion, observe only 410 responses, and finish with zero successful peeks. Coordinate the phases, assert the advance outcomes, require at least one 200 peek while advancement is active, and validate the expected complete chain shape.

examples/integration/tests/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+            address
+        );
+        if let Err(error) = socket.write_all(request.as_bytes()) {
+            panic!("{error}");
+        }
+
+        // Read a little and then walk away, which is what a client that
+        // cancels a download does.
+        let mut buffer = [0_u8; 1024];
+        let read = match socket.read(&mut buffer) {
```
joaquinbejar: 🟠 **Warning**: One arbitrary socket read can contain only the HTTP status line and headers. `taken > 0` therefore does not prove that response body bytes arrived before full materialization. Parse through the header terminator, require status 200 and at least one actual body or chunk byte, then signal and drop the socket.

examples/integration/tests/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    // And the same export, read to the end this time, is complete.
+    let whole = match client.get(&path) {
+        Ok(response) => response,
+        Err(error) => panic!("{error}"),
+    };
+    assert_eq!(whole.status, 200);
+    let text = whole.text();
+    assert!(
```
joaquinbejar: 🟠 **Warning**: CRLF terminates every CSV row, so a stream truncated cleanly at any row boundary passes this completeness check. Parse the CSV and require the expected rows and columns for every step, including the final step, or compare it with a separately validated complete export.

examples/integration/tests/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+    };
+    let Some(live) = Live::create(&client, STEPS, 8) else {
+        return;
+    };
+    live.walk();
+
+    let path = live.path("/export?dataset=option_chains&format=csv&greeks=all");
+    let delete_path = live.path("");
+
+    let exporting = {
```
joaquinbejar, line 740: 🟠 **Warning**: There is no handshake proving the export started before DELETE. Deletion can win before the spawned GET, and the accepted export 404 then makes the test pass without any in-flight race. Signal after 200 headers or the first body chunk, require DELETE 200 for the known-live id, then assert the started stream completes or terminates with the documented clean failure.

examples/integration/tests/concurrency.rs
joaquinbejar: You are right, threads started in a loop can serialise completely. All clients are released through a barrier now, and the test also asserts the version did not move, not just the cursor.

examples/integration/tests/concurrency.rs
joaquinbejar: Agreed, with one seed for both, one tape serving both would have passed. They carry different seeds now, each is walked alone first to build a reference tape, and every interleaved snapshot is compared against its own baseline; the test also asserts the two baselines differ, or it would prove nothing.

examples/integration/tests/concurrency.rs
joaquinbejar: Fixed the same way as the pair: a barrier, then exactly one 200 and CLIENTS minus one conflicts, with the cursor and the version each moving by one. An attempt that serialised is retried and five without contention fail.

examples/integration/tests/concurrency.rs
joaquinbejar: All four gaps closed. The phases are coordinated by a barrier, every advance outcome is asserted, at least one peek must succeed while the walk is moving, and each peeked chain must carry its whole ladder with both sides on every strike. It reports the count, eighteen complete peeks over eighteen steps against a live service.

examples/integration/tests/concurrency.rs
joaquinbejar: Right, a single read can be all headers. The reader parses to the header terminator, requires a 200 and at least one actual body byte, and only then drops the socket; against a live service that is 397 body bytes after about thirteen milliseconds.

examples/integration/tests/concurrency.rs
joaquinbejar: Good catch, every row ends in CRLF so that check could not fail. The re-read export is parsed now: every step present, every row as wide as its header, and every step carrying the same number of rows, which is exactly what a cut at a row boundary loses.

examples/integration/tests/concurrency.rs
joaquinbejar: Agreed, without a handshake the delete could win before the export existed and the accepted 404 hid it. The exporting thread signals after the status line and the first body bytes, the delete only runs then, and it must answer 200 for an id that is known live.

joaquinbejar:
All seven are fixed and green against a live service, run twice for flakiness. The theme was the same in every one: the test described contention it never created, so the barriers and the handshake are the substance of this change.

Issue #130, opened 2026-08-31
Automate the persisted-volume MongoDB upgrade test


joaquinbejar:
From the review of #116, where the MongoDB 8.0 to 8.2 bump was flagged as having an unverified upgrade path for an existing deployment.

## Context

The Integration job starts every service container on an EMPTY volume, so it proves the new image works and says nothing about upgrading a deployment that already holds data. The 8.0 to 8.2 procedure has a real precondition: the data directory carries a feature compatibility version, the binary refuses a directory whose FCV is above it, and raising the FCV is the step that stops being reversible.

For #116 the path was verified by hand against the exact digest being pinned: 8.0.29 seeded with a document and FCV 8.0, container stopped, image swapped to 8.2.12, started, document intact, FCV still 8.0, then raised to 8.2 successfully. That is now documented in `Docker/docker-compose.yml` next to the image.

## What to do

Make that a test rather than a paragraph, so the next bump cannot skip it.

## Acceptance criteria

- [ ] A job or script starts the PREVIOUS pinned MongoDB image on a named volume, writes a document, and records the FCV
- [ ] It swaps to the new pinned image on the same volume and asserts the container becomes healthy, the document is still there, and the FCV is unchanged
- [ ] It then raises the FCV and asserts the result
- [ ] It runs on a bump of the MongoDB image rather than on every push, since it is slow
- [ ] The same shape is considered for ClickHouse, whose upgrades also carry data-directory expectations

PR #131, opened 2026-08-31
Record the decision to carry no HTTP client in the harness

### LINKED ISSUE #117: Decide whether the integration harness may depend on an HTTP client

joaquinbejar:
Closes #117

Base branch: `main`
Blocks: #132

Stack: **#117** → #119 → #130 (this is the first)

The three issues in this stack are **independent**: none consumes anything another produces. They are stacked because that is how the run was requested; each could equally be a PR off `main`, and they can be merged bottom up without any of them needing the others.

## The decision, and that it was mine to take

#117 is an owner decision by construction: approve `reqwest` for `examples/integration`, or confirm the hand-written client stays. I took the option that **preserves current behaviour**, which is to keep it, and implemented that fully rather than leaving the issue open. If you want `reqwest` instead, say so and reversing this is small; the issue records exactly what it would take.

## What this PR actually changes

Documentation, in the three places someone meets the consequence:

- the harness module docs now state it as a recorded decision rather than a stopgap, with what it costs: **no HTTPS**, since a TLS client is a dependency of its own, so a deployment behind TLS cannot be tested and an `https://` URL is refused at construction with a message saying why; no connection reuse; and a protocol reader that grows the first time a test needs redirects or keep-alive
- `.env.example` says `OCS_INTEGRATION_BASE_URL` must be a plain `http://` address, and why
- the crate docs say the same in the testing section, so it is visible from the README

It also records what reversing the decision takes, which is genuinely small: add `reqwest` to `[workspace.dependencies]`, depend on it from this `publish = false` member alone, and replace the transport half of `ServiceClient::request`. Everything above it is written against `Response`, not against the socket.

## Checklist

- [x] No code behaviour change; the refusal of an `https://` URL already existed and is now documented
- [x] Reproducibility, stored-session shape and REST DTOs untouched
- [x] `make pre-push` clean, zero warnings, and `cargo clippy --all-features -- -D warnings` clean
- [x] `README.md` regenerated from `lib.rs`
- [x] Patch version bumped to 0.2.15 with the compose image tag

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -315,5 +315,9 @@ CLICKHOUSE_PASSWORD=password
 # tree; the tests report the version they found and skip a feature that is not
 # deployed yet rather than failing it.
 #
+# Must be a plain `http://` address. The harness carries no TLS client, by a
+# decision recorded in issue #117, so an `https://` URL is refused with a
+# message saying so rather than failing later as a protocol error.
+#
 # Where your deployment lives is your configuration; do not commit it.
 # OCS_INTEGRATION_BASE_URL=http://localhost:7070
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "optionchain_simulator"
-version = "0.2.14"
+version = "0.2.15"
 edition = "2024"
 # The maximum `rust-version` across the RESOLVED graph, not the highest direct
 # dependency: clickhouse 0.15.1, nalgebra 0.35, statrs 0.19.1 and wide 1.7 all
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -127,7 +127,7 @@ services:
     # Published by .github/workflows/docker-publish.yml on every Cargo.toml
     # version bump, or by `make docker-push` from a workstation. Bump the
     # default tag in the same PR that bumps the crate version.
-    image: ghcr.io/joaquinbejar/optionchain-simulator:${OPTIONCHAIN_VERSION:-0.2.14}
+    image: ghcr.io/joaquinbejar/optionchain-simulator:${OPTIONCHAIN_VERSION:-0.2.15}
     deploy:
       replicas: 2
       restart_policy:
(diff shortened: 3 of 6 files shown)


joaquinbejar (commented):
The no client decision is documented consistently at the harness, environment and crate documentation surfaces, with the current HTTP only behavior and reversal path stated honestly. Dependency policy, stack metadata and the synchronized version tag remain consistent, and CI is green. Formal decision APPROVE.

joaquinbejar:
Thanks. Worth noting for whoever reads this later: the decision is reversible in about the size of this diff, and #117 records exactly what it would take.

PR #132, opened 2026-08-31
Document the v1 rejection shapes rather than align them with v2

### LINKED ISSUE #119: v1 rejections do not carry the field the error shape promises

joaquinbejar:
Closes #119

Base branch: `stack/1-117-http-client-decision`
Stacked on: #131
Blocks: #133

Stack: #117 → **#119** → #130

The three issues are independent; the diff is cumulative until #131 merges, so read it against the base branch rather than against `main`.

## The decision, and that it was mine to take

#119 offered two options: make v1 carry `field` and correct the `Invalid State` prefix, or record that v1 rejections are what they are and document the difference. I took the second, because it **preserves current behaviour**: `/api/v1/chain` is frozen on rendered values, IronCondor consumes it, and adding a key or rewording a message changes what an existing client parses. If you would rather change the wire, this is the PR to say so on, and the work is small.

## What a live service actually answers, now documented

| Case | v1 | v2 |
|---|---|---|
| id that is not a UUID | `{error}`, prefixed `Invalid State`, **no `field` key** | `{error, field}` with `field` = `id` |
| id missing entirely | `{error, field}` with `field` **empty** | same |
| parameter out of range | `{error, field}` naming the field | same |

- every v1 endpoint's 400 now states which of those shapes it can answer. **DELETE documented no 400 at all** while answering one; it does now
- the crate docs carry the table above, so a client author meets the difference before their parser does
- `ValidationErrorResponse` says plainly that not every v1 400 uses it, and why

## Tests

Unchanged, deliberately: the #104 integration test already asserts exactly these shapes against a live service. Its comment now says this is the decided contract rather than a snapshot of an accident waiting on this issue. I re-read the published OpenAPI document from a running build to confirm every description landed.

## Checklist

- [x] No behaviour change; documentation and OpenAPI only
- [x] Reproducibility, stored-session shape and DTO shapes untouched
- [x] `make pre-push` clean, zero warnings; `README.md` regenerated
- [x] Patch version bumped to 0.2.16 with the compose image tag

### DIFF
--- a/examples/integration/tests/rejections.rs
+++ b/examples/integration/tests/rejections.rs
@@ -396,12 +396,15 @@ fn test_the_v1_rejection_contract_is_a_400_that_explains_itself() {
         return;
     };
 
-    // What a live service answers TODAY, asserted exactly, so a change in
-    // either direction is visible rather than silently tolerated: the
-    // malformed case carries no `field` key at all and renders with an
-    // `Invalid State` prefix, and the missing one carries an empty field.
-    // Both are what issue #119 decides; when it is decided these expectations
-    // change with it, deliberately.
+    // What a live service answers, asserted exactly, so a change in either
+    // direction is visible rather than silently tolerated: the malformed case
+    // carries no `field` key at all and renders with an `Invalid State`
+    // prefix, and the missing one carries an empty field.
+    //
+    // Issue #119 DECIDED this rather than left it open: v1 is frozen on
+    // rendered values, so the shapes stay and are documented in the OpenAPI
+    // document and the crate docs instead. These assertions are therefore the
+    // contract now, not a snapshot of an accident.
     for (case, path, expects_field, prefix) in [
         (
             "a malformed session id",
--- a/src/api/rest/handlers.rs
+++ b/src/api/rest/handlers.rs
@@ -289,7 +289,7 @@ fn json_body(bytes: Vec<u8>) -> HttpResponse {
     ),
     responses(
         (status = 201, description = "Session created successfully", body = SessionResponse),
-        (status = 400, description = "Validation failed: a parameter was non-finite, out of range (e.g. negative price/volatility), or steps/chain_size exceeded the configured limits.", body = ValidationErrorResponse),
+        (status = 400, description = "Validation failed: a parameter was non-finite, out of range (e.g. negative price/volatility), or steps/chain_size exceeded the configured limits. The body is the typed `{error, field}` naming the offending field.", body = ValidationErrorResponse),
         (status = 409, description = "Session id already exists", body = ErrorResponse),
         (status = 500, description = "Internal server error")
     )
@@ -420,7 +420,7 @@ pub(crate) struct ChainQuery {
     ),
     responses(
         (status = 200, description = "Advanced one step; served snapshot returned", body = ChainResponse),
-        (status = 400, description = "Malformed session id, or an unknown `greeks` level. The unknown-level body is the typed `{error, field}` of ValidationErrorResponse with `field` = `greeks`; a malformed id carries `error` alone."),
+        (status = 400, description = "Malformed or missing `sessionid`, or an unknown `greeks` level. The unknown-level body is the typed `{error, field}` with `field` = `greeks`. A malformed `sessionid` carries `error` ALONE, with no `field` key, and its message is prefixed `Invalid State`; a missing `sessionid` carries `{error, field}` with `field` empty. Neither is a ValidationErrorResponse: v1 is frozen on rendered values, so these shapes differ from v2, which names `id`. See issue #119."),
         (status = 404, description = "Session not found"),
         (status = 409, description = "Concurrent modification: another request advanced or modified the session; retry"),
         (status = 410, description = "Simulation completed. No more steps available"),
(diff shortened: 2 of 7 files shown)


src/api/rest/handlers.rs
```diff
@@ -612,7 +612,7 @@ pub(crate) async fn get_current_step(
     ),
     responses(
         (status = 200, description = "Session replaced", body = SessionResponse),
-        (status = 400, description = "Validation failed: a parameter was non-finite, out of range, or exceeded the configured limits.", body = ValidationErrorResponse),
+        (status = 400, description = "Validation failed: a parameter was non-finite, out of range, or exceeded the configured limits, OR the `sessionid` could not be read. A parameter failure carries the typed `{error, field}`. A malformed `sessionid` carries `error` ALONE, with no `field` key, and its message is prefixed `Invalid State`; a missing `sessionid` carries `{error, field}` with `field` empty. Neither is a ValidationErrorResponse: v1 is frozen on rendered values, so these shapes differ from v2, which names `id`. See issue #119.", body = ValidationErrorResponse),
```
joaquinbejar, line 615: 🟠 **Warning**: This publishes only `ValidationErrorResponse`, although a malformed `sessionid` returns `ErrorResponse`; PATCH repeats the mismatch at line 729, while DELETE has the inverse problem at line 841. Generated clients consume the schema, not this prose. Model both 400 bodies as a union or omit the single body schema consistently.

src/api/rest/handlers.rs
```diff
@@ -420,7 +420,7 @@ pub(crate) struct ChainQuery {
     ),
     responses(
         (status = 200, description = "Advanced one step; served snapshot returned", body = ChainResponse),
-        (status = 400, description = "Malformed session id, or an unknown `greeks` level. The unknown-level body is the typed `{error, field}` of ValidationErrorResponse with `field` = `greeks`; a malformed id carries `error` alone."),
+        (status = 400, description = "Malformed or missing `sessionid`, or an unknown `greeks` level. The unknown-level body is the typed `{error, field}` with `field` = `greeks`. A malformed `sessionid` carries `error` ALONE, with no `field` key, and its message is prefixed `Invalid State`; a missing `sessionid` carries `{error, field}` with `field` empty. Neither is a ValidationErrorResponse: v1 is frozen on rendered values, so these shapes differ from v2, which names `id`. See issue #119."),
```
joaquinbejar, line 423: 🟠 **Warning**: The statement that neither identifier failure is a `ValidationErrorResponse` contradicts the preceding sentence and the actual missing-query branch. `{error, field}` with an empty field is exactly that response type. Say only the malformed identifier uses `ErrorResponse`; the same wording needs correction at lines 546, 615, 729, and 841.

examples/integration/tests/rejections.rs
```diff
@@ -396,12 +396,15 @@ fn test_the_v1_rejection_contract_is_a_400_that_explains_itself() {
         return;
     };
 
-    // What a live service answers TODAY, asserted exactly, so a change in
-    // either direction is visible rather than silently tolerated: the
-    // malformed case carries no `field` key at all and renders with an
-    // `Invalid State` prefix, and the missing one carries an empty field.
-    // Both are what issue #119 decides; when it is decided these expectations
-    // change with it, deliberately.
+    // What a live service answers, asserted exactly, so a change in either
```
joaquinbejar, line 399: 🟠 **Warning**: This new claim says the rendered values are asserted exactly, but the assertion below only uses `starts_with`, so the rest of either message can change while the test stays green. Compare the full decided strings and reject unknown JSON keys if the complete v1 body shape is now frozen.

joaquinbejar (commented):
The runtime behavior remains frozen and the stack, version and CI are consistent, but OpenAPI still publishes the wrong body schemas, the descriptions contradict the actual missing identifier shape, and the regression test does not freeze the exact rendered message it claims. Formal decision REQUEST_CHANGES.

PR #133, opened 2026-08-31
Make the MongoDB persisted-volume upgrade a test

### LINKED ISSUE #130: Automate the persisted-volume MongoDB upgrade test

joaquinbejar:
Closes #130

Base branch: `stack/2-119-v1-rejection-shape`
Stacked on: #132
Stack: #117 → #119 → **#130** (last)

The three issues are independent; the diff is cumulative until the PRs below merge, so read it against the base branch rather than against `main`.

## Why

The Integration job starts every service container on an **empty volume**. That proves the new image runs and says nothing about upgrading a deployment that already holds data, which is exactly where MongoDB upgrades fail: the data directory carries a feature compatibility version, a binary refuses a directory whose FCV is above it, and raising the FCV is the step that stops being reversible.

For #116 I ran that procedure by hand. This makes it a test, so the next bump cannot skip it.

## What it does

`scripts/mongo_upgrade_test.sh`, in three steps that each fail loudly with the container log rather than hanging:

1. seed the **previous** image on a named volume with a document, record the FCV
2. stop it, start the **new** image on the same volume, and require the document to still be there **and the FCV to be unchanged** — this is the reversible point, both binaries serve that directory
3. raise the FCV and require it to take, with the document still intact

`make test-mongo-upgrade` runs it locally. The workflow runs it when the MongoDB pin in `Docker/docker-compose.yml` actually moves, which it detects by comparing the pin against the pull request's base, and on demand with explicit images. Not on every push: it pulls two database images and has nothing to say about a change that leaves the pin alone.

## A decision I took

ClickHouse is **not** covered, and the reason is written into the script rather than left implicit: it has no FCV handshake. A data directory written by a newer build makes the server refuse to start, which an operator sees immediately rather than discovering weeks later. If that ever changes, the script is the shape to copy.

## Verified

Run locally against the exact pins this repo carries:

```
PASS: 8.0.29 to 8.2.12 on a persisted volume, FCV 8.0 to 8.2
```

## Checklist

- [x] Runs the previous pinned image, swaps to the new one on the same volume, asserts the document, the FCV before the raise, and the raise
- [x] Runs on a MongoDB pin bump rather than every push, plus `workflow_dispatch`
- [x] ClickHouse considered, with the reason for leaving it out recorded
- [x] No library change; `make pre-push` clean, zero warnings
- [x] Patch version bumped to 0.2.17 with the compose image tag

### DIFF
--- a/.github/workflows/mongo-upgrade.yml
+++ b/.github/workflows/mongo-upgrade.yml
@@ -1,0 +1,87 @@
+# Proves an existing MongoDB deployment survives the image bump this repo pins.
+#
+# Deliberately NOT on every push: it pulls two database images and runs them in
+# sequence, which is slow, and it has nothing to say about a change that does
+# not move the MongoDB pin. It runs when that pin moves, and on demand.
+#
+# What it protects is the case the Integration job cannot see. That job starts
+# every service on an EMPTY volume, so it proves the new image runs; this one
+# starts the PREVIOUS image on a volume, writes a document, swaps the binary
+# underneath, and requires the document and the feature compatibility version
+# to survive before raising the FCV. See scripts/mongo_upgrade_test.sh.
+name: MongoDB upgrade (persisted volume)
+
+on:
+  workflow_dispatch:
+    inputs:
+      from_image:
+        description: "The image an existing deployment is running, e.g. mongo:8.0.29"
+        required: false
+        type: string
+      to_image:
+        description: "The image being upgraded to. Defaults to the pin in Docker/docker-compose.yml."
+        required: false
+        type: string
+      target_fcv:
+        description: "The feature compatibility version to raise to, e.g. 8.2"
+        required: false
+        type: string
+  pull_request:
+    paths:
+      - "Docker/docker-compose.yml"
+      - "scripts/mongo_upgrade_test.sh"
+      - ".github/workflows/mongo-upgrade.yml"
+
+jobs:
+  upgrade:
+    name: Upgrade on a persisted volume
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # The base commit is needed to tell a MongoDB bump from any other
+          # compose change.
+          fetch-depth: 0
+
+      - name: Decide whether the MongoDB pin moved
+        id: pin
+        run: |
+          set -euo pipefail
+          current=$(grep -oE 'image: mongo:[^ ]+' Docker/docker-compose.yml | head -1 | cut -d' ' -f2)
+          echo "current=$current" >> "$GITHUB_OUTPUT"
+
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "run=yes" >> "$GITHUB_OUTPUT"
+            echo "previous=${{ inputs.from_image }}" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          base="${{ github.event.pull_request.base.sha }}"
+          previous=$(git show "$base:Docker/docker-compose.yml" \
+            | grep -oE 'image: mongo:[^ ]+' | head -1 | cut -d' ' -f2 || true)
+          echo "previous=$previous" >> "$GITHUB_OUTPUT"
+
+          if [ "$previous" = "$current" ] || [ -z "$previous" ]; then
+            echo "The MongoDB pin did not move ($current); nothing to prove."
+            echo "run=no" >> "$GITHUB_OUTPUT"
+          else
+            echo "The MongoDB pin moved from $previous to $current."
+            echo "run=yes" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Upgrade a persisted volume
+        if: steps.pin.outputs.run == 'yes'
+        run: |
+          set -euo pipefail
+          from="${{ inputs.from_image || steps.pin.outputs.previous }}"
+          to="${{ inputs.to_image || steps.pin.outputs.current }}"
+          fcv="${{ inputs.target_fcv }}"
+
+          if [ -z "$fcv" ]; then
+            # The FCV is the major.minor of the image being upgraded to, which
+            # is what MongoDB accepts.
+            fcv=$(printf '%s' "$to" | sed -E 's/^mongo:([0-9]+\.[0-9]+).*/\1/')
+          fi
+
+          echo "Upgrading $from to $to, raising the FCV to $fcv"
+          ./scripts/mongo_upgrade_test.sh "$from" "$to" "$fcv"
--- a/.github/workflows/mongo-upgrade.yml
+++ b/.github/workflows/mongo-upgrade.yml
@@ -1,21 +1,27 @@
 # Proves an existing MongoDB deployment survives the image bump this repo pins.
 #
 # Deliberately NOT on every push: it pulls two database images and runs them in
-# sequence, which is slow, and it has nothing to say about a change that does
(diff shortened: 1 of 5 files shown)


.github/workflows/mongo-upgrade.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Decide whether the MongoDB pin moved
+        id: pin
+        run: |
+          set -euo pipefail
+          current=$(grep -oE 'image: mongo:[^ ]+' Docker/docker-compose.yml | head -1 | cut -d' ' -f2)
+          echo "current=$current" >> "$GITHUB_OUTPUT"
+
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "run=yes" >> "$GITHUB_OUTPUT"
+            echo "previous=${{ inputs.from_image }}" >> "$GITHUB_OUTPUT"
```
joaquinbejar: 🟠 **Warning**: GitHub expressions are inserted into the generated script before Bash parses it, so a workflow dispatch value containing command substitution or a newline is executed rather than kept as image data. The same pattern recurs at lines 76 through 78. Map these contexts and outputs through step level `env` entries, then use quoted shell variables.

scripts/mongo_upgrade_test.sh
```diff
@@ -0,0 +1,127 @@
...
+#
+# Usage:
+#   scripts/mongo_upgrade_test.sh [FROM_IMAGE] [TO_IMAGE] [TARGET_FCV]
+#
+# The defaults are the previous and current pins. Pass them explicitly when
+# bumping so the test runs against the two images the bump moves between.
+
+set -euo pipefail
+
+FROM_IMAGE="${1:-mongo:8.0.29}"
```
joaquinbejar: 🟠 **Warning**: This documented previous pin is a mutable tag, while the repository actually pinned `mongo:8.0.29@sha256:de267922bc1153d923f5c9dc429f21c11faf18299080c1ce04d6d6007097fb06`. The default Make target and an omitted dispatch input therefore do not prove the exact upgrade path claimed here. Use the digest or require the source image explicitly.

.github/workflows/mongo-upgrade.yml
```diff
@@ -0,0 +1,87 @@
...
+            exit 0
+          fi
+
+          base="${{ github.event.pull_request.base.sha }}"
+          previous=$(git show "$base:Docker/docker-compose.yml" \
+            | grep -oE 'image: mongo:[^ ]+' | head -1 | cut -d' ' -f2 || true)
+          echo "previous=$previous" >> "$GITHUB_OUTPUT"
+
+          if [ "$previous" = "$current" ] || [ -z "$previous" ]; then
```
joaquinbejar: 🟠 **Warning**: An empty base pin is treated as the same successful no op as an unchanged pin, so a lookup or parser regression silently removes this gate. Fail when either image cannot be discovered and skip only when two nonempty references are equal. The current green workflow took this skip path, so changes to the test itself also need to force one exact upgrade run.

scripts/mongo_upgrade_test.sh
```diff
@@ -0,0 +1,127 @@
...
+TO_VERSION=$(mongosh_eval 'db.version()')
+echo "   running $TO_VERSION"
+
+SURVIVED=$(mongosh_eval "db.getSiblingDB('$PROBE_DB').events.countDocuments()")
+[ "$SURVIVED" = "1" ] || fail "the document did not survive the swap, count was $SURVIVED"
+
+FCV_AFTER_SWAP=$(mongosh_eval 'db.adminCommand({getParameter: 1, featureCompatibilityVersion: 1}).featureCompatibilityVersion.version')
+[ "$FCV_AFTER_SWAP" = "$FCV_BEFORE" ] || \
+    fail "the swap moved the FCV from $FCV_BEFORE to $FCV_AFTER_SWAP; it must stay put, that is what makes this step reversible"
+echo "   document intact, FCV still $FCV_AFTER_SWAP, so the old image could still take over"
```
joaquinbejar: 🟠 **Warning**: An unchanged FCV after the new binary opens the volume does not prove the old binary can still open the now touched directory, yet this calls the point reversible. Before raising FCV, restart the exact old digest on the same volume and recheck readiness, data, and FCV, then return to the new image for the irreversible raise.

joaquinbejar (commented):
The workflow is well scoped and the stack and version are consistent, but dispatch inputs are interpolated into shell, the previous image default is not the pinned digest, discovery failures and this PR itself produce a green skip, and the claimed reversible point never restarts the old binary. The ordinary checks are green, while the MongoDB job did not execute the upgrade step. Formal decision REQUEST_CHANGES.

.github/workflows/mongo-upgrade.yml
joaquinbejar: You are right, that is script injection and I wrote it. Every context value reaches the shell through step level env now and is read as a quoted variable, and each image reference is checked against the shape a reference actually has before it goes near GITHUB_OUTPUT, which also closes the forged output line a newline would have opened.

scripts/mongo_upgrade_test.sh
joaquinbejar: Fair, a mutable tag there means the default run proves an upgrade from bytes nobody ships. The default is the digest the compose file carried before the bump, and the local run pulls exactly that.

.github/workflows/mongo-upgrade.yml
joaquinbejar: Both halves fixed, and the second was the sharper: this very PR took the skip path, so the workflow had never run its own test. Either pin missing is a failure now rather than a quiet success, and a change to the script or the workflow forces a run against the documented default pair.

scripts/mongo_upgrade_test.sh
joaquinbejar: Agreed, an unchanged FCV says nothing about whether the old binary can still open a directory the new one has opened. Step 3 now stops the new image, starts the old digest again on the touched volume, and requires it to become ready with the data and the FCV intact, and only then does the run return to the new image for the irreversible raise. Locally: rollback to 8.0.29 proven, then FCV 8.0 to 8.2.

joaquinbejar:
All four are fixed. The one I would not have caught alone is that this PR's own workflow run took the skip path, so a test about upgrades had never once upgraded anything in CI; it does now, because a change to the test forces a run.

Issue #139, opened 2026-08-31
Integration tests for a replicated deployment


joaquinbejar:
The integration suite covers every endpoint but assumes, in places, that one process answers. A replicated deployment is now the normal case, so the behaviours that only exist across instances need their own tests.

## What to cover

- **A tape is the same whichever instance serves it.** A simulation walked step by step through a balancer crosses instances; every snapshot must match one walked against a single instance under the same seed. The reproducibility file compares two simulations against each other, which cannot see a per-instance divergence that affects both.
- **No step is lost or served twice across instances.** Advance a simulation to exhaustion through the balancer and require the served cursors to be exactly `0..N-1`, with no repeats and no gaps: this is the cross-instance half of the compare-and-swap the Redis store implements in Lua.
- **A simulation created on one instance is immediately visible on another.** Create, then read in a loop until both instances have answered, all inside the retention window.
- **An export served by either instance is byte-identical.** Repeat the byte-identity check enough times to cross instances, which turns it into a statement about the deployment rather than about one process.
- **The probes reflect the instance answering them**, and a dependency down on one replica does not read as ready on another.
- **How many instances are behind the address**, reported once at the start of the run, so every other test's output can be read in that light. Counting distinct `api_requests_total` series across scrapes is enough.

## Not covered by this issue

Whether the deployment SHOULD share its caches or its admission cap: that is #135, #136 and #137. These tests describe what a client sees, so they stay true whichever way those go.

## Acceptance criteria

- [ ] Each behaviour above has a test that fails if the deployment stops honouring it
- [ ] The suite reports the instance count it detected rather than assuming one
- [ ] Everything stays skipped when `OCS_INTEGRATION_BASE_URL` is unset, and passes against a single-instance deployment too
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check` clean

Issue #137, opened 2026-08-31
Tape build coalescing is per process, so replicas duplicate the same build


joaquinbejar:
Found auditing the service for multi-instance operation. Related to the shared-cache issue, but a distinct mechanism.

## What happens

`SimulationManagerV2::builds` coalesces concurrent builds of the same factor tape: the first caller builds, the others wait on a `broadcast` channel and take the result. That is a `Mutex<HashMap<..>>` in one process.

With N replicas, N instances can be building the SAME tape at the same moment, each believing it is the only one. The coalescing that exists precisely to stop duplicate work stops working exactly when the work is most duplicated.

## What to do

A short-lived build lock in Redis, taken before the build and released after it, with waiters polling the shared cache rather than a local channel. It needs an expiry, or an instance killed mid-build blocks every other one until someone notices.

This depends on the tape living in Redis at all, so it should land after or with the shared-cache work rather than before it.

## Acceptance criteria

- [ ] One build of a given tape at a time across the deployment, not one per instance
- [ ] A lock that expires, so a killed builder does not wedge the others
- [ ] A waiter that times out builds it itself rather than failing the request
- [ ] A test with two managers over one Redis proving only one of them builds

Issue #136, opened 2026-08-31
The tape, snapshot and walk caches are per process, so replicas rebuild each other's work


joaquinbejar:
Found auditing the service for multi-instance operation.

## What happens

Three caches live in one process:

- the walk cache, `Simulator::simulation_cache`, one `RandomWalk` per simulation id
- the v2 tape cache, `SimulationManagerV2::tapes`
- the v2 snapshot cache

Behind a balancer with N replicas, consecutive steps of ONE simulation land on different instances, so each one rebuilds what the other just built. The hit rate collapses towards 1/N, and the work being repeated is the expensive kind: rebuilding a factor tape walks every step from the seed.

**Correctness is not at risk.** The tape is a pure function of the stored parameters and the seed, which is the whole point of the reproducibility contract, so whichever instance answers produces the same numbers. This is a cost problem, and on a long simulation it is a large one.

## What to do

Move the derived state to Redis, which is already the shared store and already speaks Lua for atomic writes. In rough order of value:

1. **The factor tape**, which is the expensive one to rebuild and the one a long simulation hits every step.
2. **The snapshot cache**, bounded by `OCS_MAX_CACHED_SNAPSHOT_CONTRACTS` as it is today, per deployment rather than per process.
3. The walk cache, which is largely subsumed by the tape.

A shared cache changes the eviction story: the bound stops being per process, and a stale entry now outlives the instance that made it, so the key has to carry everything that decides the values, the snapshot generation included.

## Acceptance criteria

- [ ] The factor tape is read from and written to Redis, keyed by simulation id and by whatever else decides its values, generation included
- [ ] The bound is per deployment, and the knob's documentation says so
- [ ] A cold instance serves a step already built by another without rebuilding, proven by a test rather than by inspection
- [ ] The values are identical either way: same seed, same tape, whichever instance answers
- [ ] Redis being unreachable degrades to the in-process behaviour rather than failing the request

Issue #135, opened 2026-08-31
The pricing admission cap is per process, so replicas multiply it


joaquinbejar:
Found auditing the service for multi-instance operation.

## What happens

`OCS_MAX_CONCURRENT_PRICING_JOBS` is enforced by a `LazyLock<Semaphore>` in `src/utils/admission.rs`, which lives in ONE process. An operator who sets it to 4 and runs two replicas gets **eight** concurrent pricing jobs, and does not find out until the host is saturated.

Pricing is the one genuinely expensive thing this service does: a snapshot can be `OCS_MAX_SNAPSHOT_CONTRACTS` contracts of Black-Scholes on `Decimal`, and the point of the semaphore is that a burst must not commit every core to work whose clients have gone. A cap that silently multiplies with the replica count is not that guarantee.

## Options

1. **Leave it per process and say so.** Document the knob as per-instance and let the operator divide by the replica count. Cheapest, and it makes total capacity a manual calculation that goes stale the moment someone scales.
2. **A shared lease in Redis.** The store is already there and every write in it goes through a Lua script, so a bounded token set with an expiry is a small addition. It costs a round trip on a path that is already the slow one, and it needs the expiry so a killed instance cannot hold a token forever.

Option 2 matches what the service already does for session state. Option 1 is honest but leaves the guarantee weaker than it looks.

## Acceptance criteria

- [ ] A decision recorded here
- [ ] If shared: a bounded lease in Redis with an expiry, acquired around the work `utils::admission` guards today, released on drop and on panic
- [ ] The knob documented as whichever it is, in `.env.example` and the crate docs
- [ ] A test that acquirers across two managers cannot exceed the cap together
- [ ] The behaviour when Redis is unreachable is deliberate and documented: refusing to price is a worse answer than pricing unbounded, so probably fall back to the local semaphore and warn

PR #140, opened 2026-08-31
Say that metrics belong to the instance that answers

### LINKED ISSUE #138: Scraping /metrics through the balancer sees one replica at a time
Found running the integration suite against a replicated deployment.

## What happens

Scraping `/metrics` repeatedly through the published port returns alternating numbers:

```
api_requests_total{endpoint="/metrics",method="GET",status="200"} 4
api_requests_total{endpoint="/metrics",method="GET",status="200"} 8
api_requests_total{endpoint="/metrics",method="GET",status="200"} 5
api_requests_total{endpoint="/metrics",method="GET",status="200"} 9
```

Two replicas, each counting its own traffic, answering the scrape in turn. Every counter this service publishes is per process, which is correct and normal; what is wrong is scraping a load-balanced address as if it were one target. Prometheus sees a counter that jumps up and down, reads each drop as a reset, and `rate()` becomes fiction. Three requests can move a counter by one, which is what the integration test now tolerates and reports.

## What to do

This is deployment configuration and documentation, not a code change to the metrics themselves:

- each replica must be its OWN scrape target, with an instance label, so the counters stay monotonic per series and aggregate correctly
- the compose file and the crate docs should say that plainly, because the current published-port arrangement invites exactly the mistake above
- if a single scrape endpoint is genuinely wanted, that is an aggregator in front, not something this service should pretend to do

## Acceptance criteria

- [ ] The crate docs and `.env.example` state that the counters are per process and must be scraped per replica
- [ ] `Docker/docker-compose.yml` shows how, or says why it cannot with a published port
- [ ] The `/metrics` handler documentation says which numbers are per process
- [ ] The integration suite keeps its current tolerant assertion and the line it prints when it detects more than one instance

joaquinbejar:
Closes #138

Base branch: `main`
Blocks: #141

Stack: **#138** → #139 → #136 → #135 → #137 (this is the first)

Only one hard dependency exists in this chain, the one #137 declares: the build lock needs the shared cache from #136. **#138, #139 and #135 are independent** of everything else and could each be a PR off `main`.

## Why

Running the suite against the two-replica deployment showed a counter moving by one for three requests. Scraping the published port repeatedly explains it:

```
api_requests_total{endpoint="/metrics",method="GET",status="200"} 4
api_requests_total{endpoint="/metrics",method="GET",status="200"} 8
api_requests_total{endpoint="/metrics",method="GET",status="200"} 5
api_requests_total{endpoint="/metrics",method="GET",status="200"} 9
```

Two replicas answering in turn, each counting its own traffic. Nothing about the metrics is broken; an exposition is per process by definition. What was missing is the instruction that follows: **scrape each replica as its own target and sum across the instance label**. Scraping the balanced address makes Prometheus read every apparent drop as a counter reset, and `rate()` over that is not so much wrong as meaningless.

## Where it now says so

The four places someone meets it: the `/metrics` handler and its OpenAPI description, the crate docs (so it reaches the README), `Docker/docker-compose.yml` right next to `replicas: 2`, and `.env.example`.

The same note names the other two per-process things worth knowing — the pricing admission cap (#135) and the tape and snapshot caches (#136) — and states plainly that **neither changes what a client is served**: session and simulation state lives in Redis and every write there is an atomic Lua script, so two instances cannot both commit the same step. What they cost is duplicated work, not correctness.

## Checklist

- [x] No behaviour change; documentation only
- [x] Reproducibility, stored-session shape and DTOs untouched
- [x] `make pre-push` clean, zero warnings; `README.md` regenerated
- [x] Patch version bumped to 0.2.19 with the compose image tag

### DIFF
--- a/src/api/rest/middleware.rs
+++ b/src/api/rest/middleware.rs
@@ -2,11 +2,24 @@ use crate::infrastructure::MetricsCollector;
 use actix_web::{HttpResponse, Responder, web};
 use std::sync::Arc;
 
+/// Serves this INSTANCE's metrics, in the Prometheus text format.
+///
+/// Every counter, gauge and histogram here belongs to the process that
+/// answers, which is normal and correct for a Prometheus exposition and is the
+/// thing to remember when the service runs replicated: the numbers are not a
+/// deployment total, they are one instance's share of it.
+///
+/// A deployment behind one published port therefore has to be scraped **per
+/// replica**, each as its own target with its own instance label. Scraping the
+/// balanced address instead returns whichever instance answered, so a counter
+/// appears to jump up and down between scrapes, Prometheus reads each drop as
+/// a counter reset, and every `rate()` over it is fiction. Summing across the
+/// instance label is what produces a deployment total.
 #[utoipa::path(
     get,
     path = "/metrics",
     responses(
-        (status = 200, description = "Prometheus metrics", content_type = "text/plain")
+        (status = 200, description = "This instance's Prometheus metrics. Counters are PER PROCESS: a replicated deployment must be scraped per replica, each as its own target with an instance label, and totalled by summing across that label. Scraping a load-balanced address returns one instance at a time, which reads as a counter reset.", content_type = "text/plain")
     )
 )]
 pub(crate) async fn metrics_endpoint(
--- a/src/api/rest/middleware.rs
+++ b/src/api/rest/middleware.rs
@@ -13,13 +13,28 @@ use std::sync::Arc;
 /// replica**, each as its own target with its own instance label. Scraping the
 /// balanced address instead returns whichever instance answered, so a counter
 /// appears to jump up and down between scrapes, Prometheus reads each drop as
-/// a counter reset, and every `rate()` over it is fiction. Summing across the
-/// instance label is what produces a deployment total.
+/// a counter reset, and every `rate()` over it is fiction.
+///
+/// How to aggregate depends on what the metric is, and only counters and
+/// histograms aggregate cleanly:
+///
+/// - **counters** (`api_requests_total`, `api_errors_total`,
+///   `session_creations_total`, the cache counters) and **histograms**
+///   (`api_request_duration_seconds`): take the rate per instance and sum, the
+///   usual `sum(rate(...))`. Summing raw counter values across restarts is not
+///   the same thing.
+/// - **gauges** (`active_sessions`, `simulation_cache_size`,
+///   `memory_usage_bytes`): each is that PROCESS's own view. `active_sessions`
+///   in particular is not a deployment total under any aggregation: a create
+///   served by one replica and a delete served by another leave the first
+///   replica's gauge untouched, and neither reaps nor restarts move it. Read
+///   them per instance; the authoritative live-session count is the store's,
+///   not a metric's.
 #[utoipa::path(
     get,
     path = "/metrics",
     responses(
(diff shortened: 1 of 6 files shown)


src/api/rest/middleware.rs
```diff
@@ -2,11 +2,24 @@ use crate::infrastructure::MetricsCollector;
...
+/// answers, which is normal and correct for a Prometheus exposition and is the
+/// thing to remember when the service runs replicated: the numbers are not a
+/// deployment total, they are one instance's share of it.
+///
+/// A deployment behind one published port therefore has to be scraped **per
+/// replica**, each as its own target with its own instance label. Scraping the
+/// balanced address instead returns whichever instance answered, so a counter
+/// appears to jump up and down between scrapes, Prometheus reads each drop as
+/// a counter reset, and every `rate()` over it is fiction. Summing across the
+/// instance label is what produces a deployment total.
```
joaquinbejar: 🟠 **Warning**: this aggregation rule does not hold for every metric named above. In particular, `active_sessions` is a process-local approximation: a create on replica A followed by a delete on B leaves A’s gauge unchanged, and TTL expiry or restarts are not reflected, so summing instances cannot produce the deployment’s live-session total. Scope this advice to counter rates and histograms, and document each gauge’s aggregation semantics separately; use a Redis-derived metric if a global active-session count is promised.

src/lib.rs
```diff
@@ -205,6 +205,37 @@
...
+//!
+//! Two replicas answering in turn. Prometheus reads every apparent drop as a
+//! counter reset, so `rate()` over such a series is not wrong so much as
+//! meaningless. A single scrape endpoint for a whole deployment is an
+//! aggregator's job, not something this service pretends to do.
+//!
+//! The same is true of anything else the process holds: the pricing admission
+//! semaphore bounds one instance (issue #135), and the tape and snapshot
+//! caches are per instance too (issue #136). None of that affects what a
+//! client is SERVED — the stores are shared and every write goes through an
```
joaquinbejar, line 242: 🟠 **Warning**: the v1 store does not send every write through Lua. Create, blind save, and delete use separate Redis commands; only the compare-and-swap commit that protects advancement is scripted. The consistency argument is still defensible, but state the actual guarantee as atomic CAS cursor commits rather than promising that every write is an atomic script, and update the matching compose and env comments.

joaquinbejar (commented):
REQUEST_CHANGES. The per-process model is documented clearly, but the aggregation guidance overpromises a deployment total and the Redis consistency wording promises atomic scripts for writes that use separate commands.

src/api/rest/middleware.rs
joaquinbejar: You are right, and active_sessions is the clearest case: a create on one replica and a delete on another leave the first gauge untouched, so no aggregation of it is a deployment total. The advice is now split by metric kind, counters and histograms as sum(rate(..)) and gauges as per instance, in all four places, with the note that the live session count is a question for the store rather than for a gauge.

src/lib.rs
joaquinbejar: Correct, only the compare-and-swap cursor commit is scripted; create, save and delete are single commands. The text now states the guarantee that actually holds, that two instances cannot both commit the same step, and the compose and env comments say the same.

joaquinbejar:
Both fixed. A PR whose whole content is documentation getting two claims wrong is worth the correction: the gauge one would have sent someone summing active_sessions across replicas.

PR #141, opened 2026-08-31
Cover what a client sees when replicas answer

### LINKED ISSUE #139: Integration tests for a replicated deployment

joaquinbejar:
Closes #139

Base branch: `stack/1-138-metrics-per-replica`
Stacked on: #140
Blocks: #142

Stack: #138 → **#139** → #136 → #135 → #137

The diff is cumulative until #140 merges, so read it against the base branch rather than against `main`.

## What it covers

`examples/integration/tests/replication.rs`, six flows, each written so a single-instance deployment passes them trivially — the client contract does not depend on how many processes answer:

- a simulation created through the balancer is readable through it, repeatedly. An instance serving from its own memory rather than the shared store would fail half these reads
- a walk serves every step **exactly once, in order**, and the simulation is completed for everyone afterwards. This is the cross-instance half of the compare-and-swap the Redis store implements in Lua
- two simulations under one seed agree step by step whichever instance serves each. The reproducibility file compares two simulations against each other, which cannot see a divergence affecting both
- an export is byte-identical across enough repeats to cross instances
- every `/ready` answer is consistent with the dependencies **of the instance that produced it**, so a replica whose Redis is gone cannot hide behind one whose Redis is fine
- the run reports how many instances it detected, so every other test's output can be read in that light

## A bug this exposed in the metrics test

The tolerant assertion from #134 was still wrong: it compared two scrapes that can come from **different instances**, so the second can legitimately be lower and the delta zero. Against two real replicas it failed roughly one run in two.

It now counts instances, sends traffic proportional to that count so every instance is served, and compares the **maximum** observed value, which belongs to one instance and only grows. Counting instances moved into the harness since both files want it: a counter only grows within a process, so N instances taking turns are N interleaved non-decreasing sequences, and the count is the smallest number of such sequences the scrapes split into.

Two processes over one Redis behind a round-robin proxy, plus a single instance for the trivial case:

- whole suite green **twice in a row** against two instances (13 test binaries)
- whole suite green against one instance
- the instance counter reports 2 and 1 correctly

## Checklist

- [x] Each behaviour has a test that fails if the deployment stops honouring it
- [x] The suite reports the instance count rather than assuming one
- [x] Everything still skips when `OCS_INTEGRATION_BASE_URL` is unset, and passes against a single-instance deployment
- [x] `make pre-push` clean, zero warnings, and `clippy --all-features -- -D warnings` clean
- [x] Patch version bumped to 0.2.20 with the compose image tag

### DIFF
--- a/examples/integration/src/lib.rs
+++ b/examples/integration/src/lib.rs
@@ -616,59 +616,70 @@ pub fn report_cleanup(client: &ServiceClient, path: &str, what: &str) {
     }
 }
 
-/// How many distinct processes answered `scrapes` scrapes of `/metrics`.
-///
-/// A counter is per process and only ever grows, so one instance produces a
-/// non-decreasing sequence of scrapes. N instances taking turns produce N such
-/// sequences INTERLEAVED, which reads as a series that keeps stepping
-/// backwards. The instance count is therefore the smallest number of
-/// non-decreasing sequences the observations can be split into, which is what
-/// the greedy pass below computes.
-///
-/// It is a lower bound: two instances that happen to hold the same value are
-/// indistinguishable from one. It never overcounts, which is what matters for
-/// a number this suite only reports.
-pub fn instances_behind(client: &ServiceClient, scrapes: usize) -> usize {
-    let series = "api_requests_total{endpoint=\"/metrics\",method=\"GET\",status=\"200\"}";
-    let mut observed = Vec::new();
+/// The header every response of this service carries, naming the process that
+/// produced it.
+pub const INSTANCE_HEADER: &str = "x-ocs-instance";
 
-    for _ in 0..scrapes {
-        let body = match client.get("/metrics") {
-            Ok(response) => {
-                assert_eq!(response.status, 200, "/metrics must answer 200");
-                response.text()
-            }
+/// The variable naming the individual replicas, when an operator has them.
+///
+/// Comma-separated base URLs, one per instance, bypassing the balancer. Some
+/// contracts cannot be checked through a balanced address at all — "this
+/// replica answers for itself" needs a way to reach that replica — and this is
+/// how a test gets one. Unset, those tests skip and say why.
+pub const INSTANCE_URLS_VARIABLE: &str = "OCS_INTEGRATION_INSTANCE_URLS";
+
+/// The instance that produced a response, if it said.
+///
+/// A deployment predating the header answers `None`, which every caller treats
+/// as "cannot attribute" rather than as a failure.
+#[must_use]
+pub fn responding_instance(response: &Response) -> Option<String> {
+    response.header(INSTANCE_HEADER).map(str::to_string)
+}
+
+/// One client per replica, when `OCS_INTEGRATION_INSTANCE_URLS` names them.
+///
+/// # Panics
+///
+/// When the variable is set to something that is not a usable base URL, which
+/// is a configuration mistake rather than an absent deployment.
+#[must_use]
+pub fn instance_clients() -> Vec<ServiceClient> {
+    let Some(raw) = optionchain_simulator::utils::env::read_var(INSTANCE_URLS_VARIABLE) else {
+        return Vec::new();
+    };
+
+    raw.split(',')
+        .map(str::trim)
+        .filter(|url| !url.is_empty())
+        .map(|url| match ServiceClient::new(url) {
+            Ok(client) => client,
+            Err(error) => panic!("{INSTANCE_URLS_VARIABLE} names an unusable URL: {error}"),
+        })
+        .collect()
+}
+
+/// How many distinct processes answered `probes` requests.
+///
+/// Counted from the identity each response carries, which is exact: two
+/// responses with the same value came from one process and two values came
+/// from two. A deployment that predates the header reports `None`, and a
+/// caller then knows it cannot attribute rather than guessing from a counter.
+#[must_use]
+pub fn instances_behind(client: &ServiceClient, probes: usize) -> Option<usize> {
+    let mut seen = std::collections::BTreeSet::new();
+
+    for _ in 0..probes {
+        let response = match client.get("/health") {
+            Ok(response) => response,
             Err(error) => panic!("{error}"),
         };
-        if let Some(value) = body
-            .lines()
-            .filter(|line| !line.starts_with('#'))
-            .find_map(|line| line.strip_prefix(series))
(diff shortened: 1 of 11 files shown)


examples/integration/src/lib.rs
```diff
@@ -616,6 +616,61 @@ pub fn report_cleanup(client: &ServiceClient, path: &str, what: &str) {
...
+/// How many distinct processes answered `scrapes` scrapes of `/metrics`.
+///
+/// A counter is per process and only ever grows, so one instance produces a
+/// non-decreasing sequence of scrapes. N instances taking turns produce N such
+/// sequences INTERLEAVED, which reads as a series that keeps stepping
+/// backwards. The instance count is therefore the smallest number of
+/// non-decreasing sequences the observations can be split into, which is what
+/// the greedy pass below computes.
+///
+/// It is a lower bound: two instances that happen to hold the same value are
```
joaquinbejar: 🟠 **Warning**: this detector is explicitly only a lower bound, but its callers treat it as the replica count and assume that a fixed multiple of requests crosses every instance. Two replicas can still produce one non-decreasing observation sequence, and a balancer may send every subsequent request to one replica, so the visibility, cursor and export tests can all pass without exercising another process. Keep this heuristic diagnostic-only and route contract checks to explicit per-replica targets, or require a stable instance identity that each response can report.

examples/integration/tests/replication.rs
```diff
@@ -0,0 +1,396 @@
...
+/// simulation through the balancer and compares it against a tape built from a
+/// second simulation with identical parameters, then requires each step to
+/// match the corresponding step of the other: a per-instance difference in how
+/// a step is priced would break exactly this.
+#[test]
+fn test_a_tape_is_the_same_whichever_instance_serves_it() {
+    let Some(client) = service() else {
+        return;
+    };
+    let (Some(left), Some(right)) = (
```
joaquinbejar: 🟠 **Warning**: both the subject and reference tapes are created and advanced through the same balanced client, so neither is the required standalone baseline. A replica-specific defect can affect both simulations identically, or both requests can land on the same replica, and this still passes. Issue #139 calls for one tape walked against a direct single-instance URL and the other through the balancer; add an explicit baseline client or target and compare those tapes step for step.

examples/integration/tests/replication.rs
```diff
@@ -0,0 +1,396 @@
...
+#[test]
+fn test_every_readiness_answer_is_consistent_with_itself() {
+    let Some(client) = service() else {
+        return;
+    };
+
+    let mut ready_answers = 0_usize;
+    let mut not_ready_answers = 0_usize;
+
+    for attempt in 0..(SCRAPES * 2) {
```
joaquinbejar: 🟠 **Warning**: this never creates the one-replica dependency failure the contract is about. On the normal healthy fixture every iteration is 200 with all dependencies up, so the test proves only that the response body agrees with its own HTTP status; an implementation that masks an unhealthy replica still passes. Add a controlled fixture that breaks one dependency for exactly one known replica, then assert that replica returns 503 while the healthy target remains 200 and that both are observed through the deployment path.

joaquinbejar (commented):
REQUEST_CHANGES. The suite remains able to pass without proving that different replicas served requests, it has no direct single instance baseline, and it never exercises a degraded replica. The incremental diff changes 517 lines, so please split the metrics detector and operational repair from the replication scenarios.

examples/integration/src/lib.rs
joaquinbejar: You are right, and the root cause was that nothing in a response said who produced it. Every response now carries x-ocs-instance, an opaque per-process value, so the counter is exact rather than a lower bound and a deployment without one reports "cannot attribute" instead of guessing. The checks that need a known replica go through OCS_INTEGRATION_INSTANCE_URLS rather than through the heuristic.

examples/integration/tests/replication.rs
joaquinbejar: Agreed, two tapes through the same balanced client is not a baseline. The baseline is walked against one named replica now and the subject through the balancer, and without the replica URLs the test skips saying why, since a single-instance baseline cannot be established through a balancer. Against two processes it reports: baseline from 1 instance, balanced walk served by 2.

examples/integration/tests/replication.rs
joaquinbejar: Half of this is fixed here and half is honestly out of reach of a client. The test now asks every replica directly and requires a distinct identity per address, so each answer is that instance own view rather than whichever one the balancer picked. Breaking a dependency for exactly one replica needs a fixture that owns the containers, in the shape of the mongo upgrade script, so it is issue #145 rather than a weaker assertion pretending to cover it.

joaquinbejar:
All three came from the same gap and it is closed: responses now carry an instance identity, so the suite proves it crossed instances rather than assuming it. One half of the third finding is issue #145, because a client cannot break a dependency on one replica.

PR #142, opened 2026-08-31
Share built factor tapes between instances

### LINKED ISSUE #136: The tape, snapshot and walk caches are per process, so replicas rebuild each other's work

joaquinbejar:
Closes #136

Base branch: `stack/2-139-replication-tests`
Stacked on: #141
Blocks: #143

Stack: #138 → #139 → **#136** → #135 → #137

## What changed

A built factor tape is written to Redis and read from there by any instance that needs it. Behind a balancer, consecutive steps of one simulation land on different instances, and each was rebuilding what its neighbour had just built — the hit rate falling towards 1/N on the one operation that walks every step from the seed.

**Correctness was never at risk**, and the change is written as the cost fix it is: a tape is a pure function of the stored parameters and the seed, so whichever instance answers produces the same numbers. What is saved is the rebuild.

## The key carries what decides the values

Not just the identity:

- the **snapshot generation**, so a build that changes what a step means cannot read a tape written under the old meaning
- a **UUID v5 fingerprint of the serialised parameters**, so a document rewritten under one id cannot be served the tape of the parameters it used to have. Derived from the whole serialisation rather than a hand-picked list of fields, so a new field that changes the walk is covered the day it is added rather than the day someone remembers this file
- the simulation id

## Failure is a miss, never an error

Every path through the shared cache degrades to the behaviour without it: an unreachable Redis, a document that will not decode, a tape that will not encode. Each costs a rebuild and logs; none costs a request. The in-process cache stays in front, since a local hit costs nothing where a shared one costs a round trip, and a shared hit still populates it.

The TTL matches the simulation retention window: a tape outliving its simulation is memory nobody can use, one expiring first is a rebuild.

## A decision I took

The trait trades in **encoded** tapes rather than in `FactorTape`. The domain module is private to this crate, so a public trait carrying that type would leak a private one, and the encoding belongs to the manager that owns both sides. It also keeps the store layer ignorant of a domain type, which is the direction the layering runs.

**The snapshot cache stays per instance**, which is the issue's second item. Rebuilding a snapshot from a cached tape is far cheaper than building the tape, and snapshot payloads are orders of magnitude larger; sharing them is a separate change with a different risk profile. The issue's acceptance criteria are all about the tape, and they are met.

Two processes over one Redis:

```
7099  DEBUG domain::factors: Built the factor tape steps=6 seed=42
7100  DEBUG store::tape_cache: shared tape cache hit key=optionchain:tape:4:5587…:2794…
```

Plus three unit tests over two managers sharing one cache: the second serves a tape it did not build (proven by the write count, since a build always writes), an unreachable cache degrades to building rather than failing, and two simulations never share an entry.

## Checklist

- [x] Reproducibility intact: the tape is unchanged, only where it comes from
- [x] Stored-session shape untouched; REST DTOs untouched
- [x] `make pre-push` clean, zero warnings, 935 tests
- [x] Patch version bumped to 0.2.21 with the compose image tag

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -72,11 +72,13 @@ use optionchain_simulator::infrastructure::{
     init_mongodb, resolve_log_level_from_env,
 };
 use optionchain_simulator::session::{
-    InRedisSessionStore, InRedisSimulationStore, SessionManager, SimulationManager,
+    DEFAULT_TAPE_KEY_PREFIX, InRedisSessionStore, InRedisSimulationStore, RedisTapeCache,
+    SessionManager, SimulationManager,
 };
 use optionstratlib::utils::setup_logger_with_level;
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::Duration;
 use tracing::{info, warn};
 
 /// The `main` function is the entry point of the application using the Actix Web server framework.
@@ -188,9 +190,19 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // simulations live.
     let v2_config = SimulationV2Config::from_env()?;
     let simulation_store = Arc::new(InRedisSimulationStore::new(
-        redis_client_v2,
+        Arc::clone(&redis_client_v2),
         None, // the documented v2 prefix
         Some(v2_config.retention_secs()),
+    ));
+    // Built tapes are shared through the same Redis, so a step served by a
+    // different instance does not rebuild what its neighbour already walked
+    // (issue #136). The window matches the simulations': a tape that outlives
+    // its simulation is memory nobody can use, and one that expires first
+    // costs a rebuild.
+    let shared_tapes = Arc::new(RedisTapeCache::new(
+        Arc::clone(&redis_client_v2),
+        DEFAULT_TAPE_KEY_PREFIX,
+        Duration::from_secs(v2_config.retention_secs()),
     ));
     // Snapshot persistence is opt-in (`OCS_SNAPSHOT_PERSISTENCE_ENABLED`). When
     // it is off the manager never learns the feature exists; when it is on, the
@@ -202,7 +214,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // export reads back through the same handle — the routes take it off the
     // manager rather than being passed a second one, because two handles could
     // be configured differently and there is only ever one warehouse.
-    let mut simulation_manager = SimulationManager::new(simulation_store, v2_config);
+    let mut simulation_manager =
+        SimulationManager::new(simulation_store, v2_config).with_shared_tapes(shared_tapes);
     match ClickHouseSnapshotRepository::from_env()? {
         Some(warehouse) => {
             warehouse.ensure_schema().await?;
--- a/src/main.rs
+++ b/src/main.rs
@@ -203,6 +203,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         Arc::clone(&redis_client_v2),
         DEFAULT_TAPE_KEY_PREFIX,
         Duration::from_secs(v2_config.retention_secs()),
+        // The same knob that bounds this instance's own map bounds the shared
+        // cache, so the number an operator writes is how many tapes the
+        // deployment keeps rather than how many each replica keeps.
+        v2_config.max_cached_tapes,
     ));
     // Snapshot persistence is opt-in (`OCS_SNAPSHOT_PERSISTENCE_ENABLED`). When
     // it is off the manager never learns the feature exists; when it is on, the
--- a/src/session/manager_v2.rs
+++ b/src/session/manager_v2.rs
@@ -27,7 +27,7 @@ use crate::domain::series::{SeriesBuilder, SeriesSnapshot, SnapshotCache};
 use crate::infrastructure::{SimulationSnapshotRepository, SimulationV2Config, SnapshotRecord};
 use crate::session::model::SessionState;
 use crate::session::snapshot_record::{snapshot_quote_count, snapshot_record};
-use crate::session::store::SimulationStore;
+use crate::session::store::{SharedTapeCache, SimulationStore, tape_key};
 use crate::session::{SessionV2, SimulationParametersV2};
 use crate::utils::ChainError;
 use std::collections::HashMap;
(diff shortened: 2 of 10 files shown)


src/session/store/tape_cache.rs
```diff
@@ -0,0 +1,241 @@
...
+            }
+        }
+    }
+
+    async fn put(&self, key: &str, encoded: &str) {
+        let key = format!("{}{key}", self.prefix);
+        let seconds = self.ttl.as_secs().max(1);
+        if let Err(error) = self
+            .client
+            .set(&key, encoded.to_string(), Some(seconds))
```
joaquinbejar: 🟠 **Warning**: each tape is written as an independent TTL key, but there is no deployment-wide count, eviction policy, or delete operation. `OCS_MAX_CACHED_TAPES` still bounds only each manager’s local map, and deleting or reaping a simulation leaves its shared key until TTL. This misses issue #136’s explicit per-deployment bound and can let Redis grow to every tape built during the retention window. Carry the configured bound into this cache and enforce insertion, recency, and eviction atomically.

src/session/manager_v2.rs
```diff
@@ -617,6 +642,14 @@ impl SimulationManager {
 
         let id = simulation.id;
 
+        // Another instance may already have walked this. Consulted before the
+        // build lock rather than after, so a hit costs one round trip instead
+        // of queueing behind a build that is about to be redundant.
+        if let Some(tape) = self.shared_tape(simulation).await {
```
joaquinbejar, line 650: 🟠 **Warning**: this awaited lookup widens the owner-election race. Another local request can finish and cache the tape while this Redis GET is in flight; if this GET already observed a miss, this caller can then find no active builder and rebuild an already-warm tape. Recheck the local cache after the await before electing a new owner.

src/session/manager_v2.rs
```diff
@@ -650,6 +683,13 @@ impl SimulationManager {
 
         let result = self.build_tape(simulation).await;
 
+        // Offer it to the other instances. After the local cache, so a failure
+        // here costs nothing this process needs, and before the broadcast, so
+        // a waiter that goes on to serve is not racing the write.
+        if let Ok(tape) = &result {
+            self.share_tape(simulation, tape).await;
```
joaquinbejar: 🟠 **Warning**: this awaits the remote write before removing the single-flight entry and broadcasting the completed tape. A slow Redis write delays every local waiter, and cancellation during this await leaves the sender stored in `builds`, so subscribed requests can wait forever even though the local tape is warm. Publish locally first and send the shared write through a bounded best-effort path.

src/main.rs
```diff
@@ -188,10 +190,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
         None, // the documented v2 prefix
         Some(v2_config.retention_secs()),
     ));
+    // Built tapes are shared through the same Redis, so a step served by a
+    // different instance does not rebuild what its neighbour already walked
+    // (issue #136). The window matches the simulations': a tape that outlives
+    // its simulation is memory nobody can use, and one that expires first
+    // costs a rebuild.
+    let shared_tapes = Arc::new(RedisTapeCache::new(
+        Arc::clone(&redis_client_v2),
```
joaquinbejar, line 203: 🟠 **Warning**: the shared cache uses the same Redis client as the authoritative simulation store above, and peek or advance reads that store before consulting the cache. A Redis outage therefore fails at `store.get`; it does not degrade to an in-process rebuild as the new docs and acceptance criterion claim. The `BrokenTapes` test combines a broken cache with an in-memory store and does not represent this production wiring. Scope the promise to cache-command failures after a successful state read, or use an independently degradable backend.

src/session/manager_v2.rs
```diff
@@ -1662,6 +1808,133 @@ mod tests {
...
+        assert_eq!(
+            second.cached_tapes(),
+            1,
+            "a shared hit must still populate the local cache, or every step pays the round trip"
+        );
+
+        // And the answer is the same market, which is the whole point of
+        // sharing rather than rebuilding.
+        assert_eq!(
+            served_first.1.spot, served_second.1.spot,
```
joaquinbejar: 🟠 **Warning**: step zero’s spot is always the initial price, so this assertion does not verify that the stochastic tape survived the shared-cache round trip. Compare the complete `FactorTape` before and after encoding, or serve and compare every row including later spots and base volatility, to pin the same-seed identical-tape contract across instances.

joaquinbejar (commented):
REQUEST_CHANGES. The shared tape path lacks the required deployment wide bound, keeps local waiters behind remote cache operations, can duplicate builds around an in flight lookup, and does not meet the production fallback or full tape equality acceptance criteria. The incremental diff changes 596 lines, so please split the Redis adapter and bounds from the manager wiring and tests.

src/session/store/tape_cache.rs
joaquinbejar: You are right on all three parts, and the bound was the one I claimed and did not build. The write is a Lua script now that inserts, records recency and evicts past OCS_MAX_CACHED_TAPES atomically, so two instances cannot both see room, and delete and cleanup drop what a gone simulation left rather than waiting out the TTL. A live test caught a detail worth keeping: recency in seconds ties, and tied scores make the eviction order lexicographic, so it is milliseconds.

src/session/manager_v2.rs
joaquinbejar: Good catch, the await widens exactly that window. The local cache is rechecked after it, before electing an owner, so a caller whose neighbour warmed the tape mid-lookup serves it instead of rebuilding.

src/session/manager_v2.rs
joaquinbejar: Agreed, and the cancellation half is the worse one: waiters stuck on a tape already warm in this process. The local cache and the broadcast come first now, and the shared write follows, so a slow Redis costs the deployment a rebuild elsewhere and costs this request nothing.

src/main.rs
joaquinbejar: Fair, the promise was broader than the wiring supports: peek and advance read the store from the same Redis first, so an outage fails there rather than degrading to a rebuild. The docs now say what is true, that a failing cache command costs a rebuild rather than a request, and that what degrades is the saving and not the service dependence on its store.

src/session/manager_v2.rs
joaquinbejar: Right, step zero is the initial price whatever the walk did. The test compares the stored tape against one built directly from the same parameters, every row including the later spots and base volatilities, and asserts the walk actually moved so the comparison cannot be vacuous.

joaquinbejar:
All five fixed. The first was the one that mattered: I claimed a per-deployment bound and shipped independent TTL keys, so the cache could have grown to every tape built during the retention window.

PR #143, opened 2026-08-31
Make the pricing bound the deployment's, not the process's

### LINKED ISSUE #135: The pricing admission cap is per process, so replicas multiply it

joaquinbejar:
Closes #135

Base branch: `stack/3-136-shared-tape-cache`
Stacked on: #142
Blocks: #144

Stack: #138 → #139 → #136 → **#135** → #137

## The decision the issue asked for

The issue offered two: document the knob as per-instance, or make it a shared lease. I took **the shared lease**, because it is what an operator writing `OCS_MAX_CONCURRENT_PRICING_JOBS=4` means, and because the service already keeps its session state in Redis with atomic scripts, so the machinery is there rather than new.

## How it holds

A sorted set of leases scored by the instant each was taken. Acquiring is one Lua script — the expiry sweep, the count and the insert cannot interleave with another instance doing the same:

1. drop leases older than the window, which is what stops a killed instance holding one forever
2. if fewer than the limit remain, add this one
3. otherwise wait and ask again

The window is a fallback rather than a timer — a finished job releases — and it is minutes against a pricing job of seconds, so a running job cannot lose its lease to the sweep.

## It cannot fail a request

Every failure path degrades to the **per-process semaphore**, which is exactly the bound the service had before this existed: an unreachable Redis, a clock that will not read, a wait that runs out. Pricing unbounded would be a worse answer to a Redis outage than pricing more concurrently than configured, and refusing to price would be worse still.

The local semaphore also stays in front of the shared one: it is free and cancellable, so a client that goes away while queued costs nothing, and only a job that passed the cheap gate spends a round trip.

## Layering

The port (`SharedPricingGate`) lives in `utils::admission` and the Redis adapter in `infrastructure`. `utils` is what both the API renderers and the session manager depend on, so it must not depend on a driver; `main` wires the two.

## Verified against a live Redis

Two gates sharing a bound of two: both get a lease, **the third is refused**, and releasing one frees a slot for the other instance. Marked `#[ignore]` like the other live-service tests, so the default suite stays hermetic and CI's integration job runs it.

## Checklist

- [x] Reproducibility, stored-session shape and REST DTOs untouched
- [x] The knob documented as deployment-wide, with the fallback, in `.env.example` and the crate docs
- [x] `make pre-push` clean, zero warnings, 935 tests
- [x] Patch version bumped to 0.2.22 with the compose image tag

### DIFF
--- a/src/infrastructure/pricing_gate.rs
+++ b/src/infrastructure/pricing_gate.rs
@@ -1,0 +1,245 @@
+//! The deployment-wide pricing bound, held in Redis.
+//!
+//! `utils::admission` bounds the pricing jobs of ONE process, which is the
+//! wrong unit once the service runs replicated: an operator asking for four and
+//! running two replicas gets eight, and learns it when the host saturates
+//! (issue #135). This is the adapter that makes the same number mean the
+//! deployment.
+//!
+//! # How it holds
+//!
+//! A sorted set of leases, scored by the instant each was taken. Acquiring is
+//! one Lua script, so the expiry sweep, the count and the insert cannot
+//! interleave with another instance doing the same:
+//!
+//! 1. drop every lease older than the lease window — that is what stops a
+//!    killed instance from holding one forever;
+//! 2. if fewer than the limit remain, add this one and say yes;
+//! 3. otherwise say no, and let the caller wait and ask again.
+//!
+//! The window is a fallback, not a timer: a job that finishes releases its
+//! lease. It has to be longer than a pricing job or a running job would lose
+//! its lease to the sweep and let another in.
+//!
+//! # It never fails a request
+//!
+//! Every error here is reported as "no lease", and `admit_blocking` then
+//! proceeds under the per-process bound, which is the behaviour the service
+//! had before this existed. An unreachable Redis makes the deployment price
+//! more concurrently than configured; refusing to price at all would be a worse
+//! answer to the same outage.
+
+use crate::infrastructure::RedisClient;
+use crate::utils::admission::SharedPricingGate;
+use async_trait::async_trait;
+use std::sync::Arc;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use tracing::{debug, warn};
+use uuid::Uuid;
+
+/// The key the leases live under.
+pub const DEFAULT_PRICING_GATE_KEY: &str = "optionchain:pricing:leases";
+
+/// How long a lease survives without being released.
+///
+/// Longer than any pricing job: a running job whose lease expired would let a
+/// second one in, which is the bound leaking. A snapshot at the contract cap
+/// is seconds, so minutes is the safe side of that.
+const LEASE_WINDOW: Duration = Duration::from_secs(300);
+
+/// How long to sleep between attempts when every lease is taken.
+///
+/// Short enough that a freed lease is picked up promptly, long enough that a
+/// queue of waiters does not become a load test of its own.
+const RETRY_INTERVAL: Duration = Duration::from_millis(50);
+
+/// Acquires a lease if the deployment is under its bound.
+///
+/// `KEYS[1]` the sorted set; `ARGV[1]` now in milliseconds, `ARGV[2]` the lease
+/// window in milliseconds, `ARGV[3]` the limit, `ARGV[4]` this lease's token.
+/// Returns 1 when the lease was taken and 0 when the bound is full.
+const ACQUIRE_SCRIPT: &str = r"
+local now = tonumber(ARGV[1])
+local window = tonumber(ARGV[2])
+local limit = tonumber(ARGV[3])
+
+redis.call('ZREMRANGEBYSCORE', KEYS[1], '-inf', now - window)
+if redis.call('ZCARD', KEYS[1]) < limit then
+  redis.call('ZADD', KEYS[1], now, ARGV[4])
+  redis.call('PEXPIRE', KEYS[1], window * 2)
+  return 1
+end
+return 0
+";
+
+/// A pricing bound shared by every instance through Redis.
+pub struct RedisPricingGate {
+    client: Arc<RedisClient>,
+    key: String,
+    limit: usize,
+}
+
+impl RedisPricingGate {
+    /// Builds a gate allowing `limit` pricing jobs across the deployment.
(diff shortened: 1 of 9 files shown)


src/infrastructure/pricing_gate.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /// One attempt. `Ok(true)` took a lease, `Ok(false)` the bound is full.
+    async fn try_acquire(&self, token: &str) -> Result<bool, String> {
+        let now = Self::now_millis().ok_or_else(|| "the clock is before the epoch".to_string())?;
+        let mut conn = self.client.connection_manager();
+
+        let taken: i64 = redis::Script::new(ACQUIRE_SCRIPT)
+            .key(&self.key)
+            .arg(now.to_string())
```
joaquinbejar: 🟠 **Warning**: lease expiry is based on each instance's SystemTime and the score is never renewed. Clock skew can make one replica reap another's live lease, and a valid job can outlive the fixed 300 second window and lose its lease while still running. Use Redis TIME inside the script and renew the lease until the supervised blocking job finishes.

src/infrastructure/pricing_gate.rs
```diff
@@ -0,0 +1,245 @@
...
+
+            let waited = started.elapsed().unwrap_or(Duration::ZERO);
+            if waited >= deadline {
+                warn!(
+                    limit = self.limit,
+                    waited_secs = waited.as_secs(),
+                    "waited for a deployment-wide pricing lease without getting one; proceeding \
+                     under this instance's own bound"
+                );
+                return None;
```
joaquinbejar: 🟠 **Warning**: None conflates an unreachable gate with a healthy gate at capacity. After 30 seconds of saturation every replica falls back and admit_blocking starts extra jobs under only its local semaphore, so the deployment wide cap is bypassed exactly under sustained load. Keep waiting cancellably when Redis reports full, or return a distinct saturated result and reserve local fallback for actual gate failures.

src/utils/admission.rs
```diff
@@ -77,12 +138,27 @@ where
...
+    // The local permit first, then the deployment-wide lease: the local one is
+    // free and cancellable, so a client that goes away while queued costs
+    // nothing, and only a job that has already passed the cheap gate spends a
+    // round trip on the shared one.
+    let lease = match SHARED_GATE.get() {
+        Some(gate) => gate.acquire(SHARED_LEASE_WAIT).await,
+        None => None,
+    };
+
     let outcome = tokio::task::spawn_blocking(job)
```
joaquinbejar: 🟠 **Warning**: the lease is owned by this cancellable request future, while spawn_blocking keeps running after its JoinHandle is dropped. A disconnect here releases the local permit immediately and skips the Redis release, so CPU work continues outside the local bound and the shared slot stays occupied until expiry. Move ownership of both permits into a supervisor that outlives the caller and releases them only when the blocking work actually finishes, then cover cancellation and panic with tests.

joaquinbejar (commented):
REQUEST_CHANGES. The distributed gate can exceed its configured deployment bound under clock skew and healthy saturation, and cancellation or a long valid pricing job can detach the lease lifetime from the actual blocking work. The layering and stack metadata are otherwise consistent and all checks pass.

src/infrastructure/pricing_gate.rs
joaquinbejar: Both halves were real. The expiry now reads TIME from the Redis server inside the script, so one replica cannot reap another on a skewed clock, and a supervisor renews a held lease every hundred seconds until the job actually ends. Renewal only refreshes a lease still in the set, since re-adding a reaped one would push the deployment past its bound.

src/infrastructure/pricing_gate.rs
joaquinbejar: Agreed, and that is the worst possible moment to fall back. A full gate is no longer a failure: the caller waits for as long as it takes, cancellably, because the wait lives in its own future. Only a gate that cannot be reached returns None and drops to the local semaphore.

src/utils/admission.rs
joaquinbejar: You are right, and this was the one that let unbounded pricing through. Both permits are handed to a supervisor task that outlives the request and gives them back only when the blocking work finishes, so a disconnect no longer frees a slot the CPU is still using. Cancellation and panic both have tests now, and everything before the spawn stays cancellable so a queued client still costs nothing.

joaquinbejar:
All three fixed. The third was the one that mattered most: a client disconnecting released the bound while its pricing job kept running, so a burst of abandoned requests priced unbounded on both the local and the shared cap.

PR #144, opened 2026-08-31
One instance builds a tape, not one per instance

### LINKED ISSUE #137: Tape build coalescing is per process, so replicas duplicate the same build

joaquinbejar:
Closes #137

Base branch: `stack/4-135-shared-admission`
Stacked on: #143
Stack: #138 → #139 → #136 → #135 → **#137** (last)

The diff is cumulative until the PRs below merge, so read it against the base branch rather than against `main`. This is the one PR in the chain with a hard dependency: the build claim needs the shared cache from #136.

## What changed

`SimulationManager::builds` coalesces concurrent builds of one tape within a process: the first caller builds, the rest wait. Across instances that map is invisible, so N replicas could each believe they were the only builder and run N copies of exactly the job the coalescing exists to avoid.

A build is now claimed in the same Redis that holds the tape, before it starts, with `SET NX` and an expiry. Two instances cannot both own it, and an instance killed mid-build cannot wedge the others because the claim expires on its own.

## Waiting, and giving up

A waiter polls the shared cache rather than a channel — there is no channel between processes — and gives up after a bounded wait to build the tape itself. That is duplicated work in a rare case, weighed against a request that never answers because it waited on an owner that had died.

Everything else fails towards building too: an unreachable cache **grants** the claim, and a release that does not land expires. Nothing here can turn a Redis problem into an unanswered request.

The wait is a field rather than a constant, so the give-up path is tested in 200 ms instead of the 30 seconds production wants.

Two processes over one Redis, both asked for the same 400-step simulation at once:

```
7099  0 builds
7100  1 build   ("claimed the build" … "Built the factor tape")
7099        DEBUG peek: another instance is building this tape; waiting for it
```

One build, one waiter, one tape. Plus two unit tests over two managers: only one writes a tape when both peek simultaneously, and an abandoned claim does not stop the next caller building.

## Checklist

- [x] One build of a given tape at a time across the deployment
- [x] A claim that expires, so a killed builder does not wedge the others
- [x] A waiter that times out builds it itself rather than failing
- [x] Two managers over one cache prove only one builds
- [x] `make pre-push` clean, zero warnings, 937 tests
- [x] Patch version bumped to 0.2.23 with the compose image tag

### DIFF
--- a/src/session/manager_v2.rs
+++ b/src/session/manager_v2.rs
@@ -33,10 +33,30 @@ use crate::utils::ChainError;
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
-use std::time::Instant;
+use std::time::{Duration, Instant};
 use tokio::sync::{broadcast, mpsc};
 use tracing::{debug, info, instrument, warn};
 use uuid::Uuid;
+
+/// How long a caller waits for the instance that claimed a build.
+///
+/// Long enough for a tape of the maximum step count, short enough that an
+/// owner which died leaves the others building rather than waiting out the
+/// claim's own expiry.
+const SHARED_BUILD_WAIT: Duration = Duration::from_secs(30);
+
+/// How often a waiter looks for the tape the owner is building.
+const SHARED_BUILD_POLL: Duration = Duration::from_millis(50);
+
+/// What asking for the deployment-wide build claim produced.
+enum SharedBuild {
+    /// This caller builds.
+    Owner,
+    /// Another instance built it, and here it is.
+    Waited(FactorTape),
+    /// Another instance owns the claim but published nothing in time.
+    Unavailable,
+}
 
 /// One cached factor tape and the last time it was used.
 struct TapeEntry {
@@ -90,6 +110,13 @@ pub struct SimulationManager {
     /// failure of the shared one is treated as `None` for that call: a cache
     /// that cannot be reached is a miss, never a failed request.
     shared_tapes: Option<Arc<dyn SharedTapeCache>>,
+    /// How long to wait for the instance that claimed a build before building
+    /// it here.
+    ///
+    /// A field rather than a constant so a test can prove the give-up path
+    /// without waiting out the production value, which is deliberately as long
+    /// as the slowest build.
+    shared_build_wait: Duration,
     /// Where served snapshots are queued for filing, when the operator turned
     /// persistence on. `None` is the default and the whole feature is then
     /// absent from the serving path — no connection, no latency, no failure
@@ -155,6 +182,7 @@ impl SimulationManager {
                 config.max_cached_snapshot_contracts,
             ))),
             shared_tapes: None,
+            shared_build_wait: SHARED_BUILD_WAIT,
             warehouse: None,
         }
     }
@@ -168,6 +196,17 @@ impl SimulationManager {
     #[must_use]
     pub fn with_shared_tapes(mut self, cache: Arc<dyn SharedTapeCache>) -> Self {
         self.shared_tapes = Some(cache);
+        self
+    }
+
+    /// How long to wait for another instance's build before building here.
+    ///
+    /// Exists for tests: production wants the default, which is sized for the
+    /// slowest build, and a test that has to reach the give-up path should not
+    /// spend that long doing it.
+    #[must_use]
+    pub fn with_shared_build_wait(mut self, wait: Duration) -> Self {
+        self.shared_build_wait = wait;
         self
     }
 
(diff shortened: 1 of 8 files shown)


src/session/store/tape_cache.rs
```diff
@@ -0,0 +1,304 @@
...
+                // waiting for a claim nobody can grant costs the client.
+                warn!(%error, %key, "the build claim could not be taken; building anyway");
+                true
+            }
+        }
+    }
+
+    async fn release_build(&self, key: &str) {
+        let key = format!("{}build:{key}", self.prefix);
+        if let Err(error) = self.client.delete(&key).await {
```
joaquinbejar: 🟠 **Warning**: this unconditional DEL is not ownership safe. If this builder's 120 second claim expires and another instance acquires the key, the stale builder deletes the new owner's claim and a third builder can enter. Return a unique token and release with an atomic compare and delete only when the stored value still matches, then cover expiry and takeover with the required two manager live Redis test.

src/session/manager_v2.rs
```diff
@@ -648,7 +720,36 @@ impl SimulationManager {
...
 
-        let result = self.build_tape(simulation).await;
+        // Within this process one caller builds and the rest wait, which is
+        // what `builds` above decided. Across instances the same question is
+        // asked of the shared cache: without it, N instances all believing
+        // they are the only builder run N copies of the one job the
+        // coalescing exists to avoid (issue #137).
+        let result = match self.claim_shared_build(simulation).await {
+            SharedBuild::Owner => {
+                let built = self.build_tape(simulation).await;
```
joaquinbejar: 🟠 **Warning**: the request future owns both the Redis claim and local broadcast entry across this cancellable await, while spawn_blocking continues after cancellation. A disconnect can finish and cache the tape locally but skip sharing it, releasing the claim, removing builds and notifying existing subscribers, so those subscribers can hang indefinitely. Run the owner workflow in a task or guard whose lifetime follows the blocking build, and test a dropped owner.

src/session/manager_v2.rs
```diff
@@ -701,6 +802,94 @@ impl SimulationManager {
...
+        }
+
+        debug!(
+            simulation_id = %simulation.id,
+            "another instance is building this tape; waiting for it"
+        );
+        let deadline = Instant::now() + self.shared_build_wait;
+        while Instant::now() < deadline {
+            tokio::time::sleep(SHARED_BUILD_POLL).await;
+            if let Some(tape) = self.shared_tape(simulation).await {
```
joaquinbejar: 🟠 **Warning**: each 50 ms poll calls shared_tape, which recomputes tape_key by serializing all parameters. Historical simulations may contain 100000 prices, so a 30 second wait performs about 600 large synchronous serializations per replica on the async runtime. Pass the key computed above into the polling read and reuse it.

joaquinbejar (commented):
REQUEST_CHANGES. The distributed single flight can delete a newer owner's claim, cancellation can strand both local and Redis waiters, and polling repeatedly serializes large replay inputs on the async path. The stack and contract metadata are otherwise consistent and all checks pass.

src/session/store/tape_cache.rs
joaquinbejar: You are right, and the takeover is exactly the case that breaks it. Claiming returns a token now and releasing is a compare-and-delete that only removes a claim still carrying it, so a slow builder gives back nothing rather than the successor key. Covered by two caches over one live Redis, including the stale release and the successor release after it.

src/session/manager_v2.rs
joaquinbejar: Agreed, this was the same shape as the admission fix and worse here, because the waiters had nobody left to publish to them. The owner workflow runs in a task that outlives the request and the owner waits on the same broadcast as everyone else, so a disconnect cancels only the wait while the build, the publish, the share and the release still happen. A dropped owner has a test.

src/session/manager_v2.rs
joaquinbejar: Good catch, that is six hundred serializations of a historical parameter set per waiter. The key is computed once by the caller and passed down through the shared lookup, the claim, the poll loop, the share and the release, so no path recomputes it.

joaquinbejar:
All three fixed. The middle one was the worst: a client that disconnected mid-build left the local waiters with nobody to publish to them, the deployment-wide claim held until expiry, and the finished tape unshared.

PR #147, opened 2026-08-31
Prove readiness is per replica by isolating one from Redis (#145)

### LINKED ISSUE #145: A fixture that breaks one replica's dependency and proves the probes report it
From the review of #141, which asked for a test proving that a replica whose dependency is down reports itself unready while its healthy neighbour keeps serving.

## Why it is not in that PR

The integration suite talks to a deployment over HTTP. It can ask every replica for its own readiness — that landed in #141, using the per-replica URLs and the instance identity each response now carries — but it cannot BREAK a dependency for exactly one replica. Stopping Redis for one container and not the other is an operator action, so the test needs a fixture that owns the containers rather than a client that talks to them.

## What to build

A compose-based fixture, in the shape of `scripts/mongo_upgrade_test.sh`, which already does this kind of thing for the MongoDB upgrade:

1. bring up two service instances and their dependencies
2. cut ONE instance off from Redis, by disconnecting it from the network rather than by stopping Redis, so the other stays healthy
3. assert the isolated instance answers `503` from `/ready` naming Redis as `unreachable`, while the other answers `200`
4. assert the identity in each answer is the instance that was asked, so the assertion is about the replica and not about whichever one the balancer chose
5. restore the network and assert the isolated instance reports itself ready again without a restart, which is what `/ready` promises

## Acceptance criteria

- [ ] A script that arranges the failure and cleans up after itself, whatever the outcome
- [ ] The failing replica is identified by the `x-ocs-instance` header rather than assumed
- [ ] Recovery without a restart is asserted, not only the failure
- [ ] Run on demand, and in CI on a change to the health module or the compose files
- [ ] Documented next to the probes in the crate docs, since what it proves is what an orchestrator relies on

joaquinbejar:
## Summary

The crate docs promise that a replica whose dependency is gone reports itself unready while its neighbour keeps serving, and that it recovers without a restart. That promise is what an orchestrator acts on, and until now nothing executed it: the integration suite can ask each replica for its readiness and tell which one answered, but it cannot break a dependency for exactly one replica, because that is an operator action on the containers rather than something a client of the deployment can arrange.

`scripts/replica_isolation_test.sh` arranges it and asserts the whole cycle, and CI runs it on a change to the probes, to what they probe, or to how the service is packaged.

## Changes

- Two replicas run over shared dependencies on two networks, Redis on one and MongoDB on the other, and one replica is disconnected from the Redis network. Redis is never stopped: that would break both replicas and prove nothing about a per-instance signal, and keeping MongoDB reachable is what makes the readiness body name the dependency that failed rather than all of them.
- Each probe is addressed to ONE replica by container name from a curl container sitting on the network the isolation does not touch. No published ports and no balancer, so every assertion is about a named replica, checked against the `x-ocs-instance` header the answer carries.
- The isolated replica must answer 503 with `not_ready`, `redis` down under one of the two fixed reason categories, and `mongodb` still up. Its neighbour must still answer 200 as itself.
- Liveness is asserted alongside: `/health` must stay 200 on the isolated replica, since a dependency failure that took it down would have an orchestrator restart a healthy process and turn one outage into two.
- Recovery is asserted, not only the failure. The network is reconnected and the SAME instance id must report itself ready again; the id is generated once per process, so an unchanged id is the proof that no restart happened.
- `.github/workflows/replica-isolation.yml` builds the image and runs the fixture on `workflow_dispatch` and on a pull request touching `health.rs`, `instance.rs`, the infrastructure probes, the Dockerfile, the compose file or the fixture itself. Deliberately not on every push: it builds the service image and runs five containers.
- The crate docs gain the paragraph that says what the fixture proves, next to the probes it is about.

## Technical decisions

- **curl rather than the runtime image's busybox wget.** The first version read the probes with `wget -S` from inside each container. It works for a 200 and not for the answer this fixture exists to produce: busybox does not hand back the response headers of a 503, so the `x-ocs-instance` attribution was missing exactly where it matters. The curl image is pinned by digest like every other image here.
- **Created, then attached, then started.** A replica started without Redis exits rather than coming up unready, so both networks are attached to a created container before it starts. Otherwise the fixture would be testing the startup path instead of the probes.
- **A loud abort.** An `ERR` trap reports the line, because `set -e` ends a run in silence and a fixture that stops halfway looks exactly like one that passed a step it never reached.
- **No new crate dependency.** Everything here is a shell script and a workflow.

## Testing

- [x] `make pre-push` run locally and clean, strict clippy clean, 944 tests passing
- [x] The fixture itself run locally against an image built from this branch: both replicas ready, the isolated one 503 naming `redis` with MongoDB up, `/health` still 200, the neighbour 200 as itself, and the same instance id ready again after the network came back
- [x] The fixture proved it can fail: an earlier revision was caught by its own attribution assertion when the 503 carried no instance header

## Checklist

- [x] Code follows `rules/global_rules.md` and `CLAUDE.md`
- [x] No Rust behaviour changed: this adds a fixture, a workflow and documentation
- [x] `version` in the root `Cargo.toml` bumped to 0.2.24, patch, with the compose default image tag moved with it
- [x] `src/lib.rs` docs updated and `README.md` regenerated via `make readme`

Closes #145

### DIFF
--- a/.github/workflows/replica-isolation.yml
+++ b/.github/workflows/replica-isolation.yml
@@ -7,7 +7,7 @@
 # which is what scripts/replica_isolation_test.sh has and a client of the
 # deployment does not.
 #
-# Deliberately NOT on every push: it builds the service image and runs four
+# Deliberately NOT on every push: it builds the service image and runs five
 # containers, and it has nothing to say about a change that touches neither the
 # probes, nor what they probe, nor how the service is packaged.
 name: Replica isolation (readiness per instance)
@@ -15,10 +15,21 @@ name: Replica isolation (readiness per instance)
 on:
   workflow_dispatch:
   pull_request:
+    # Everything that decides what this fixture observes, not only the modules
+    # named after it. The probe bodies and the instance header are the visible
+    # half; the wiring that chooses which dependencies are probed, mounts the
+    # routes and installs the header can regress on its own, and the clients
+    # the probes call decide what "unreachable" even means.
     paths:
       - "src/api/rest/health.rs"
       - "src/api/rest/instance.rs"
+      - "src/api/rest/routes.rs"
+      - "src/api/rest/controller.rs"
       - "src/infrastructure/health.rs"
+      - "src/infrastructure/redis/**"
+      - "src/infrastructure/mongodb/**"
+      - "src/infrastructure/config/**"
+      - "src/main.rs"
       - "Docker/Dockerfile"
       - "Docker/docker-compose.yml"
       - "scripts/replica_isolation_test.sh"
--- a/.github/workflows/replica-isolation.yml
+++ b/.github/workflows/replica-isolation.yml
@@ -1,0 +1,41 @@
+# Proves the readiness probe is per REPLICA, not per deployment.
+#
+# The Integration job runs one instance against healthy dependencies, so it
+# sees a 200 and nothing else. What an orchestrator actually relies on is the
+# other case: one replica loses a dependency, says so about ITSELF, and its
+# neighbour keeps serving. That needs an operator's hands on the containers,
+# which is what scripts/replica_isolation_test.sh has and a client of the
+# deployment does not.
+#
+# Deliberately NOT on every push: it builds the service image and runs four
+# containers, and it has nothing to say about a change that touches neither the
+# probes, nor what they probe, nor how the service is packaged.
+name: Replica isolation (readiness per instance)
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "src/api/rest/health.rs"
+      - "src/api/rest/instance.rs"
+      - "src/infrastructure/health.rs"
+      - "Docker/Dockerfile"
+      - "Docker/docker-compose.yml"
+      - "scripts/replica_isolation_test.sh"
+      - ".github/workflows/replica-isolation.yml"
+
+jobs:
+  isolation:
+    name: A replica reports its own broken dependency
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      # Built here rather than inside the script so the build is a step of its
+      # own in the log: a compile error is then a build failure rather than a
+      # fixture that appears to fail at its first assertion.
+      - name: Build the service image
+        run: docker build -f Docker/Dockerfile -t ocs-replica-isolation:ci .
+
+      - name: Isolate one replica from Redis
+        run: ./scripts/replica_isolation_test.sh ocs-replica-isolation:ci
--- a/scripts/replica_isolation_test.sh
+++ b/scripts/replica_isolation_test.sh
@@ -100,13 +100,21 @@ probe() {
         curl -sS -i --max-time 5 "http://$target:$PORT$path" 2>/dev/null || true
 }
 
+# Both readers below print an empty value rather than failing when what they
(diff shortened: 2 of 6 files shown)


joaquinbejar:
Copilot declined the first review with a quota limit, so that round happened in this session instead, over the fixture and the workflow rather than over Rust. Copilot reviewed on the retry and found two things worth fixing, both addressed above.

scripts/replica_isolation_test.sh
```diff
@@ -0,0 +1,274 @@
...
+#
+# Through a variable rather than through stdout on purpose: inside a command
+# substitution, `fail` would exit the SUBSHELL, and the run would end on the
+# assignment with nothing said about which wait ran out.
+wait_for_ready_code() {
+    local container="$1" want="$2" label="$3" code
+    local attempt
+    for attempt in $(seq 1 60); do
+        ANSWER=$(probe "$container" "/ready")
+        code=$(status_code "$ANSWER")
```
joaquinbejar, line 137: 🟠 **Warning**: with `set -euo pipefail`, `status_code` returns nonzero whenever curl has not produced an HTTP status line, so this assignment aborts the script on the first startup or reconnect race instead of continuing all sixty attempts. Make a missing status parse succeed with an empty value, or guard the assignment here, so the retry loop actually retries.

.github/workflows/replica-isolation.yml
```diff
@@ -0,0 +1,41 @@
...
+#
+# Deliberately NOT on every push: it builds the service image and runs four
+# containers, and it has nothing to say about a change that touches neither the
+# probes, nor what they probe, nor how the service is packaged.
+name: Replica isolation (readiness per instance)
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
```
joaquinbejar, line 23: 🟠 **Warning**: this filter omits files that directly define the behavior under test. `src/main.rs` selects the probes, `src/api/rest/routes.rs` mounts `/health` and `/ready`, and `src/api/rest/controller.rs` installs `x-ocs-instance`. A regression in any of them can merge without running this fixture. Include those wiring files and the Redis and MongoDB probe adapters, or use a broader relevant glob.

joaquinbejar (commented):
REQUEST_CHANGES. The fixture's retry loop can abort on the first transient connection failure, and its path filters skip several files that wire the readiness and attribution behavior it is meant to protect. The isolation assertions, cleanup, documentation, versioning, and six checks are otherwise sound.

scripts/replica_isolation_test.sh
joaquinbejar: You are right, and it is the same shape that ended an earlier run of mine on a missing header. Both readers are awk now and print an empty value instead of failing the pipeline, so the first attempt arriving before the server does is a retry rather than the end of the run.

.github/workflows/replica-isolation.yml
joaquinbejar: Agreed, I filtered on the modules the fixture is named after rather than on what decides its result. The filter now also covers main.rs, which chooses the probes, routes.rs and controller.rs, which mount the endpoints and install the header, and the Redis and MongoDB clients that decide what unreachable means.

joaquinbejar:
Both fixed. The retry one was the more serious of the two: the fixture would have aborted on its first attempt in CI and reported nothing about the behaviour it exists to protect.

PR #151, opened 2026-08-31
Build the published image with the arrow export it advertises (#148)

### LINKED ISSUE #148: The published image cannot serve the arrow export it advertises
## What happens

The integration suite, run against a deployment of 0.2.24, reports:

```
INFO: arrow is advertised but not built here: Validation Error: format: arrow is unavailable:
this build does not have the `arrow-export` feature; use packed, json or csv
```

The OpenAPI document the same deployment serves lists `arrow` among the export formats, so a client reading the contract asks for a format the service then refuses. The refusal is well formed and names the reason, which is why the suite skips rather than fails; what is wrong is that the document and the build disagree.

## Why

`Docker/Dockerfile` builds with `make release`, which is `cargo build --release` with default features, and `arrow-export` is off by default. Nothing in the image or the compose file turns it on, so every published image refuses `format=arrow` while advertising it.

## What to decide

One of two, and either is fine as long as the document and the build agree:

1. **Build the image with the feature.** `arrow-export` becomes part of what a deployment offers, and the format matrix in the integration suite covers it against a real deployment instead of skipping.
2. **Stop advertising it where it is absent.** The schema is generated from the handler's annotations, so this means gating the advertised value on the feature, and saying in the crate docs that arrow is a build-time option.

Option 1 is the smaller change and the one that matches how the format is documented today. Option 2 costs a conditional in the API surface and leaves consumers with a format they can only get by building their own image.

## Acceptance criteria

- [ ] A deployment either serves `format=arrow` or does not advertise it
- [ ] The integration format matrix stops reporting a skip for arrow against a deployment, or the skip is replaced by an assertion that the format is absent from the document too
- [ ] Whatever is decided is stated in the crate docs beside the export formats
- [ ] If the feature is built in, the image size and build time change are noted in the PR

Found by running the integration suite against a deployment of 0.2.24.

joaquinbejar:
## Summary

A deployment's OpenAPI document lists `arrow` among the export formats, and every published image refuses it: `Docker/Dockerfile` builds through `make release`, which built with default features, and `arrow-export` is off by default. A client that reads the contract asks for a format the service declines. The refusal is a well-formed 400 naming the field, which is why the integration suite skipped rather than failed, and why the disagreement survived four releases.

Chosen option 1 of the two the issue lists: build the feature into the image. It is the smaller change and it matches how the format is documented; option 2 would have put a conditional in the API surface and left consumers a format they can only get by building their own image.

## Changes

- `make release` carries `arrow-export`, through a `RELEASE_FEATURES` variable so `make release RELEASE_FEATURES=` still builds without it. Off by default in `Cargo.toml` stays right for a library consumer, who should not pull the arrow tree to use the crate; a deployment is the opposite case, because it publishes the document that offers the format.
- The integration suite gains `test_the_deployment_serves_every_format_it_advertises`. The rest of the export matrix still treats an advertised-but-absent format as a skip, which is what lets it run against any build; this test is where that stops being acceptable, because what it holds is the agreement between the document and the binary behind it.
- The crate docs say what the published image carries and how to build without it.
- `Makefile` joins the replica-isolation workflow's path filter, since it now decides what the image can serve.

## Technical decisions

- **The new test fails against a deployment older than this PR**, on purpose. That is the defect being reported, not a flake: until an image built from here is deployed, the document and the build disagree. Verified both ways below.
- **No new dependency.** `arrow` was already an optional dependency of this crate.

## Measurements

| build | binary |
|---|---|
| `make release RELEASE_FEATURES=` | 34 844 848 bytes |
| `make release` | 36 022 464 bytes |

1.2 MB, about 3.4%, for the arrow tree. Compile time on this machine moved from about 10 s to about 13 s incremental; the image's cold build is dominated by the dependency graph either way.

## Testing

- [x] Against a locally run binary built by `make release`: `INFO: every advertised format is served: json, csv, arrow, packed`, 7 export tests passed
- [x] Against a binary built without the feature, and against the deployed 0.2.24: the new test fails naming `arrow`, which is the state this PR fixes
- [x] `make pre-push` clean, strict clippy clean

## Stack

- Base branch: `tests-cross-instance-coverage`
- Stacked on: #150
- Blocks: #153
- Stack: #150 (merged) → **this** → #153
- The diff is cumulative until #150 merges, so read it against the base branch rather than against `main`.

Closes #148

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,9 +66,19 @@ all: test fmt lint build
 build:
 	cargo build
 
+# Features the release binary carries, and therefore what the published image
+# serves (Docker/Dockerfile builds through this target).
+#
+# `arrow-export` is off by default in Cargo.toml because a library consumer
+# should not pay for the arrow dependency to use the crate. A DEPLOYMENT is the
+# opposite case: the OpenAPI document it serves advertises `format=arrow`, so an
+# image without the feature refuses a format its own contract offers (issue
+# #148). Override with `make release RELEASE_FEATURES=` to build without it.
+RELEASE_FEATURES ?= arrow-export
+
 .PHONY: release
 release:
-	cargo build --release
+	cargo build --release $(if $(RELEASE_FEATURES),--features "$(RELEASE_FEATURES)",)
 
 # Run tests
 .PHONY: test
--- a/examples/integration/tests/exports.rs
+++ b/examples/integration/tests/exports.rs
@@ -383,6 +383,45 @@ fn arrow_cell(column: &dyn arrow::array::Array, index: usize) -> String {
     }
 }
 
+/// A deployment serves every format its own document advertises.
+///
+/// The two can disagree: `arrow-export` is a Cargo feature, so a build without
+/// it refuses `format=arrow` in a typed 400 while the OpenAPI document the same
+/// process serves still lists the format. That is a contract a client cannot
+/// follow, and it shipped in every published image up to 0.2.25 (issue #148).
+///
+/// The rest of this file treats an advertised-but-absent format as a skip,
+/// which is what lets it run anywhere. This test is where that stops being
+/// acceptable, because what it holds is not the format matrix but the
+/// agreement between the document and the build behind it.
+#[test]
+fn test_the_deployment_serves_every_format_it_advertises() {
+    let Some(client) = service() else {
+        return;
+    };
+    let Some(exported) = Exported::create(&client) else {
+        return;
+    };
+    let offered = offered_formats(&client, &exported);
+
+    assert!(
+        offered.absent.is_empty(),
+        "this deployment advertises {:?} and cannot serve them: {:?}. A format in the document \
+         is a promise to a client that reads it, so either the build carries the format or the \
+         document stops offering it",
+        offered
+            .absent
+            .iter()
+            .map(|(format, _)| format.as_str())
+            .collect::<Vec<_>>(),
+        offered.absent
+    );
+    println!(
+        "INFO: every advertised format is served: {}",
+        offered.usable.join(", ")
+    );
+}
+
 /// Every dataset carries the same rows and the same values in every format the
 /// deployment advertises and serves.
 #[test]
--- a/.github/workflows/image-contract.yml
+++ b/.github/workflows/image-contract.yml
@@ -1,0 +1,89 @@
+# Runs the export contract against the IMAGE, not against a cargo build.
+#
+# The export tests describe what a deployment serves, and a deployment is the
+# published image. Nothing else in CI puts the two together: the hermetic suite
+# runs in-process with whatever features the job passes, and the replica
+# isolation job builds the image but only asks it about readiness. That gap is
+# how an image that advertised `format=arrow` and refused it shipped four times
+# (issue #148).
+#
+# So this builds the image the way the Dockerfile does and asks it for every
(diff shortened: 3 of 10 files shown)


Makefile
```diff
@@ -66,9 +66,19 @@ all: test fmt lint build
...
 	cargo build
 
+# Features the release binary carries, and therefore what the published image
+# serves (Docker/Dockerfile builds through this target).
+#
+# `arrow-export` is off by default in Cargo.toml because a library consumer
+# should not pay for the arrow dependency to use the crate. A DEPLOYMENT is the
+# opposite case: the OpenAPI document it serves advertises `format=arrow`, so an
+# image without the feature refuses a format its own contract offers (issue
+# #148). Override with `make release RELEASE_FEATURES=` to build without it.
```
joaquinbejar, line 76: 🟠 **Warning**: This documented opt out still builds a service whose OpenAPI advertises Arrow while the endpoint rejects it, recreating the exact contract mismatch in issue 148. Keep the library feature optional, but gate the advertised format for featureless binaries and test that branch, or do not present this as a supported deployment build.

examples/integration/tests/exports.rs
```diff
@@ -383,6 +383,45 @@ fn arrow_cell(column: &dyn arrow::array::Array, index: usize) -> String {
...
+/// process serves still lists the format. That is a contract a client cannot
+/// follow, and it shipped in every published image up to 0.2.25 (issue #148).
+///
+/// The rest of this file treats an advertised-but-absent format as a skip,
+/// which is what lets it run anywhere. This test is where that stops being
+/// acceptable, because what it holds is not the format matrix but the
+/// agreement between the document and the build behind it.
+#[test]
+fn test_the_deployment_serves_every_format_it_advertises() {
+    let Some(client) = service() else {
```
joaquinbejar, line 399: 🟠 **Warning**: This assertion skips in every current pull request job because `OCS_INTEGRATION_BASE_URL` is never set. The replica isolation workflow builds the exact Docker image but only checks readiness, so a published image without Arrow would still pass. Run this test against that image in CI, or extend the image fixture to request an Arrow export.

joaquinbejar (commented):
REQUEST_CHANGES. The published image now carries Arrow, but the documented featureless release path still advertises a format it rejects and no pull request check exercises Arrow from the built image. The version metadata, documentation synchronization, and remaining checks are consistent.

Makefile
joaquinbejar: Fair, that opt out shipped the same mismatch this PR is about. The format parameter description is a constant selected by the feature now, so a build without arrow does not advertise it and the typed 400 stays for anyone who asks anyway. A test asserts the document and the build agree, and CI runs the suite both with the feature and without it, so both branches are covered.

examples/integration/tests/exports.rs
joaquinbejar: You are right that nothing asked the image itself. There is a new workflow that builds it the way the Dockerfile does, runs it over Redis and MongoDB service containers and runs the whole export matrix against it, so the assertion runs against the artefact that ships rather than against a cargo build.

joaquinbejar:
Both fixed. The first was the sharper one: the featureless build advertised a format it refused, which is the very defect this PR set out to remove, one build away.

joaquinbejar:
Copilot's first pass hit its quota; the review that followed came through and is answered above.

Issue #152, opened 2026-08-31
A stored tape exports different bytes from a replayed one


joaquinbejar:
## What happens

With snapshot persistence on, the same tape exports different BYTES depending on whether its steps had been served before the export. The values agree; the rendering does not, in the last digit or two of a computed price:

```
stored:   0,...,0.54,0.55,0.5415620196854147,...
replayed: 0,...,0.54,0.55,0.5415620196854148,...

stored:   1,...,3.53,3.54,3.5336205103787828,...
replayed: 1,...,3.53,3.54,3.5336205103787823,...
```

Both rows describe the same quote. The mid price crosses a float on the way into storage and back, and the shortest round-trip rendering of that float is not the rendering of the `Decimal` the replay computes.

Reproduced against a locally run 0.2.26 with `OCS_SNAPSHOT_PERSISTENCE_ENABLED=true` over ClickHouse: create two simulations with the same parameters and seed, walk one to the end, export both as `dataset=option_chains&format=csv`. Found by the persistence coverage added for #149.

## Why it matters

`test_the_same_export_twice_is_byte_identical` holds today because both exports come from the same path. This is the case it does not cover: a consumer that diffs two exports of the same tape — IronCondor's reproducibility gate is exactly that shape — sees a change that is not there, and the same simulation exported before and after its rows were filed is not the same file.

## What to decide

1. **Do not store the derived value.** The mid is `(bid + ask) / 2`; recomputing it on read from the stored bid and ask, in `Decimal`, would make both paths render the same digits, provided bid and ask survive the round trip exactly.
2. **Store it as a decimal rather than a float**, which costs a schema change and a migration on an existing table.
3. **Render both paths through the same rounding**, so the export never shows more digits than the domain means.

Option 1 looks smallest and removes a stored column rather than adding one, but it needs checking that a bid and an ask survive storage exactly before anything is derived from them.

## Acceptance criteria

- [ ] Two exports of the same tape, one served from storage and one replayed, are byte identical
- [ ] The integration test in `examples/integration/tests/persistence.rs` stops reporting a byte-level near miss and compares bytes again
- [ ] Whatever is chosen is stated where the export formats are documented
- [ ] The change is checked against an existing table, not only a fresh one, if it touches the schema

Created from the coverage work for #149.

joaquinbejar:
[...]

PR #153, opened 2026-08-31
Say which export path a deployment exercises, and compare the two (#149)

### LINKED ISSUE #149: Snapshot persistence is never exercised against a deployment
## What happens

A deployment of 0.2.24 answers readiness with two dependencies:

```json
{"status":"ready","dependencies":[{"name":"redis","status":"up"},{"name":"mongodb","status":"up"}]}
```

No `clickhouse`, which means snapshot persistence is off: the warehouse probe is only registered when the manager has one. Every export served by that deployment therefore replays the tape, and the branch that reads back persisted snapshots — the one that exists so a long export does not re-walk — is exercised only by the in-process tests.

## Why it matters

The export is where the warehouse pays off and where it can fail in ways the in-process tests do not see: a schema that drifted, a query that times out, a row count that disagrees with the replay. The integration suite compares formats and ranges against each other, so a warehouse that returned subtly different rows would be invisible today, because both sides of every comparison come from the same replay.

## What to build

1. Enable the warehouse in the deployed stack, or in a stack the suite can be pointed at.
2. Add an integration test that tells the two paths apart and skips loudly when persistence is off: with the warehouse on, `/ready` names `clickhouse`, and an export of a simulation whose steps were served earlier must equal the same export of a simulation walked and exported immediately.

## Acceptance criteria

- [ ] The suite reports whether the deployment it is pointed at has persistence on, rather than silently testing one of the two paths
- [ ] A test that an export backed by persisted snapshots matches the same export produced by replay, skipped with a reason when persistence is off
- [ ] The compose file documents which variables turn it on and what it costs

Found by auditing what the integration suite covers against a deployment of 0.2.24.

joaquinbejar:
## Summary

An export is served two ways: with snapshot persistence off the tape is replayed, and with it on the steps a client has already been served are read back from the warehouse instead of being walked again. The suite could not tell them apart, so every comparison it made had replay on both sides and a warehouse returning subtly different rows would have been invisible.

Running the new comparison found exactly that, which is why this PR does **not** turn persistence on by default: see the decision below.

## Changes

- `examples/integration/tests/persistence.rs`, new. `test_the_deployment_says_whether_it_persists_its_snapshots` reads the readiness body — a `clickhouse` dependency is persistence being on, since the warehouse probe exists only when the manager has a warehouse — and says which path the run exercises. `test_a_stored_tape_exports_what_a_replayed_one_does` walks one tape to the end so its steps are filed, leaves a second tape of the same seed unwalked, and requires the two exports to agree cell by cell, retrying while the detached write queue drains.
- The compose file documents `OCS_SNAPSHOT_PERSISTENCE_ENABLED`: what turning it on buys, what it costs today, and the two consequences an operator should know (the schema is created at boot, so an unreachable warehouse fails the boot; `/ready` gains a `clickhouse` dependency, so a replica whose warehouse is down reports itself unready).
- The crate docs say the readiness body is where the export path is visible.

## The decision this PR took, and why

The issue offers "enable the warehouse in the deployed stack". I wrote that change, ran the suite against it, and reverted the default to `false`.

With persistence on, a stored export renders a computed mid price a digit differently from a replayed one:

```
stored:   ...,0.54,0.55,0.5415620196854147,...
replayed: ...,0.54,0.55,0.5415620196854148,...
```

The values agree; the bytes do not. And it is not only across simulations: an export taken before the rows are filed differs from the same export seconds later, so `test_the_same_export_twice_is_byte_identical` and `test_an_export_is_byte_identical_across_instances` both fail against a warehouse-backed deployment. Flipping the default would trade a documented guarantee for the saving, which is an owner's call and not mine to make silently.

Filed as #152 with the reproduction, three options and the acceptance criteria. The default flips when it is fixed, and the comparison here goes back to bytes.

Meanwhile the comparison is value by value, with the byte-level near miss printed and named rather than hidden. A truncated decimal, a timestamp that lost its zone, a reordered column or a missing row is still a failure.

## Testing

- [x] Against a locally run 0.2.27 with `OCS_SNAPSHOT_PERSISTENCE_ENABLED=true` over ClickHouse: both new tests pass and report the near miss as #152
- [x] Against a deployment with persistence off: the comparison skips with its reason and the reporting test says which path is exercised
- [x] `make pre-push` clean, strict clippy clean; the hermetic suite is unaffected

## Stack

- Base branch: `stack/1-148-arrow-in-the-image`
- Stacked on: #151
- Blocks: #154
- Stack: #150 (merged) → #151 → **this** → #154
- The diff is cumulative until #151 merges, so read it against the base branch rather than against `main`.

Closes #149

### DIFF
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -129,7 +129,7 @@ services:
     # Published by .github/workflows/docker-publish.yml on every Cargo.toml
     # version bump, or by `make docker-push` from a workstation. Bump the
     # default tag in the same PR that bumps the crate version.
-    image: ghcr.io/joaquinbejar/optionchain-simulator:${OPTIONCHAIN_VERSION:-0.2.26}
+    image: ghcr.io/joaquinbejar/optionchain-simulator:${OPTIONCHAIN_VERSION:-0.2.27}
     deploy:
       # More than one instance is the intended shape: session and simulation
       # state lives in Redis, and the write that decides who advanced a
@@ -157,6 +157,23 @@ services:
       CLICKHOUSE_DB: ${CLICKHOUSE_DB:-default}
       CLICKHOUSE_PORT: ${CLICKHOUSE_PORT:-8123}
       CLICKHOUSE_HOST: ${CLICKHOUSE_HOST:-clickhouse}
+      # This stack SHIPS the warehouse two services up, and still does not use
+      # it by default (issue #149). Set this to true to turn snapshot
+      # persistence on; what that buys is an export that reads back the steps
+      # already served instead of re-walking them.
+      #
+      # What it costs today, measured rather than assumed: an export served
+      # from storage renders a computed price a digit differently from one that
+      # replayed the tape, so two exports of the SAME tape stop being byte
+      # identical (issue #152). The values agree; the bytes do not, and a
+      # consumer that diffs exports sees a change that is not there. The
+      # default flips to true when that is fixed.
+      #
+      # The other consequences an operator should know: the schema is created
+      # at boot, so a warehouse that cannot be reached fails the boot rather
+      # than degrading quietly, and `/ready` gains a `clickhouse` dependency, so
+      # a replica whose warehouse is down reports itself unready.
+      OCS_SNAPSHOT_PERSISTENCE_ENABLED: ${OCS_SNAPSHOT_PERSISTENCE_ENABLED:-false}
       MONGODB_URI: ${MONGODB_URI:-mongodb://admin:password@mongodb:27017}
       MONGODB_DATABASE: ${MONGODB_DATABASE:-optionchain_simulator}
       MONGODB_STEPS_COLLECTION: ${MONGODB_STEPS_COLLECTION:-steps}
--- a/examples/integration/tests/persistence.rs
+++ b/examples/integration/tests/persistence.rs
@@ -1,0 +1,328 @@
+//! Which of the two export paths a deployment actually exercises.
+//!
+//! An export can be served two ways. With snapshot persistence off, the tape is
+//! replayed: every step is walked again to render the rows. With it on, the
+//! steps a client has already been served are in the warehouse, and the export
+//! reads them back rather than re-walking, which is the whole point of storing
+//! them.
+//!
+//! The rest of this suite cannot tell the two apart, and that is the problem
+//! (issue #149): every comparison it makes has replay on both sides, so a
+//! warehouse returning subtly different rows would be invisible. These tests
+//! say which path the deployment in front of them uses, and hold the two to
+//! producing the same bytes where they can be compared.
+//!
+//! What they cannot do is force the path. Nothing in the client-facing surface
+//! reports whether an export was read back or re-walked, so a deployment with
+//! persistence on and an empty warehouse simply replays and still passes. The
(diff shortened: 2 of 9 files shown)


examples/integration/tests/persistence.rs
```diff
@@ -0,0 +1,328 @@
...
+            let stored = served.export(&query);
+            assert_eq!(
+                stored.status,
+                200,
+                "a stored {dataset} export must serve: {}",
+                stored.text()
+            );
+            stored_body = stored.body.clone();
+            differences = compare(dataset, &stored_body, &expected.body);
+            if differences.is_empty() {
```
joaquinbejar: 🟠 **Warning**: An empty differences list is also what happens before any detached write reaches ClickHouse, because the export then replays every missing step. Breaking here can green this test without exercising a persisted row, which is the exact false green issue 149 asks to remove. Wait for an observable persisted row or source signal before comparing, then assert the stored and replayed results.

Docker/docker-compose.yml
```diff
@@ -157,6 +157,23 @@ services:
...
       CLICKHOUSE_PORT: ${CLICKHOUSE_PORT:-8123}
       CLICKHOUSE_HOST: ${CLICKHOUSE_HOST:-clickhouse}
+      # This stack SHIPS the warehouse two services up, and still does not use
+      # it by default (issue #149). Set this to true to turn snapshot
+      # persistence on; what that buys is an export that reads back the steps
+      # already served instead of re-walking them.
+      #
+      # What it costs today, measured rather than assumed: an export served
+      # from storage renders a computed price a digit differently from one that
+      # replayed the tape, so two exports of the SAME tape stop being byte
```
joaquinbejar, line 167: 🟠 **Warning**: This correctly records that stored and replayed exports are not byte identical, but the authoritative crate docs still promise “Either source renders the same bytes” at `src/lib.rs:622`, and the export module repeats that guarantee. Qualify those public docs in this PR until issue 152 is fixed; otherwise the deployment documentation and public reproducibility contract contradict each other.

joaquinbejar (commented):
REQUEST_CHANGES. The comparison can pass before ClickHouse stores any row, and its own deployment note contradicts the public byte identity guarantee. The stack, version metadata, and remaining checks are consistent.

examples/integration/tests/persistence.rs
joaquinbejar: You are right, and there was no signal to wait on, which is why it was written that way. There is one now: a counter of the quote rows the warehouse has accepted, incremented on the write that succeeded rather than on the advance that queued it. The test waits for it to rise before comparing and skips loudly when it never does, so a green result means storage was actually exercised.

Docker/docker-compose.yml
joaquinbejar: Agreed, the deployment note and the public docs said opposite things. Both are qualified now: the values agree row for row, byte-identical repeats hold with persistence off which is the default, and where it is on the bytes agree once #152 lands. The export module carried the same promise twice and both are fixed.

joaquinbejar:
Both fixed. The first one mattered most: the comparison could go green before a single row reached ClickHouse, which is the false green this PR was supposed to remove, so the deployment now publishes a counter of the rows it has filed and the test waits on it.

joaquinbejar:
Copilot's first pass hit its quota; the review that followed came through and is answered above.

PR #154, opened 2026-08-31
Render an exported value by its number, not by how it was stored (#152)

### LINKED ISSUE #152: A stored tape exports different bytes from a replayed one

joaquinbejar:
## Summary

An export served from the warehouse rendered a computed price a digit differently from the same export replayed, so two exports of one simulation stopped being byte identical the moment its rows were filed. A consumer that diffs exports — IronCondor's reproducibility gate is exactly that shape — saw a change that was not there.

## What was actually wrong

Not the column: it is `Decimal(38, 28)` and the round trip preserves the value exactly. A `Decimal` is a mantissa and a scale, so one number has many forms, and the round trip changes the form — the value is scaled out to twenty-eight decimals on the way in and comes back with its trailing zeros stripped. `Decimal::to_f64` reads the mantissa, and two forms of the same number can land on adjacent floats:

```
0.004924864250536683924737972      -> 0.0049248642505366836
0.0049248642505366839247379720     -> 0.004924864250536684
```

Measured on the reference arithmetic, 87 of 2000 values render differently depending only on the scale they carry.

## Changes

- `decimal_to_f64` normalises before converting, and a new `positive_to_f64` routes every `Positive` through it. The export module already funnelled both sources through one `QuoteView` for exactly this reason; the conversion inside it was the part that was not blind to representation. Every conversion in the module goes through the two helpers now, including the underlying and volatility datasets and `days_to_expiration`.
- The integration comparison in `persistence.rs` goes back to comparing bytes.
- The compose file turns snapshot persistence on, which is what #149 deferred until this was fixed, with the operator consequences documented beside it.
- The crate docs state the rule: an exported value renders by its number, not by how it was stored.

## Technical decisions

- **Normalising beats the alternatives the issue listed.** Not storing the mid and recomputing it on read would have left every other stored column with the same sensitivity; a schema change would have needed a migration and fixed nothing about the rendering. This is one line at the choke point, and it makes the wire value a function of the number alone.
- **No behaviour change for anyone reading a replayed export**: normalising cannot change the value, only which of two adjacent floats a padded form picks, and the padded form only ever came from storage.

## Testing

- [x] `test_a_value_renders_the_same_whatever_scale_it_carries`: over two thousand values with long expansions, the priced form and the stored form must render the same bits. Fails without the fix, naming the value
- [x] `test_a_stored_quote_views_as_the_priced_one_does`: one layer up, through `QuoteView`, with the column's round trip reproduced
- [x] Against a locally run service with the warehouse on, the whole integration suite passes: 61 tests, including `test_the_same_export_twice_is_byte_identical` and `test_an_export_is_byte_identical_across_instances`, which failed before this
- [x] `make pre-push` clean, strict clippy clean, 946 workspace tests

## Stack

- Base branch: `stack/2-149-persistence-coverage`
- Stacked on: #153
- Stack: #150 (merged) → #151 → #153 → **this**
- It sits on #153 because the acceptance criteria include the integration test that only exists there, and because turning persistence on in the compose file is the promise that PR made. The diff is cumulative until #153 merges, so read it against the base branch.

Closes #152

### DIFF
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -2,9 +2,16 @@ version: '3.8'
 
 # Deployable stack: the four services the binary actually needs. Redis and
 # MongoDB are HARD requirements (src/main.rs propagates their connection errors
-# with `?`, so the process exits without them); ClickHouse is optional
-# (src/domain/simulator.rs logs the failure and runs with database_repo: None,
-# which only degrades the Historical source).
+# with `?`, so the process exits without them).
+#
+# ClickHouse is optional to the CRATE and required by THIS stack. The historical
+# source degrades without it (src/domain/simulator.rs logs the failure and runs
+# with database_repo: None), but this file turns snapshot persistence on, and
+# that makes the warehouse a startup dependency — the schema is created at boot
+# — and a readiness one: `/ready` then names `clickhouse`, so a replica whose
+# warehouse is down reports itself unready and takes no work. Set
+# OCS_SNAPSHOT_PERSISTENCE_ENABLED=false to go back to a stack where ClickHouse
+# is only the historical source.
 #
 # The admin UIs (mongo-express, redis-commander) and the host-published
 # infrastructure ports live in docker-compose.dev.yml, NOT here: `docker stack
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -129,7 +129,7 @@ services:
     # Published by .github/workflows/docker-publish.yml on every Cargo.toml
     # version bump, or by `make docker-push` from a workstation. Bump the
     # default tag in the same PR that bumps the crate version.
-    image: ghcr.io/joaquinbejar/optionchain-simulator:${OPTIONCHAIN_VERSION:-0.2.27}
+    image: ghcr.io/joaquinbejar/optionchain-simulator:${OPTIONCHAIN_VERSION:-0.2.28}
     deploy:
       # More than one instance is the intended shape: session and simulation
       # state lives in Redis, and the write that decides who advanced a
@@ -157,23 +157,20 @@ services:
       CLICKHOUSE_DB: ${CLICKHOUSE_DB:-default}
       CLICKHOUSE_PORT: ${CLICKHOUSE_PORT:-8123}
       CLICKHOUSE_HOST: ${CLICKHOUSE_HOST:-clickhouse}
-      # This stack SHIPS the warehouse two services up, and still does not use
-      # it by default (issue #149). Set this to true to turn snapshot
-      # persistence on; what that buys is an export that reads back the steps
-      # already served instead of re-walking them.
+      # This stack SHIPS the warehouse two services up, so it uses it (issues
+      # #149 and #152). What it buys is an export that reads back the steps
+      # already served instead of re-walking them, which is the expensive half
+      # of answering one.
       #
-      # What it costs today, measured rather than assumed: an export served
-      # from storage renders a computed price a digit differently from one that
-      # replayed the tape, so two exports of the SAME tape stop being byte
-      # identical (issue #152). The values agree; the bytes do not, and a
-      # consumer that diffs exports sees a change that is not there. The
-      # default flips to true when that is fixed.
+      # It was held off until an export served from storage rendered the same
+      # bytes as one that replayed the tape (#152); it does now, so this is on.
       #
-      # The other consequences an operator should know: the schema is created
-      # at boot, so a warehouse that cannot be reached fails the boot rather
-      # than degrading quietly, and `/ready` gains a `clickhouse` dependency, so
-      # a replica whose warehouse is down reports itself unready.
-      OCS_SNAPSHOT_PERSISTENCE_ENABLED: ${OCS_SNAPSHOT_PERSISTENCE_ENABLED:-false}
+      # The consequences an operator should know: the schema is created at
+      # boot, so a warehouse that cannot be reached fails the boot rather than
+      # degrading quietly, and `/ready` gains a `clickhouse` dependency, so a
+      # replica whose warehouse is down reports itself unready. Set it to false
+      # to go back to replay-only.
+      OCS_SNAPSHOT_PERSISTENCE_ENABLED: ${OCS_SNAPSHOT_PERSISTENCE_ENABLED:-true}
       MONGODB_URI: ${MONGODB_URI:-mongodb://admin:password@mongodb:27017}
       MONGODB_DATABASE: ${MONGODB_DATABASE:-optionchain_simulator}
       MONGODB_STEPS_COLLECTION: ${MONGODB_STEPS_COLLECTION:-steps}
--- a/examples/integration/tests/persistence.rs
+++ b/examples/integration/tests/persistence.rs
@@ -242,19 +242,43 @@ fn test_a_stored_tape_exports_what_a_replayed_one_does() {
         return;
     };
 
+    // The whole tape, rendered from a walk inside the export itself. Its row
+    // count is also how many quote rows the warehouse has to accept before the
+    // other tape can be served from storage rather than replayed.
+    let reference = fresh.export("dataset=option_chains&format=csv");
+    assert_eq!(
+        reference.status,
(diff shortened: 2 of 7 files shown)


joaquinbejar:
Copilot declined with a quota limit, so the review happened in this session instead. What it looked at, since the change is one line at a conversion: that normalising cannot move a value, only pick between two adjacent floats for a padded form that only storage produces; that every conversion in the module goes through the two helpers, including the ones outside the quote view; and that the property test actually exercises the case, which it asserts by counting the values it checked and failing if too few reached the comparison. Both tests were run against the code with the fix reverted and they fail there, naming the value.

src/api/rest/export.rs
```diff
@@ -1402,13 +1402,13 @@ fn json_rows(
             "step": step,
             "simulated_at": simulated_at,
             "symbol": symbol,
-            "price": row.spot.to_f64(),
+            "price": positive_to_f64(row.spot),
```
joaquinbejar, line 1422: 🟠 **Warning**: This canonicalizes only the text encodings. `visit_typed_rows` still emits `row.spot` and `row.base_volatility` through `Positive::to_f64` directly in `binary.rs:405` and `:411`, although the public contract says packed and Arrow carry the same f64 values as JSON and CSV. A high scale factor row can therefore become adjacent floats across formats. Make the canonical converter reusable by `binary.rs` and cover a padded factor value in the cross format tests.

examples/integration/tests/persistence.rs
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        assert!(
+            differences.is_empty(),
+            "the {dataset} export of a tape whose steps were served differs from the same export \
+             of the same seed rendered on the spot: {differences:?}. Same seed, same parameters, \
+             so the rows must not depend on whether they came back from storage"
+        );
+
+        assert_eq!(
```
joaquinbejar, line 327: 🟠 **Warning**: This byte assertion can still compare replay against replay. While the detached writer is draining, or if it has failed, the export falls back to replay for every missing step, so the first identical response makes `differences` empty and exits the retry loop before any persisted row is read. Require positive provenance that all requested steps were served from storage, or query ClickHouse for them, before accepting the byte comparison.

src/api/rest/export.rs
```diff
@@ -1630,6 +1656,144 @@ mod tests {
...
+                Err(error) => panic!("the stored mid must read back: {error}"),
+            }
+        };
+        assert_eq!(
+            stored_mid, priced_mid,
+            "the round trip must preserve the value, or this test is about something else"
+        );
+
+        let strike = exactly(Decimal::from(5_000_i64));
+        let priced_view = QuoteView::stored(&QuoteRow::new(strike, exactly(priced_mid)).with_call(
```
joaquinbejar: 🟠 **Warning**: This regression test does not exercise either source adapter or a representation change. `priced_mid` is normalized before the simulated round trip, that round trip strips back to the same mantissa and scale, and both values are passed through `QuoteView::stored`. The test therefore passes even if `QuoteView::replayed` or either adapter bypasses the helper. Build the priced side through the replayed adapter and assert that the raw pre-fix conversions differ before checking the canonical result.

Docker/docker-compose.yml
```diff
@@ -157,6 +157,20 @@ services:
...
+      #
+      # It was held off until an export served from storage rendered the same
+      # bytes as one that replayed the tape (#152); it does now, so this is on.
+      #
+      # The consequences an operator should know: the schema is created at
+      # boot, so a warehouse that cannot be reached fails the boot rather than
+      # degrading quietly, and `/ready` gains a `clickhouse` dependency, so a
+      # replica whose warehouse is down reports itself unready. Set it to false
+      # to go back to replay-only.
+      OCS_SNAPSHOT_PERSISTENCE_ENABLED: ${OCS_SNAPSHOT_PERSISTENCE_ENABLED:-true}
```
joaquinbejar, line 180: 🟠 **Warning**: Flipping this default makes ClickHouse a hard startup and readiness dependency for this deployable stack, but the file header still says ClickHouse is optional and only degrades the historical source. Update that top level operational contract in the same change so operators are not given the opposite failure behavior.

joaquinbejar (commented):
REQUEST_CHANGES. The canonical conversion is sound, but the binary scalar path bypasses it, neither new test proves the storage or adapter boundary it claims to cover, and enabling persistence leaves the stack documentation contradicting its hard ClickHouse dependency. The incremental scope, version metadata, layering, and checks are otherwise consistent.

src/api/rest/export.rs
joaquinbejar: Good catch, the contract says the binary formats carry the same values and that path converted on its own. Both call sites go through the canonical converters now, which are reachable from that module, and there is a test over a value whose two forms bind to different floats without them.

examples/integration/tests/persistence.rs
joaquinbejar: Right, waiting for the counter to twitch is satisfied by any other simulation. It now requires the filed rows to rise by at least this tape own quote row count, taken from the replayed export, and skips loudly when they never do.

src/api/rest/export.rs
joaquinbejar: You are right, that test had no teeth: both sides went through the stored adapter and the round trip returned the form it started from. The priced side goes through the replayed adapter now, and the value is searched for rather than assumed, with the test first requiring the two forms to render differently through the raw conversion so it cannot pass on a value that never carried the problem. Reverting the fix fails it, naming the value.

Docker/docker-compose.yml
joaquinbejar: Agreed, the header promised the opposite failure behaviour. It now says that this stack makes ClickHouse a startup dependency, because the schema is created at boot, and a readiness one, so a replica whose warehouse is down takes no work, and how to go back to a stack where it is only the historical source.

joaquinbejar:
All four fixed. The one about the test itself was the most useful: it would have passed with the fix reverted, which makes it worse than no test. It now goes through both adapters and refuses to run on a value that does not exhibit the defect.

